PR #319, opened 2014-04-03
Split provisioning of WP sites

cfoellmann:
WP sites provisioning is not "hardcoded" into the provision.sh anymore.
Makes it easier to abandon these sites for the user


jeremyfelt:
Definitely still on the table. A most important next step. :)

cfoellmann:
@jeremyfelt can we finally move this in?

_____________________
We could also move these sites out of the repo.
My idea:
1. remove all site from the VVV repo
2. move site configs (provision scripts) to a new repo (e.g. `vvv-sites`)
3. add config file(s) to `provision/` to pull in selected sites from the `vvv-sites` repo in the provision process

Config files could be as simple as this (should be possible!?):
[local.wordpress-trunk.dev.sh]
```sh
https://github.com/Varying-Vagrant-Vagrants/vvv-sites/local.wordpress-trunk.dev
```

[local.wordpress.dev.sh]
```sh
https://github.com/Varying-Vagrant-Vagrants/vvv-sites/local.wordpress.dev
```

This most simple sh file could be read in the provision process and provisioned. This way a simple custom file would allow for easy integration of custom sites to be provisioned.

What do you guys think?

iandunn:
I think it'd be nice if it also supported pulling in an entire group of sites, rather than having to list each one. e.g.,

```bash
[vvv-sites]
https://github.com/Varying-Vagrant-Vagrants/vvv-sites/

[wordpress-meta-environment]
https://github.com/iandunn/wordpress-meta-environment/
```

ericandrewlewis:
Let's consider scaffolding. Pretend we have a command like `vvv create site`. Entering this command takes you through an interactive shell, perhaps:

`Would you like to install WordPress for this site? (y/n)`
```
What type of WordPress install would you like?
 [1] Simple WordPress install
 [2] Trunk
 [3] Development repo
```
`What domain would you like to use for this site (e.g. yourdomain.dev)?`

The scaffold creates a folder in `www/`, and inside installs your preferred flavor of WordPress, and creates a `vvv-hosts` and `vvv-nginx.conf` file with the answers you supplied.

jeremyfelt:
Scaffolding is definitely tempting for the future as a built in command. It can probably live here for now - https://github.com/bradp/vv - as that seems to accomplish things well for those who need it.

I think these steps are likely for the immediate future:
* Break existing VVV core sites out into individual repositories.
* Swap provisioning for these sites in `provision.sh` by pulling the repo and firing `vvv-init.sh`.
* Provide a way to configure which of these sites is setup by default.
* Apply this to other repositories.

I also like @iandunn's point about pulling in a collection of scripts. This all deserves a bunch of thought.

I'm going to find time this weekend to go through a couple things here and get other repositories setup. We can then start making some progress with PRs pretty quickly.

cfoellmann:
@jeremyfelt can you rename `vvv-multisite` to `vvv-sites` so I can begin setting up the sites there? Or is that premature?

cfoellmann:
I use this branch just to easily delete the standard sites :hear_no_evil: 

@jeremyfelt can we finally start  #534 with this?

cfoellmann:
@jeremyfelt feature branch-worthy?

iandunn:
Looks like the new PR for this is #611, for those who want to follow it.

cfoellmann:
Yes, sorry.
Because you cannot change the target of a PR I needed to create a new one.

Issue #442, opened 2014-09-05
Configure for a Public Network Issue


salcode:
## Background
- Vagrant 1.6.3
- VirtualBox 4.3.13
- Mac OS X 10.9.4

VVV is working properly for private_network.
I can load the default webpage at http://vvv.dev/ and http://192.168.50.4/

I'm coming to VVV after being a long time MAMP user.
## What I'm Trying to Do

I'm trying to make my VVV websites accessible from another computer on my network using the dynamically assigned IP to my machine (similar to MAMP).
I have read https://github.com/Varying-Vagrant-Vagrants/VVV/wiki/Configure-for-a-Public-Network and #263 (this ticket addresses loading specific sites within VVV, not my connection issue)

Based on this reading I've done the following.
## Changes I've made

In my **Vagrantfile**, I've made the following modification

```
#config.vm.network :private_network, ip: "192.168.50.4"
config.vm.network :public_network
```
## Commands I've run

Run `vagrant reload --provision`

When my machine comes back up, I'm given the choices:

```
1) en0: Wi-Fi (AirPort)
2) en1: Thunderbolt 1
3) en2: Thunderbolt 2
4) bridge0
5) p2p0
    default: What interface should the network bridge to?
```

and choose **option 1**
## I expect

When the machine finishes booting, I expect to see the vvv.dev webpage when I go to 
http://127.0.0.1/, [http://localhost/](http://localhost/), or http://10.0.1.6/ 
(the last of these being the IP address I get from running `ipconfig getifaddr en0`)
## What actually happens

I am unable to load the webpage from any of these locations

Clearly, I'm missing something but I haven't been able to determine what.  Any suggestions on next steps in troubleshooting this?

tconroy:
I'm having an issue where if I enter the public network IP directly ( for me it's 192.168.1.17:8080 ) I am at the VVV dashboard, however I can't seem to navigate to the wordpress-default directory. How are you accessing your wordpress installs?

salcode:
@tconroy If you're seeing the dashboard at http://192.168.1.17/:8080,
try http://local.wordpress.192.168.1.17.xip.io:8080

tconroy:
Yup that did it, thanks. Guess I'll need to use the xip.io service.

PR #445, opened 2014-09-12
Add network config info for host local IP address

salcode:
By default, the guest machine is accessed by the host machine via
192.168.50.4.  Instead, we want to make the guest machine accessible
from the host machine's IP address.  This will allow us to access the
guest machine from other devices on the network (e.g. a smart phone)

I had a lot of issues setting this up and think an example in the config file would be helpful.

However, if others aren't having issues with this setup, adding this information may be overkill.

See #442


jeremyfelt:
Closed via #560, thanks @salcode!

PR #464, opened 2014-10-01
Add mailcatcher support

### LINKED ISSUE #337: Mailcatcher?
I'm setting up VVV after not using it a while and I've gotten used to using [Mailcatcher](mailcatcher.me) during development. I am up to building this into VVV but wanted to see if this is something that others would want as well. I saw in the provision.sh Postfix is already getting setup and some may WANT emails to actually be sent out rather than catching them.

EHLOVader:
This adds the setup and configuration for mailcatcher.

I have chosen to use RVM to install mailcatcher to provide the best support for the application and chose to install RVM from source per recommendations of the Ruby community.

This configures the PHP smtp to use catchmail and access to mailcatcher is available on port 1080.

Resolves #337


grappler:
:+1:

EHLOVader:
Updated the PR with a rebase. Also rewording, I can squash those later.

jeremyfelt:
Thanks, @EHLOVader. Will be looking soon! :)

dwainm:
Thanks for for this solution @EHLOVader ! 

The provisioner has installed mailcatcher as I can access the interface here:   http://vvv.dev:1080/

I ran some tests on http://local.wordpress-trunk.dev/ but none of my emails came in through mailcatcher. I also attempted to setup an smtp service with the following settings but with no success:

host: localhost
port: 1025
tls/ssl : none
authentication: none 

What other configurations should be done before this will work?

EHLOVader:
It shouldn't need anything else. It should be handling the PHP
configuration and the service starting.

Let me look into it.

For the email test, did you just attempt the reset password?

EHLOVader:
Ahah, I found the problem. 

Switching to an upstart config and php5enmod to clean up the other configs. The issue is with running in the background as a daemon and there doesn't appear to be much in the way of a fix yet.

https://github.com/sj26/mailcatcher/issues/182

This has happened before where the dependencies had some issue or failure and they broke mailcatcher. The use of upstart is pulled from [serversforhackers.com](https://serversforhackers.com/mailcatcher/) so hopefully it is more reliable.

Testing this once more and then will rebase and push.

EHLOVader:
@dwainm Give the PR a try again. There were issues which have apparently been a problem with Mailcatcher when run in background/daemon mode. I believe i have fixed it and it should run perfectly.

grappler:
I tested it and it works. Do you know if it is possible to have a custom url like the other services then to use the port?

cfoellmann:
If the mailcatcher is generating the UI itself that would not be possible because the 80 and 443 ports are already used by nginx.

EHLOVader:
I tried to get it to do that. I don't think it will work like that though. 

I attempted to make a reverse proxy config for nginx where it took the port and made it a path. Perhaps I did it wrong but it didn't want to work.

dwainm:
@EHLOVader thanks for the updates. On running the latest commit I found the following errors when provision:

```
==> default:  * rsync'd /srv/config/homebin                          to /home/vagrant/bin
==> default:  * RVM                                                 1.26.9
==> default:  * Mailcatcher                                         0.5.12
==> default: /tmp/vagrant-shell: line 375: syntax error near unexpected token `<<<'
==> default: /tmp/vagrant-shell: line 375: `<<<<<<< HEAD'
went wrong.
```

I'm not sure if I should have destroyed my vagrant box after the update or if this is a bug.

zamoose:
Sounds like you had a merge conflict on a Git pull and the diff is still in one of your files. Might be time forms fresh checkout or at the very least a scrub for merge conflict diffs in your files. 

--
Doug Stewart

dwainm:
@zamoose doing a fresh install now.

dwainm:
After running vagrant up on @EHLOVader mailcatcher branch. The interface doesn't work, but it seems mailcatcher is somewhat installed. When I run the mail catcher command within vagrant this is what it gives me: 
/usr/local/rvm/rubies/ruby-2.2.0/lib/ruby/2.2.0/rubygems/dependency.rb:315:in `to_specs': Could not find 'i18n' (~> 0.7) - did find: [i18n-0.6.11] (Gem::LoadError)
Checked in 'GEM_PATH=/usr/local/rvm/gems/ruby-2.2.0@mailcatcher:/usr/local/rvm/gems/ruby-2.2.0@global', execute `gem env` for more information

EHLOVader:
I can take a look at it again. I did a basic install for Mailcatcher but it
seems development with Mailcatcher is volatile enough that it has broken
three times, usually because of the dependencies.

dwainm:
After reloading and refreshing all the things the Mailcather interface simply doesn't get any emails. I've checked my php sendmail path and it is set correctly:

sendmail_from	:      no value	no value
sendmail_path:	/usr/local/rvm/bin/catchmail -f admin@local.dev

I've tried testing this boing getting a password reset email but still nothing. Not sure why it works for everyone else though.

EHLOVader:
I think that the i18n error is expected. It will still output even with the mitigation to the dependency issue found here https://github.com/sj26/mailcatcher/issues/155#issuecomment-53204461

The problems with PHP not sending through mailcatcher though is odd. There was another issue with mailcatcher and background process where the smtp server would timeout. That should have been fixed though in my most recent commits and rebase. The details of that problem can be found here. https://github.com/sj26/mailcatcher/issues/182

Do you know if you had a long wait on the "forgot password" page?

dwainm:
@EHLOVader  I had long waits yes, after entering the address and submitting.

EHLOVader:
Yup. I am getting that error about i18n afterwards from the shell too. I want to see if I can find a way to install all the dependencies from a specific version instead of dealing with this happening. Locking them in to a specific point in time. One that worked.

dwainm:
@EHLOVader not sure if this will help but I came across this method:  http://mark.chouinard.me/mailcatcher-vvv/

EHLOVader:
Thanks @dwainm, the one problem with that might exist with that method is that it runs in the background by default. I am currently using the foreground parameter to the mailcatcher command, inside an upstart script. Grabbed this attempt at a sensible start script from servers for hackers https://serversforhackers.com/mailcatcher/

I am currently testing removing the fix mentioned before, it was what removed the i18n and reinstalled v 0.6.11 which was the last known version to work.

EHLOVader:
OK... Looks like i18n is working on v0.7 now so the problem I was fixing by removing it and installing a prior version was to fix something wrong specifically with 0.7.beta1 it seems. Could you try pulling the latest commit and vagrant up (fresh)? Thanks @dwainm

dwainm:
@EHLOVader sorry for the late reply. I'm on a cruise ship with a bad line.

After checking out the latest changes on the 'mailcatcher' branch and a successful vagrant up, mailcatcher works without issue.

Thank you for adding more awesomeness to VVV!

config/php5-fpm-config/mailcatcher.ini
```diff
@@ -0,0 +1 @@
+sendmail_path = "/usr/local/rvm/bin/catchmail -f admin@local.dev"
```
cfoellmann: lets add a new line at the end here

config/init/vvv-start.conf
```diff
@@ -9,4 +9,5 @@ service nginx start
 service php5-fpm start
 service memcached start
 service mysql start
+service mailcatcher start
 end script
```
cfoellmann: lets add a new line at the end here when we are on it.

config/init/mailcatcher.conf
```diff
@@ -0,0 +1,8 @@
+description "Mailcatcher"
+
+start on runlevel [2345]
+stop on runlevel [!2345]
+
+respawn
+
+exec /usr/local/rvm/bin/mailcatcher --foreground --http-ip=0.0.0.0
```
cfoellmann: lets add a new line at the end here

cfoellmann:
this works like a charm.
I really love this.

dwainm:
+1 @cfoellmann . I've been running this branch exclusively for the past 3 months. Can't wait for it to go into master.

jdevalk:
Been using this all day today, works like a charm.

EHLOVader:
So glad to hear this is working as well as it did for me for all of you.

Added the whitespace requested. I have in the past had issues with it on commit or perhaps something with my past IDEs.

Let me know if you would like a rebase to consolidate these into one commit. I refrained for now.

cfoellmann:
You can just merge `develop` into the branch.
You should also change the target branch of the PR to `develop`. Problem is that you can not change it without creating a new PR :sad

dwainm:
@EHLOVader @cfoellmann I suggest only opening up a new PR if @jeremyfelt confirms that it will be merged. This will just keep a good history of all the issues for others trying this out.

EHLOVader:
Oh no! I noticed that there was a `develop` branch. Had I missed that all this time or did it start during the new version release or license change? 

Could I just rebase to `develop`... I don't think you would set the upstream for my branch to this repo right? it would just be more compatible with develop if I rebased it to that.

I am sorry if I had just missed it, but a new PR is definitely possible I will wait for a response. The people using this branch is one of the reasons I didn't want to rebase and force push anyway.

dwainm:
@EHLOVader  A simpler solution would be to create a new branch off the 'develop' branch and then to cherry-pick the 3 commits from this branch into the newly create branch.

EHLOVader:
This PR is no longer valid.

I have submitted a new PR based on `develop` and done some quick testing. I haven't found any problems but if those who are using this branch could confirm the replacement PR is #632

I will keep this branch around for a while but it will not be kept forever. 

@jdevalk @dwainm @cfoellmann Thanks for your help.

dwainm:
Thanks @EHLOVader will check it out on the other side :)

PR #465, opened 2014-10-03
Tweaked xdebug

iamntz:
Tweaked Xdebug to play nice: 

- added correct paths
- added JIT as remote mode 
- turned off autostart; instead, you need to pass `XDEBUG_PROFILE=1` as a GET or set a cookie


cfoellmann:
@iamntz can you have a look at #586 to see if there is anything to do make your changes work with the xdebug versions workaround implemented there?

jeremyfelt:
I'm going to close this out for now. Thank you very much for the PR, I think we're mostly at the point where we need to be.

I may update the default path for xdebug to `/tmp/`, as that's where files are being stored now and webgrind seems to work as expected with all of the current data.

PR #473, opened 2014-10-13
Update xdebug.ini to use same IP as Vagrantfile

Jimtrim:


Mte90:
uhm but a ifconfig -a return:
`inet addr:192.168.50.4`

PR #475, opened 2014-10-14
Add database backups to project folders.

Daronspence:
I've added support for database backups within project folders. Especially useful if you are tracking your project in git. No more sending the database separately! 

Works out of the box for projects created with the [VVV-Site-Wizard](https://github.com/aliso/vvv-site-wizard) and can be leveraged in projects whose folder name and database are the same.
Eg. `{vvv}/www/myproject/` and `myproject.sql`.

If the names do not match up, a warning "error" is thrown and the script continues.

I also added color coding and formatting to the output to make things prettier in the shell. 

Readme file was updated to explain the seemingly hidden feature to new users.

Traditional backups to `{vvv}/database/backups/` still work as expected.


ThatGerber:
My concerns as a user: 

**It doubles the amount of disk space used by VVV.**

I run high content websites and my database files are very large (250mb gzipped. ~1gb inflated). The amount of storage space taken up by the database on each installation will double. There will be the standard backup location and and the secondary backup location. Will there be an option to disable the secondary database file for committers who do not commit a database with their repo (for security or other reasons)?

**Adds further confusion to where database files are being loaded from**

Are they being pulled from the databases folder? Are they being pulled from the project root? I don't like storing the same information in two different places because it makes the user question the "source of knowledge".

**It's a problem that can be solved with a pre-commit/post-commit/post-merge hook**

It's not a problem that needs to be solved by VVV. This is something that could be better implemented through git hooks, since it's really an extension on the functionality of Git.

Example Post Merge Hook:

````
# Post Merge Hook
# MIT © Sindre Sorhus - sindresorhus.com, Chris W. Gerber - chriswgerber.com

# git hook to run a command after `git pull` if a specified file was changed
# Run `chmod +x post-merge` to make it executable then put it into `.git/hooks/`.

changed_files="$(git diff-tree -r --name-only --no-commit-id ORIG_HEAD HEAD)"

check_run() {
    echo "$changed_files" | grep --quiet "$1" && eval "$2"
}

# Check for database file
db_file='foo.sql'
db_dir='/vagrant/www/foo'

vagrant_command='vagrant ssh -c "wp db import $db_dir/$db_file"'

check_run "$db_file" "$vagrant_command"
````

Integrated with a setup/deploy script, it can be generated automatically.

jeremyfelt:
Thanks for the work on this pull request. I'm going to leave database backups alone right now, as I think it's a workflow issue that can be solved with custom scripting at the moment. We need to have a longer think about how VVV should be involved with the backup process, if at all. Thanks!

PR #479, opened 2014-10-26
Support for Parallels Desktop

TobiasBg:
Parallels Desktop offers an official and open-source Vagrant Provider plugin at http://parallels.github.io/vagrant-parallels/ and https://github.com/Parallels/vagrant-parallels.
The only real differences to our Virtualbox configuration that I have spotted after a quick glance are the box name and the way how machine settings are changed. Both of this can be overridden with provider-specific declarations in the Vagrantfile, as far as I can see.

I'll use this as a tracking issue for my experiments with adding support to VVV.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -128,7 +128,7 @@ Vagrant.configure("2") do |config|
     # those are specific to Virtualbox. The folder is therefore overridden with one that
     # uses corresponding Parallels mount options.
     config.vm.provider :parallels do |v, override|
-      override.vm.synced_folder "database/data/", "/var/lib/mysql", :mount_options => [ "share" ]
+      override.vm.synced_folder "database/data/", "/var/lib/mysql", :mount_options => []
     end
   end
 
@@ -161,7 +161,7 @@ Vagrant.configure("2") do |config|
   # those are specific to Virtualbox. The folder is therefore overridden with one that
   # uses corresponding Parallels mount options.
   config.vm.provider :parallels do |v, override|
-    override.vm.synced_folder "www/", "/srv/www/", :owner => "www-data", :mount_options => [ "share" ]
+    override.vm.synced_folder "www/", "/srv/www/", :owner => "www-data", :mount_options => []
   end
 
   # Customfile - POSSIBLY UNSTABLE
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -17,6 +17,13 @@ Vagrant.configure("2") do |config|
     v.customize ["modifyvm", :id, "--natdnsproxy1", "on"]
   end
 
+  # Configuration options for the Parallels Provider
+  config.vm.provider :parallels do |v|
+    v.update_guest_tools = true
+    v.optimize_power_consumption = false
+    v.memory = 1024
+  end
+
   # Forward Agent
   #
   # Enable agent forwarding on vagrant ssh commands. This allows you to use identities
@@ -29,6 +36,11 @@ Vagrant.configure("2") do |config|
   # box containing the Ubuntu 14.04 Trusty 64 bit release. Once this box is downloaded
   # to your host computer, it is cached for future use under the specified box name.
   config.vm.box = "ubuntu/trusty64"
+
+  # The Parallels Provider uses a different naming scheme.
+  config.vm.provider :parallels do |v, override|
+    override.vm.box = "parallels/ubuntu-14.04"
+  end
 
   config.vm.hostname = "vvv"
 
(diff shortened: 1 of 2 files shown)


TobiasBg:
@legal90 Very cool, thanks for the update on this! I'm still on PD 9 and everything is working fine with that now as well (the file access rights are slightly different in the guest VM, compared to the Virtualbox, but I did not run into any issues with that in practice). With this feature in PD 10.1.2, we might be able to strive for even closer matching PD/Virtualbox VMs in the future. But having support for PD 9 is probably also nice :-)

jeremyfelt:
This is pretty great, @TobiasBg, thanks! And thank you @legal90 for dropping in as well!

I checked out this PR and downloaded Parallels 10 for the first time to try things out. Here was my experience.

On the first `vagrant plugin install vagrant-parallels`, I got an error with Bundler inside Vagrant.

`Net::HTTPNotFound: No gems found matching "vagrant-vmware-fusion" "3.2.0" nil`

Strange, but on the next attempt the plugin installed fine. I'm going to blame Vagrant on that one. I'm still on 1.6.5, not sure if 1.7.2 would react differently.

On first `vagrant up --provider=parallels`, I got an error about a conflicting IP address after the box downloaded for the first time:

```
==> default: Setting the name of the VM: vvv_default_1420696114870_19801
The specified host network collides with a non-hostonly network!
This will cause your specified IP to be inaccessible. Please change
the IP or name of your host only network so that it no longer matches that of
a bridged or non-hostonly network.
```

I'm really confused because I destroyed my VirtualBox version of VVV completely. There must be some residual somewhere looking at 192.168.50.4.

I changed the IP in `Vagrantfile` to 192.168.60.4 and `vagrant up` then started booting the machine just fine to go through provisioning.

Provisioning went flawlessly. I had forgotten to remove the various WP directories, so it was nice how it was able to `svn up` in place.

At first I had to correct my hosts entries as they had some old `192.168.50.4` IPs. Once those were gone, then all of the default sites worked great!

For the unscientific speed test:
* Multisite unit tests took 3.09 minutes in Parallels, standard tests took 2.05 minutes.
* Multisite unit tests took 4.59 minutes in VirtualBox, standard tests took 4.35 minutes.

If those times are right, that's pretty fantastic and I might think about ditching VirtualBox for core dev. Wow.

I think my only worry from the entire process is around the IP address. Though my guess is most users will want to use Parallels because they do not have VirtualBox. We're probably ok to see what happens.

As long as we aren't breaking anything on the other end we're probably ok. I ran through a clean `vagrant up` with VirtualBox to run the unit tests again and things seemed to be fine.

* Any fears with merging this in and starting to officially support Parallels as a provider?
* Does the run as user for `tar` and `npm` have any adverse effects for the VirtualBox provisioner?

TobiasBg:
Very cool, @jeremyfelt! Great testing results!

I'm not sure about that
`
Net::HTTPNotFound: No gems found matching "vagrant-vmware-fusion" "3.2.0" nil
`
error, I have never seen that. I assume that it was a temporary network glitch, maybe.

For that IP address conflict: I remember something like this (not sure that it was the same), if I used the Virtualbox VVV, shut that down, and then used Parallels. I always had to reboot my Mac and then start over fresh. I assume that Virtualbox integrates into the OS/network drivers/etc. very tightly and doesn't clean up/unlock everything immediately. That shouldn't be an issue in practice however, except for people who must use both Virtualbox and Parallels for whatever they are doing.

From my side, there are no fears regarding merging this. I've been using this setup for my dev work (a little bit of WP core and plugin development) for the last couple weeks (since opening the PR) without any issues. I'm still on PD 9 and have used Vagrant 1.6.5 mainly, but also successfully updated to 1.7.1 a couple days ago, all with this PR on VVV master.

The file changes should all be safe for Virtualbox users, as they only apply to the Parallels provider, except for that sudo tar/npm thing, which affects everybody. I haven't had any issues with this however, and as we have been doing it with the `wp` command for a while it should be fine.

hubdotcom:
In tests, parallels seems to indeed run much faster

*Virtualbox 1.7.2*
==> default: Provisioning complete in 2218 seconds

```wrk -d120 http://local.wordpress.dev/
Running 2m test @ http://local.wordpress.dev/
  2 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   199.83ms   60.22ms 544.45ms   91.40%
    Req/Sec    25.40      4.45    34.00     67.08%
  6192 requests in 2.00m, 52.68MB read
Requests/sec:     51.59
Transfer/sec:    449.44KB```

vagrant destroy ~ 36 seconds

*Parallels 10.2.0*
==> default: Provisioning complete in 462 seconds

```wrk -d120 http://local.wordpress.dev/
Running 2m test @ http://local.wordpress.dev/
  2 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    93.04ms   17.12ms 151.35ms   66.67%
    Req/Sec    54.63      8.62    75.00     65.45%
  13173 requests in 2.00m, 112.07MB read
Requests/sec:    109.76
Transfer/sec:      0.93MB```

vagrant destroy ~ 11 seconds

PR #488, opened 2014-11-04
Add db_restore command in homebin so one-off restores can be done

sc0ttkclark:
`db_restore` maps to `/srv/database/import-sql.sh` so no duplicate code here, however adding the command to the homebin makes it more easily accessible when doing manual backup/restore, outside of `vagrant halt` or `vagrant provision` (while vagrant is already up and running).


jeremyfelt:
Related #489

jeremyfelt:
It's been a long time coming. ⌚

PR #507, opened 2014-11-19
Allow someone to port-forward into the guest

aubreypwd:
Allow port `:8484` to forward into the guest so we can port forward our external IP's to an internal one since the IP of the guest is outside of the router's range. For example, with my firmware I can't forward to an IP outside of `192.168.2.x` range, or I would just forward `:80` to `192.168.50.4`. Also, helps when using http://xip.io and testing on mobile devices.

Mainly PRing this because I keep having to modify it and gets overwritten intermittently from upstream. I tried configuring `:80` to forward in without success.


jeremyfelt:
Hi @aubreypwd, thanks for the PR! I'm going to leave this out of the default configuration for now, as it's something that should likely be decided in individual configurations. It is possible to add this to `Customfile` so that it isn't overwritten by future upstream changes.

Of note, we just added docs to `Vagrantfile` for port forwarding through #560 via #445.

aubreypwd:
Sorry for the poor PR, didn't realize it included my memory change. Thanks for the information.

PR #511, opened 2014-11-22
[PR] Add inline status updates to initial provisioning

jeremyfelt:
There's a point, especially in WordPress develop, where things
appear to hang for quite a while when `npm install` fires. While
this is expected, we should at least note that it's happening.


negativefix:
I realized today that when `vagrant up --provision` issued npm update process take quite deal of time to complete in Wordpress Develop. And no message is shown except _Updating npm packages..._.  If possible, it might be good to display a similar message to users.

PR #512, opened 2014-11-24
Add box name

EHLOVader:
This is an improvement upon the original [PR #460](https://github.com/Varying-Vagrant-Vagrants/VVV/pull/460)

This now uses the current directory name for the box name, and this allows you to effectively configure the box name without actually configuring it.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,7 +1,9 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 
+dir = Dir.pwd
 vagrant_dir = File.expand_path(File.dirname(__FILE__))
+vagrant_name = File.basename(dir)
 
 Vagrant.configure("2") do |config|
 
@@ -9,11 +11,16 @@ Vagrant.configure("2") do |config|
   # with possible backward compatible issues.
   vagrant_version = Vagrant::VERSION.sub(/^v/, '')
 
-  # Configuration options for the VirtualBox provider.
+  # Give vm a name
+  config.vm.define vagrant_name do |v|
+  end
+
+  # Configurations from 1.0.x can be placed in Vagrant 1.1.x specs like the following.
   config.vm.provider :virtualbox do |v|
     v.customize ["modifyvm", :id, "--memory", 1024]
     v.customize ["modifyvm", :id, "--natdnshostresolver1", "on"]
     v.customize ["modifyvm", :id, "--natdnsproxy1", "on"]
+    v.name = vagrant_name
   end
 
   # Configuration options for the Parallels provider.


cfoellmann:
@EHLOVader can you fix this PR please?

cfoellmann:
It would also be nice if we could name the `vvv.dev` host like `<box-name>.dev`/`<box-name>.vvv` (if we change from .dev -> .vvv)
This way the entries in the hosts file would not collide

EHLOVader:
Sorry about the delay. Does that work now? I still don't know what I did to lose that one line. 

Rebased to a single commit also. Hopefully the PR is useful. 

Tell me more about the host name switch? Was this to prevent the prior vvv.dev collisions with future vvv instances that use PWD name?

cfoellmann:
no problem. I can't think of a way to create a compatible branch with a missing line :8ball: 
About the ending change https://github.com/Varying-Vagrant-Vagrants/VVV/issues/583

When you run multiple vvv boxes - which is already possible without your fix :imp: - `vvv.de` and `vvv` collide.
Another issue is the ip address. I will look into selecting it dynamically. Which should be fine for all dev stuff that does not require access via IP.

cfoellmann:
+ https://github.com/Varying-Vagrant-Vagrants/VVV/issues/544

EHLOVader:
Line should be there, and those are the two problems I thought you might have had in mind. I am indifferent to the TLD I don't foresee myself symlinking the hosts file. But I also wouldn't mind if it changed. I can see how .vvv might make sense. Perhaps though .<box-name> would avoid collisions and allow for some configuration. 

I understand that you can run multiple vvv boxes now, but it still requires manually editing the vagrant file. I thought this would be helpful to prevent box name collisions. Especially if you are pulling down PRs or branches a lot of the time to test/dev on vvv itself I thought it would make things easier.

cfoellmann:
To get back to this:
Would this be a "breaking change"? Are there any bad effects by doing this to an existing vvv box?
Does anybody see/experience any problems with this?

jeremyfelt:
I'm actually wondering why I had trouble the first time. I thought I had tested and noticed a conflict, but my testing now shows that the box name has no trouble being changed between `vagrant halt` and `vagrant up`. If the box name is changed in `Vagrantfile` when the box is already started, nothing will change at all in VirtualBox. I think we're good!

I'm going to open a new PR with @EHLOVader's changes against `develop` and then make a small tweak to the code.

jeremyfelt:
Thanks @EHLOVader!

EHLOVader:
My pleasure! Thanks for doing the extra leg work to check it for breaking changes. I didn't think it did, but hadn't given it rigorous tests.

PR #524, opened 2014-12-14
Update readme to include vv, the sexy new site wizard.

bradp:


jeremyfelt:
This was merged as 915b4146, thanks @bradp!

Issue #113, opened 2013-08-07
Replace trunk WP from core.svn with develop.svn


jeremyfelt:
I think we should be able to do this without breaking much. My primary concern is that the new trunk has some documentation along with it in VVV that explains how to use it properly.

Todo:
- `www/wordpress-trunk` is checked out from develop.svn.wordpress.org
- `www/wordpress-unit-tests` is removed from new VVV installations
- Install grunt and dependencies (node) in the VM
- Confirm working unit tests inside new `wordpress-trunk/tests`
- Modify nginx config accordingly

Do we setup a URL pointing to `www/wordpress-trunk/build` and a URL pointing to `www/wordpress-trunk/src` to test both environments, or just src?

PR #527, opened 2014-12-15
Grunt build task should run after both after initialization AND update

ntwb:
Previously the Grunt build task `grunt` (Alias of `grunt build`) would only run if the `/build` folder did not exist, `grunt build` should be performed after initial initilaztion of the WordPress develop repo AND after updating the repo.

### DIFF
--- a/provision/provision.sh
+++ b/provision/provision.sh
@@ -542,6 +542,8 @@ PHP
 		cd /srv/www/wordpress-develop/
 		echo "Running npm install for the first time, this may take several minutes..."
 		npm install &>/dev/null
+		echo "Running grunt for the first time, this may take a minute..."
+		grunt
 	else
 		echo "Updating WordPress develop..."
 		cd /srv/www/wordpress-develop/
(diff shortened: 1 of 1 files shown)


jeremyfelt:
I think we're best off leaving `grunt build` as a process that is separate from provisioning. I'm starting to think we should keep `svn up` separate as well. The environment of the server should be considered somewhat separate from the environment of each project at some point.

> You also raise the point that `wordpress-trunk` is the same as `wordpress-develop`'s `/build` state, why does `wordpress-trunk` even exist then in VVV?

This is a very good question. The original purpose was so that plugin and theme developers could easily test their work against trunk. In 2013, when the `develop.svn` repo was introduced, I [decided](https://github.com/Varying-Vagrant-Vagrants/VVV/issues/113#issuecomment-22915076) to go with `wordpress-develop` so that we didn't conflict with anyone's existing `wordpress-trunk` workflow. See #113 for the full convo.

> How about we totally remove the `wordpress-trunk` SVN checkout which would save a considerable amount of time, use `grunt build` along with a symlinked `/build` folder for `wordpress-trunk`?

I would be definitely be in favor of reducing the number of SVN checkouts. 😃  See freshly opened #921.

I am also very much in favor of splitting off provisioning for these things into individual repositories. See #611. Then VVV can be extended to support many (or 0) default configs. A more proper config for meta development could probably be applied at that point as well. ❤️

Thanks for leaving your thoughts so long ago @ntwb, sorry for the delay. :)

ntwb:
All good @jeremyfelt, as time has passed and I've used VVV much more than when I initially created this issue I agree with your thoughts here Jeremy 👍

PR #528, opened 2014-12-15
Quote a few variables in provision.sh

bradp:
Quoting prevents word splitting and glob expansion, and prevents the script from breaking when input contains spaces, line feeds, glob characters and such.

If `var=my value`, the space is retained with quoting. Otherwise it will split and provide unexpected results. Nothing in the provision script is currently broken, its just a best practice to do it this way.


jeremyfelt:
Just ran through a fresh provisioning with this and all seems good. Let's do it.

PR #564, opened 2015-02-03
wp-cli support for local machine

### LINKED ISSUE #563: wp-cli support for local machine
Opening an issue here from [generator-vvv](https://github.com/lkwdwrd/generator-vvv/issues/39) I figured out that it can be accomplished in core and would benefit more this way.

> wp-cli is used in the VM for installation and configuration, but it would be really nice to have wp-cli support from outside the VVV VM. This should be possible through a custom wp-cli.yml in the project path.

> You could define the database connection for the VM using the external user and then you can use some of the scaffolding tools offered. I do foresee an issue with the folder paths however.

I actually have an implementation so I will be able to submit a PR for this shortly.

EHLOVader:
Closes #563 

This defines a custom `wp-cli.yml` for external use on the host machine. 

If you have `wp-cli` installed locally and wanted to use it there instead of through vagrant ssh or other methods this should do it. 

It accomplishes this by requiring a few php files to silence uneccessary warnings and to redefine the database for connecting from outside the VM. 

So far in all local testing it has worked great.

### DIFF
--- a/config/wp-cli/local.php
+++ b/config/wp-cli/local.php
@@ -1,0 +1,8 @@
+<?php
+
+define('DB_HOST', 'vvv.dev:3306');
+
+define('DB_USER', 'external');
+define('DB_PASSWORD', 'external');
+
+
--- a/config/wp-cli/quiet.php
+++ b/config/wp-cli/quiet.php
@@ -1,0 +1,4 @@
+<?php 
+error_reporting(0);
+@ini_set('display_errors', 0);
+define( 'WP_DEBUG', false );
--- a/wp-cli.yml
+++ b/wp-cli.yml
@@ -1,0 +1,3 @@
+require:
+  - config/wp-cli/local.php
+  - config/wp-cli/quiet.php


jeremyfelt:
This would be great, thanks @EHLOVader! I'm going to move it to the 1.3.0 milestone. Let's test it a bunch and get it in.

EHLOVader:
I also found out later that the internal `wp-cli` could be improved some by including the `config/wp-cli/quiet.php` itself. One poorly developed plugin in any WordPress site could turn a lovely green stream of output into a striped red mess. 

I have found quite a few that throw notices and more in modern PHP releases in the official plugin repository. 

You might look into this more and how it could be improved using the global `~/.wp-cli/config.yml` this `wp-cli.yml` and possibly even a `wp-cli.local.yml` in each project if needed.

[Config docs](http://wp-cli.org/config/)

jeremyfelt:
Finally tested it and I'm in love. Thanks again @EHLOVader, this is fantastic. :+1:

EHLOVader:
:feelsgood:

You have a chance to look into those various configs you could stash in the parent folders locally and on the server?  Also looking over it again the config references vvv.dev. 

Did you remove the .dev in response to DNS issues in a more recent issue/PR?

PR #570, opened 2015-02-07
Disable the treatment of 404 by Nginx

ram89:
By disabling the treatment of 404 by Nginx, it lets Wordpress the charge to manage the error 404 by charging the 404 page of the theme.

### DIFF
--- a/config/nginx-config/nginx-wp-common.conf
+++ b/config/nginx-config/nginx-wp-common.conf
@@ -47,7 +47,7 @@ if (!-e $request_filename) {
 location ~ \.php$ {
     # Try the files specified in order. In our case, try the requested URI and if
     # that fails, try (successfully) to pass a 404 error.
-    try_files      $uri =404;
+    # try_files      $uri =404;
 
     # Include the fastcgi_params defaults provided by nginx
     include        /etc/nginx/fastcgi_params;


jeremyfelt:
Hi @ram89, thanks for the PR!

I'm not sure that this is necessary. For example:

* If I visit `http://src.wordpress-develop.dev/index.php` or `http://src.wordpress-develop.dev/`, WP handles this properly without a 404.
* If I visit `http://src.wordpress-develop.dev/this-page-does-not-exist/`, WP handles the 404.
* If I visit `http://src.wordpress-develop.dev/index2.php`, Nginx handles the 404 because the PHP file does not exist.

Is there a specific case that does not match that behavior?

ram89:
I was getting the issue when I was developing the 404 for a custom install made by VV by Bradp. I was testing mycustomdomain.dev/frank and I was getting the 404 of Nginx. The only way to make my wordpress install to handle it was to deactivate the 404 in Nginx config.

hubdotcom:
Did you have permalinks set?

jeremyfelt:
I'm going to close this out as the 404 handling by Nginx in this scenario is by design. Feel free to continue to use the thread for troubleshooting though.

PR #575, opened 2015-02-13
set your GitHub token easily for Composer

cfoellmann:
Just create a simple text file named `provision/github.token` with only your gh token to automatically set the token for composer

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@
 /provision/provision-custom.sh
 /provision/provision-pre.sh
 /provision/provision-post.sh
+/provision/github.token
 
 # Ignore custom trigger scripts in config/homebin.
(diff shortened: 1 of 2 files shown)


jeremyfelt:
Ooh, interesting.

cfoellmann:
anyone else tested this?
It is pretty nice to have when working on VVV itself. The checkouts via composer is very slow once you reach the anonymous github api limit.

khromov:
:+1: Very nice

jeremyfelt:
Seems smart to me. I thought a bit about whether `/provision/` was the correct place for the toke file, but I think it makes sense in the end as other custom stuff is there and it's related directly to provisioning.

Thanks @cfoellmann!

Issue #574, opened 2015-02-12
set PHPCS default to WordPress


pbearne:
Hi Guys

Just little suggestion lets set the default standard to WordPress for the installed PHPCS 

this is the command needed
phpcs --config-set default_standard WordPress

Paul

westonruter:
Actually, it should be `WordPress-Core`. The `WordPress` standard is a superset of all the standards and shouldn't be used in practice. And this command should only be run when first cloning the repo for the first time.

PR #576, opened 2015-02-13
provision tools via composer

cfoellmann:
It _should_ be easy to speed up the provision process by using composer over git checkouts.
My composer setup in this PR works but I can not get the nginx aliases to work. 

__Anyone have an idea?__

The aliases would keep the current url structure for the vvv.dev vhost

__________________
Implements #574


jeremyfelt:
One reason we use the git clone of wp-cli is to provide an easy way to contribute there as well. We should probably leave that as is.

I thought the same thing about the WordPress code standards sniffer, but that may not be the case. @westonruter?

Otherwise, for things like PHPMyAdmin, this would definitely be an improvement as far as maintainability goes.

westonruter:
I definitely would encourage the WordPress Coding Standards repo to be distributed as a git clone as well to facilitate contributing.

cfoellmann:
I am totally with you. Was thinking the same after I sent the pr.
Will revert WP-cli and cs

cfoellmann:
at the current state the aliases work but no php is processed.
Anyone have an idea how to fix the aliased folders to process the .php files?

jeremyfelt:
@cfoellmann Sorry to leave you hanging so long. :) Do you want a feature branch to merge this into for testing?

cfoellmann:
Any idea how to get php working for the provisioned tools?

cfoellmann:
can someone _PLEASE_ fix the vhost aliases?
I am about to abandon VVV because nginx is just fucking stupig (or I am).

cfoellmann:
Currently the aliases work for .php files and for these only

cfoellmann:
@jeremyfelt please add `in progress` label

cfoellmann:
Will close this for now.
The destination branch is wrong + I need to revisit this topic when I got some free time again.

PR #581, opened 2015-02-26
added @pbearne for the contribution of #574

cfoellmann:


cfoellmann:
can we merge this to give props where it is due?

jeremyfelt:
Makes sense, good catch!

Issue #258, opened 2014-01-21
Docker


mastef:
Oh I just realized I posted this to a fork. It would possibly make more sense to use Docker? ( And Vagrant still as a wrapper for non-unix clients of course )

See the full case for it at : https://github.com/ericmann/vvv-apache/issues/1

tannerchung:
Was getting interested in trying to use docker with VVV and was excitedly reading this comment archive and found that the http://github.com/varying-vagrant-vagrants/vvv-docker is dead. Was this project shut down?

Issue #98, opened 2013-07-16
Facilitate switching PHP versions between  5.2.4, 5.3.x, 5.4+


westonruter:
Since the minimum [requirements](http://wordpress.org/about/requirements/) for WordPress are 5.2.4, and since the [vast majority](http://wordpress.org/about/stats/) of WordPress installs are still on PHP 5.2 followed by PHP 5.3, it is not very helpful that VVV comes with PHP 5.4.17RC1. In order to ensure broad compatibility in plugins and themes developed across the WordPress ecosystem, it's important that they be testable in the lowest version of PHP. It would be ideal if there could be some command to switch between PHP versions on the fly, or even better to assign different PHP versions to each site instance. I'm not sure what is feasible at the moment.

PR #586, opened 2015-03-11
Add PHPBrew to VVV

ndisidore:
## Why
This is my initial attempt at adding PHPBrew to VVV. Over the course of development, there were times where I more closely wanted to mimic our production environment, which may not be the latest version of PHP, may or may not have memcached installed, etc. Rather than just install a single version of PHP, I stumbled upon PHPBrew and thought it would be beneficial to integrate into VVV. PHPBrew allows you to easily build php versions on the fly, and turn on/off extensions like memcached, imagemagic, and APC.

More information on PHPBrew, along with it's command list, can be found here: https://github.com/phpbrew/phpbrew

## How to use
This installs phpbrew to /opt/phpbrew and registers a new binary, `phpswap`, that is a wrapper for phpbrew functionality. Switching PHP versions is as simple as issuing a single command from terminal
`phpswap 5.3.28`
This looks for the version installed locally, and, it doesn't find it, grabs the source and builds from scratch with the set list of extensions.
Restart the fpm process for the currently running version of PHP with `phpswap fpm-restart`
You can switch the version persist across reboots by using the -s flag, e.g. `phpswap -s 5.3.28`
Switch back to the system version with `phpswap system`.

## Notes
* Even though the script executes `phpbrew use/switch $PHP_VER`, it doesn't seem to register unless you run the command separately, e.g. `phpswap 5.5.20 && phpbrew use 5.5.20` or `phpswap -s 5.5.20 && phpbrew switch 5.5.20`
* After enabling/disabling extensions, issue `phpswap fpm-restart` for the changes to take effect on the web server


cfoellmann:
this sounds pretty cool.
Especially the check you added for xdebug. I will definitely check this out but I am afraid that this is too progressive for VVV that it might take a pretty long time to get merged :sob:

ndisidore:
Thanks! The nice thing about this implementation is that it still installs a system php version and phpbrew is turned off by default, meaning phpbrew is invisible to users unless they explicitly tell it to turn on.

I personally find it useful because while in an ideal world, we'd all be running the latest and greatest, sadly more often than not I find that's not the case.

I'll try pitching it there too

jeremyfelt:
Hi @ndisidore, this does sound pretty cool. Thanks for the PR!

We've had discussions (#98 comes to mind) around swappable PHP versions, but it's been hard for me to personally imagine how it fits in VVV itself. I think Docker (#258) has been the most interesting possibility previously, but that would be a huge challenge to integrate with VVV. Your proposal for PHPBrew is the most intriguing yet.

I like that we can still have a default PHP version installed and that `phpswap` only needs to be used as desired.

I've created a feature branch `feature/phpbrew` on the VVV repository. Would you mind changing the PR to be against that branch instead? We can then merge it in and make it easier to test as a whole.

Would love to hear feedback from others on this as well.

andyadams:
I tried phpbrew via @ndisidore's branch and it is *magic* for me. Here's my use case:

I develop plugins & themes for different clients. Some of them are custom-built for a single site, others are built for distribution to multiple sites. I use VVV to quickly set up a site for each individual client, but I was hitting issues where PHP 5.5 wasn't throwing errors that older versions did. It was embarrassing to hand a client broken stuff, so I wanted a way to test older versions of PHP as I developed. 

This PR (so far) has been lovely. Thank you @ndisidore!

config/bash_profile
```diff
@@ -37,4 +37,9 @@ eval "$(grunt --completion=bash)"
...
 
 # PHPCS path
-export PATH="$PATH:/srv/www/phpcs/scripts/"
\ No newline at end of file
+export PATH="$PATH:/srv/www/phpcs/scripts/"
+
+# PHPBrew Exports
+export PHPBREW_ROOT=/opt/phpbrew
+export PHPBREW_HOME=/opt/phpbrew
+. /opt/phpbrew/bashrc
```
cfoellmann: lets add a new empty line at the end

provision/provision.sh
```diff
@@ -178,12 +199,30 @@ if [[ $ping_result == "Connected" ]]; then
...
 	npm install -g npm-check-updates
 
-	# xdebug
+	# phpbrew
 	#
-	# XDebug 2.2.3 is provided with the Ubuntu install by default. The PECL
-	# installation allows us to use a later version. Not specifying a version
-	# will load the latest stable.
-	pecl install xdebug
+	# Install phpbrew and make available systewide
```
cfoellmann: typo: systewide

helgatheviking:
I like this. I was just able to switch my vagrant box to 5.3.28 because a client just reported some errors that I never caught because I didn't test against PHP pre-5.4

ndisidore:
Opened a new pull request to the feature branch based on [this comment](https://github.com/Varying-Vagrant-Vagrants/VVV/pull/586#issuecomment-91965397). 

[Add PHPBrew to VVV #635](https://github.com/Varying-Vagrant-Vagrants/VVV/pull/635)

PR #587, opened 2015-03-13
Added VMware Fusion provider

robneu:


cfoellmann:
I can't test this but this would allow VVV to run on 3 VM providers and that is awesome.

robneu:
This is the setup I use and it works fine with the latest version of VMware. I've been using it for a while, so it's been working for the last couple versions as well.

jeremyfelt:
Closed via merge into `develop`. I'm assuming this works as I don't have VMWare Fusion and Robert says it does. :smile:

Thanks @robneu!

PR #588, opened 2015-03-14
Allow custom phpMyAdmin config, without committing secret info to the repo

grantnorwood:
It's super helpful to have one phpMyAdmin instance for both local and remote databases.  By allowing custom configuration, without accidentally making passwords and other private info public, this becomes safe and convenient.


cfoellmann:
1. generally I like the idea of a custom config
2. if you want an mysql adminstration tool for local and remote you may be better of with a native admin tool like https://www.mysql.com/products/workbench/
3. I would recommend to look for a __full__ `config.inc.custom.php` and copy it _instead_ of the normal `config.inc.php` and not include the custom config. this way you can overwrite all settings.
4. you should not mix stuff in one PR. Please move the phpmyadmin update to a separate PR

grantnorwood:
Thanks for the feedback, @cfoellmann.  Here are my thoughts:

1. Me, too!
1. phpMyAdmin is my preferred tool, so being able to connect to everything from one place whenever possible is easier for me in general.
1. I thought the same thing at first!  But I decided that adding the `include()` at the end of the file is a great way to allow a default config, which can be extended/overridden when necessary.  For example, I'm both adding some new custom servers as well as overriding the default blowfish secret.  However, while I feel adding a separate custom file with the new/overridden config is more intuitive, and may even be easier to maintain across phpMyAdmin versions if the `config.inc.php` file is ever updated, I don't really have a strong opinion on this so I'm certainly willing to change that.
1. Apologies, I just kept committing to my my master branch after creating the pull request.  How do you recommend cherry-picking that 3rd commit of the last 4 into it's own pull request?  Should I be doing these in feature branches and creating one PR per branch?

**I'd like some feedback from others on number 3 above**, just to make sure the most helpful method for everybody is selected.

Also, I assume that there should be some documentation on whichever method is implemented, should I simply create a PR for the wiki updates, as well?

Again, thanks!

cfoellmann:
about 4. I would recommend to use feature branches in general. Makes it easier locally and helps to not block a PR by some independent part of it.

grantnorwood:
Agreed!

cfoellmann:
Maybe we can make it like the provision.sh workflow https://github.com/Varying-Vagrant-Vagrants/VVV/blob/master/.gitignore#L13-L15
your implementation would then be `config.inc-pre.php`

grantnorwood:
I feel that scatters things around a bit more.  Instead, I think that having a simple file copy of `config.inc.custom.php` immediately below where the `config.inc.php` file is copied makes it easier to understand what's happening, and it's grouped along with the other phpMyAdmin-related code.

Still occurs in `provision.sh`, though.

jeremyfelt:
Thanks for the PR @grantnorwood!

It does make sense to allow for a custom configuration. I think I'd rather see the check directling in `config.inc.php`:

```
if ( file_exists( 'config.inc.custom.php' ) ) {
    include( 'config.inc.custom.php' );
}
```

Then we don't need to maintain the check for it in `provision.sh` at all. In fact, I might roll back b18ca1347453da386733192930497b2c6f286218 as that can be an entirely custom configuration.

If that makes sense, can you reopen that in a PR against the `develop` branch?

jeremyfelt:
Closing this as fixed in #688 with a slightly alternate approach using a cherry-picked commit from here.

PR #590, opened 2015-03-14
Added grunt-rtlcss package to be installed since WP core is replacing cssjanus with rtlcss.

enejb:
See https://core.trac.wordpress.org/ticket/31332 and 
https://core.trac.wordpress.org/changeset/31573
for more info.


jeremyfelt:
Makes sense and looks good, thanks @enejb!

Issue #469, opened 2014-10-08
Symlinks in /www


jdoerck:
having a problem with vagrant disk following symlinks. I'd expect symlinked directories referenced in {vvv-dir}/www/ to be echoed on the disk, but they, for me at least, are not.

As far as I can tell its not a permissions issue.

Any help would be greatly appreciated.

jeremyfelt:
I tried to hack on this some more in the hopes of providing some additional information and got....mostly nothing. :grin:

I'm assuming an approach like this in OSX (Yosemite):
* On my host machine I have a web project in `~/Development/cool-web-project/`
* On my host machine I have VVV in `~/Development/vvv/`
* I create a symlink with `ln -s ~/Development/cool-web-project/ ~/Development/vvv/www/cool-web-project`
* `vagrant ssh`
* `cd /srv/www/cool-web-project/` - Nope.

I ran into a [post with promise](http://www.0xcb0.com/2013/04/08/create-symlinks-inside-a-virtualbox-shared-folder/) that did not work for me. This involved using `VBoxManage` to set a symlink creation option at the command line rather than in `Vagrantfile`. I'm almost certain this is intended to allow for symlink creation inside the VM to an existing shared directory—something that does appear to work.

I also tried getting somewhere by manually using `mount` inside the VM so that Vagrant itself wasn't controlling it. I'm guessing there's a possible end to that road, but I'm not confident in it.

It also seems like something **could** happen if the guest additions were updated, but I'm not convinced.

I'd almost wonder if [Vagrant's rsync folders](https://docs.vagrantup.com/v2/synced-folders/rsync.html) would fit for a scenario where a symlink on a host needs to exist inside the virtual machine.

Related #224

/cc @ryandonsullivan, @robneu

rezzz-dev:
I was actually able to get the symlink to work on the VM by using the Customfile method.  However on the host, it just created a directory and didn't actually symlink to where it needed to be. 

Any thoughts on how to get the host to respect VM symlink?

cfoellmann:
For me the stackoverflow method works #592

rezzz-dev:
Thanks @cfoellmann - Guess that works for Win, however I am on a Mac (OS X Yosemite to be specific) - I dropped that line in and nothing happened unfortunately.

Essentially the only way I can get my VM to work with the symlinks is to use the Customfile method, then I just have to remember to go specifically into the linked folder if I want to code.  Not ideal, but does work

[...]

PR #592, opened 2015-03-17
Allow symlinks in shared folders (in VIrtualBox)

cfoellmann:
Not allowing symlinks in the first place is a security/stability safety. SInce this is for development it should be possible.

MIght need additional changes for Parallels + Fusion


jeremyfelt:
Is this something that can be added in `Customfile` only? From conversation in #469, this method may only work in Windows. I'd like to know what other possible side effects of having it enabled are.

PR #608, opened 2015-04-13
install phpcs via composer

cfoellmann:
phpcs needs to be installed via composer to reduce the provision process significantly. My last `vagrant provision` took __7039__ seconds! mostly due to git clones.

### DIFF
--- a/provision/provision.sh
+++ b/provision/provision.sh
@@ -463,7 +463,7 @@ if [[ $ping_result == "Connected" ]]; then
 		fi
 	fi
 	# Install the standards in PHPCS
-	sudo /srv/www/phpcs/scripts/phpcs --config-set installed_paths ./CodeSniffer/Standards/WordPress/
+	sudo /srv/www/phpcs/scripts/phpcs --config-set installed_paths /srv/www/phpcs-wordpress/
 	sudo /srv/www/phpcs/scripts/phpcs --config-set default_standard WordPress-Core
 	sudo /srv/www/phpcs/scripts/phpcs -i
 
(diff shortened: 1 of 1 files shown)


cfoellmann:
I fixed the oversight of setting the correct path to the WP coding standards.
This should save 5-10 minutes for the first provision.

cfoellmann:
@jeremyfelt please add `in progress` label

jeremyfelt:
Thanks @cfoellmann!

* Is the switch in directories from `/srv/www/phpcs/CodeSniffer/Standards/WordPress` to `/srv/www/phpcs-wordpress` going to mess with any current installations?
* Why `sudo` now on the phpcs lines?

provision/provision.sh
```diff
@@ -476,9 +463,9 @@ if [[ $ping_result == "Connected" ]]; then
 		fi
 	fi
 	# Install the standards in PHPCS
-	/srv/www/phpcs/scripts/phpcs --config-set installed_paths ./CodeSniffer/Standards/WordPress/
-	/srv/www/phpcs/scripts/phpcs --config-set default_standard WordPress-Core
-	/srv/www/phpcs/scripts/phpcs -i
+	sudo /srv/www/phpcs/scripts/phpcs --config-set installed_paths /srv/www/phpcs-wordpress/
```
westonruter: This absolute path will break PHPCS when invoked from outside of Vagrant. This is why a relative path is used currently.

westonruter:
The path to the WPCS shouldn't be changed.

jeremyfelt:
@cfoellmann do you want to address the path naming and update the PR accordingly? It seems like this would help as long as we're consistent with the location.

jeremyfelt:
I think we need to regroup on this in a new pull request. I've opened issue #922 to describe the intent here. Thanks for the work on this @cfoellmann. :)

PR #333, opened 2014-04-20
add bootstrap dashboard that picks up added sites

Tarendai:
Upgrades the default dashboard to:

 - Look in the /srv/www folder for added sites and list them alongside the defaults
 - Descriptions of built in tools
 - A link to the documentation wiki
 - Bootstrap 3 styling

screen shot 2014-04-20 at 21 46 11

PR #609, opened 2015-04-13
set topdown/vvv-dashboard as new default dashboard

cfoellmann:
This replaces the "old" dashboard with https://github.com/topdown/VVV-Dashboard while keeping the dashboard-custom.php function

______________
replaces + closes #333


cfoellmann:
@jeremyfelt please add `needs-feedback` label

robneu:
Just thought I'd point this out as another alternative. I'm partial to it since I contributed to it. :P https://github.com/leogopal/VVV-Dashboard

stevenkword:
And another: https://github.com/stevenkword/vvv-dashboard-custom

topdown:
That default/vvv-init.sh needs some revision. I made some big updates that require jQuery. I think the changes make it worth the addition of jQuery. I didn't want to use CDN for offline purposes.

cfoellmann:
@topdown updated.

jeremyfelt:
I love all these custom dashboards, so much great stuff. I'd rather not tie VVV directly into any of them at the moment. I think what we have is basic and ugly, and that's fine. I can definitely see adding a section in the README (more visible than the wiki) with a list of dashboards similar to our [helpful extensions](https://github.com/Varying-Vagrant-Vagrants/VVV#helpful-extensions) section. It may even make sense to add a small note to the default dashboard to make it more obvious that something custom can be installed.

I'm absolutely okay to make any changes that make it easier to distribute a custom dashboard, though it seems like the current model works pretty well.

Issue #229, opened 2013-12-11
Current errors or red lines during the first `vagrant up`


jeremyfelt:
During provisioning on the first `vagrant up`, there are quite a few areas where red text floats by. These are often not errors, though sometimes are. It can be somewhat jarring and it would be nice to avoid as many as possible.

I'm including a checklist of all the errors currently being shown on a new clone of the repo as of 5c1033dd45d7b179aa3a9196e0feab0443af7e44. Work can be done on this over the next couple releases if needed.

* [ ] `stdin: is not a tty` right when provisioning begins.
* [x] gpg related errors when signing new apt sources
    * `gpg: new configuration file /root/.gnupg/gpg.conf' created`
    * `gpg: WARNING: options in /root/.gnupg/gpg.conf' are not yet active during this run`
    * `gpg: requesting key 7BD9BF62 from hkp server keyserver.ubuntu.com`
    * `gpg: requesting key A2F4C039 from hkp server keyserver.ubuntu.com`
    * `gpg: requesting key E5267A6C from hkp server keyserver.ubuntu.com`
    * `gpg: requesting key E1DF1F24 from hkp server keyserver.ubuntu.com`
    * `gpg: requesting key C7917B12 from hkp server keyserver.ubuntu.com`
* [ ] `dpkg-preconfigure: unable to re-open stdin: No such file or directory` after initial apt packages are downloaded and installation is about to begin.
* [ ] mysql configuration related errors
    * `[Warning] Using unique option prefix key_buffer instead of key_buffer_size is deprecated and will be removed in a future release. Please use the full name instead.`
    * `[Warning] Setting lower_case_table_names=2 because file system for /var/lib/mysql/ is case insensitive
mysql start/running, process 11658`
* [ ] `Creating config file /etc/php5/mods-available/pdo.ini with new version`
* [ ] `Creating config file /etc/php5/fpm/php.ini with new version`
* [ ] `Creating config file /etc/php5/cli/php.ini with new version`
* [ ] `Creating config file /etc/php5/mods-available/gd.ini with new version`
* [ ] `Creating config file /etc/php5/mods-available/imap.ini with new version`
* [ ] `Creating config file /etc/php5/mods-available/mcrypt.ini with new version`
* [ ] `Creating config file /etc/php5/mods-available/mysql.ini with new version`
* [ ] `Creating config file /etc/php5/mods-available/mysqli.ini with new version`
[...]

PR #610, opened 2015-04-13
fix 'stdin: is not a tty'

cfoellmann:
Needs testing!! Works for me so far.

Needs a description but I dont know what to write other than 'fixes `stdin: is not a tty`'
props @artemkaint

SEE: https://github.com/Varying-Vagrant-Vagrants/VVV/issues/229#issuecomment-76149619


artemkaint:
I found small problem with this fix. 
vagrant ssh -c 'some command' does not work

Now I am using 
<pre><code>config.ssh.shell = "bash" </code></pre>
and in provision 
<pre><code>export DEBIAN_FRONTEND=noninteractive </code></pre>
for fix "dpkg-preconfigure: unable to re-open stdin: No such file or directory"

cfoellmann:
Thanks for testing.
It is some deeper vagrant issue

cfoellmann:
Maybe @artemkaint can pitch in on this !?
What do you say? What is your experience?

jcart:
Why was this closed? What is the resolution?

cfoellmann:
@jrb this solution doesn't work correctly.
have a look at #633

zamoose:
@cfoellmann:
It works perfectly over on HGV for us. What do you mean, "doesn't work"?

cfoellmann:
This fix screws with `vagrant ssh`

zamoose:
In what way? Again -- it works perfectly for HGV. The fix given in the other ticket doesn't work during the first provisioning, IIRC.

zamoose:
Ahh, @cfoellmann I see what you mean -- it breaks `vagrant ssh -c [command]`. `vagrant ssh` continues to work correctly, but you cannot send arbitrary commands to the VM with `-c` with that fix.

cfoellmann:
How is #633 working for you?

Issue #544, opened 2015-01-13
.dev domain namecollisions


wturrell:
I'm finding, I can no longer access .dev sites served by vagrant from my (host) browser, even with the usual hosts file entries present and correct.

This is apparently due to something [ICANN did on purpose](https://icann.org/namecollision) – namely unused `.dev` and certain other domains are resolved to 127.0.53.53 to alert people to name collisions;  .dev is now, or soon will be, a valid TLD…

Weirdly though, the hosts file entries added by [hostsupdater](https://github.com/cogitatio/vagrant-hostsupdater) (192.168.50.4 for `vvv.dev` and the various WordPress branches) are being ignored by everything except Google Chrome – no matter how many times I restart the browser / clear it's history / clear the OS X DNS cache etc.

1) is there a way to make the browser/system honour /etc/hosts regardless? 
2) should we change .dev to something else in the project to avoid future confusion – e.g. `.vvv`?

wturrell:
[Asked about this on StackOverflow](http://stackoverflow.com/questions/27928920/127-0-53-53-name-collisions-hosts-file-ignored/) and subsequently discovered the cause; if /etc/hosts file is a symlink OS X does a DNS lookup first and only looks at /etc/hosts if it doesn't resolve.

Unclear why Chrome still worked (and it was only Chrome stable that did, not Canary.)

cfoellmann:
I am on Windows and do not have the issue but I like the idea of using `.vvv`

evelynriossf:
Thanks for this answer.  I had a similar problem while using Bedrock with VVV. For me, the problem only manifested with Chrome 40.0.2214.111 (64-bit) on OSX, and not with any other browser (Firefox, etc..).
Changing .dev to .vvv fixed the problem, as it doesn't resolve.

jeremyfelt:
Interesting stuff. `.dev` has always been weird, though seemed appropriate. I wouldn't mind switching to `.vvv` at some point in a backward compatible way.

morganestes:
Just to update this, I'm starting to get errors from ICANN when trying to view VVV sites in Chrome (47.0.2526.106 on OS X 10.11.2):

    dial tcp 127.0.53.53:80: connection refused

I can view the site by IP in Chrome, and by domain name in Safari. Dig returns the message:

    vvv.dev.		3599 IN	TXT "Your DNS configuration needs immediate attention see https://icann.org/namecollision"

[...]

PR #611, opened 2015-04-13
Split provisioning of WP sites

### LINKED ISSUE #102: Segment PHP logging to a separate log for each site
As more and more sites get added to a VVV instance, the PHP error log can quickly get overwhelmed with entries from all of the sites combined. So it would be ideal if each site got its own separate PHP log which could be tailed.

cf. https://github.com/10up/varying-vagrant-vagrants/commit/1fb1a47992250d81a61c4e0171e84eea66eb4ee1#commitcomment-3676483

### LINKED ISSUE #313: Ability to disable installation of default WP sites
As much as I like the idea of having quick and easy access to the different versions of WP installed by default with VVV, in practice this is only rarely useful for me personally, while making the vagrant provision process take longer.

I would like to propose two things:
1) The ability to disable the installation of the default WP sites (by commenting out a block in the Vagrantfile).

2) Split off the WP provisioning scripts into a separate file. This would help with the above, make it easier to customize/hack the provision.sh file and give the option of running the script independently of the provision process to add the WP site manually.

cfoellmann:
Move provisioning of standard WP site from `provision.sh` to split tasks.

Close #313
Close #102
__________
Moved from #319 to reflect the new git workflow.


jeremyfelt:
Hi @cfoellmann. Still haven't forgotten about this. :)

Just had a thought while reviewing the changes on how we can transition to using these **and** to using defaults provided externally at the same time.

* Via #544, we should pick a new TLD at some point. `.test` may be the right answer. It's worth thinking about using the same TLD that [HGV decides on](https://github.com/wpengine/hgv/issues/98) for consistency for users.
* In the new break-out sites provided by this PR, we could add some kind of check where things are only updated/created if this is an existing VVV configuration. (Not sure what that check is yet).
* If it is a new VVV config, pull the new TLD sites from individual GitHub repositories instead of having them packaged with VVV.
* Then provide a way of choosing which repositories to use.

Thoughts?

TheLastCicada:
Why not .vvv as a TLD to ensure a unique namespacing and avoid collisions with things like HGV, which conceivably could be running on the same machine testing the same sites you have on VVV?  Also, branding and things.

jeremyfelt:
But typing in `www.domain.vvv`? It seems funky. Using `.test` would fall in line with what the IETF intended. I could go either way, really. `.vvv` is apparently what I [thought originally](https://github.com/Varying-Vagrant-Vagrants/VVV/issues/544#issuecomment-75455718). :)

iandunn:
+1 for using one of the officially reserved TLDs, that's what they're for. Switching from one pseudo TLD to another isn't really solving the problem, it's just side-stepping it.

Granted, `.vvv` is less likely to issued than `.dev`, but it's still possible, and using a real TLD that's intended for local hosts lets OSs and browsers know what their dealing with, and potentially be smarter about handling them (e.g., OS X could skip the resolution process and go straight to `/etc/hosts`).

What percentage of VVV users are running HGV simultaneously? I would guess it's pretty small, and it's not hard for those users to change the TLD on their install.

Maybe it'd be best to make it easily configurable, but default to a reserved TLD, since that's best for the majority?

TheLastCicada:
The IETF argument is a good one.  Especially if anybody ever follows those guidelines and they actually mean something (I remain skeptical on that).

rheinardkorf:
Any progress on this?  Would love to split provisioning. Would help with #851

cfoellmann:
Closed in favor of https://github.com/Varying-Vagrant-Vagrants/VVV/pull/980

PR #616, opened 2015-04-19
Move network configuration up

bastibeckr:
It seems that vagrant-hostsupdater needs to know private network 
configuration (IP address). At least that sounds logical to me.
I had to change the IP address and found that vagrant-hottsupdater did
not write the new IP into `/etc/hosts`.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -53,6 +53,22 @@ Vagrant.configure("2") do |config|
   end
 
   config.vm.hostname = "vvv"
+  
+  # Private Network (default)
+  #
+  # A private network is created by default. This is the IP address through which your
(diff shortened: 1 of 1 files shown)


jeremyfelt:
Hi @bastibeckr! Thanks for the ticket and sorry for the delay.

I'm not sure this change needs to be made. At this point in `Vagrantfile`, config options are being set so that they are available when the plugin fires. This data should all be available to the plugin no matter the order. Is it possible something else was causing the trouble? I haven't noticed any issues with this positioning and we've had it configured this way since the beginning.

jeremyfelt:
Going to close this out for now. I haven't seen any problems with fresh boxes related to the positioning of the network configuration in `Vagrantfile`.

PR #619, opened 2015-04-21
Expose additional MySQL logs for peformance and troubleshooting.

### LINKED ISSUE #596: MySQL slow queries and queries not using index logs
I know it is easy to turn on manually, but I thought it might be useful to enable by default and configure the log location in `/srv/log` instead of `/var/log` which was the default. 

Would others, plugin developers or people working with applications outside of WordPress find that useful?

```
# Here you can see queries with especially long duration
log_slow_queries        = /srv/log/mysql_slow.log
long_query_time = 2
log-queries-not-using-indexes
```

Also not enabled by default a mysql error log might be useful.

```
log_error = /srv/log/mysql_error.log
```

EHLOVader:
This PR resolves #596

I enabled the error logs and also enabled the performance logging.
Mysql logs will be exposed through the /srv/log folder.

### DIFF
--- a/config/mysql-config/my.cnf
+++ b/config/mysql-config/my.cnf
@@ -64,9 +64,9 @@ max_allowed_packet	= 128M
 log_error = /var/log/mysql/error.log
 #
 # Here you can see queries with especially long duration
-#log_slow_queries	= /var/log/mysql/mysql-slow.log
-#long_query_time = 2
-#log-queries-not-using-indexes
+log_slow_queries        = /srv/log/mysql_slow.log
+long_query_time = 2
+log-queries-not-using-indexes
 
 [mysqldump]
 quick
--- a/config/mysql-config/my.cnf
+++ b/config/mysql-config/my.cnf
@@ -61,7 +61,7 @@ max_allowed_packet	= 128M
 #
 # Error log - should be very few entries.
 #
-log_error = /var/log/mysql/error.log
+log_error = /srv/log/mysql_error.log
 #
 # Here you can see queries with especially long duration
 log_slow_queries        = /srv/log/mysql_slow.log


cfoellmann:
do a `mysqld -help`.
Maybe we should go with:
```
slow_query_log = 1
slow_query_log_file = /srv/log/mysql_slow.log
long_query_time = 10
log_queries_not_using_indexes = 1
```
and we also (might) need to touch the log files and set permissions.
Any idea how to get stuff logged? Simulating slow query and/or errors?

EHLOVader:
I couldn't think of a way to test it last night either.. But i was sure
that the syntax was correct. I just uncommented what was there.

I also confirm the settings loaded by calling `SHOW VARIABLES`

Testing unused index shouldn't be to hard. And after some `SLEEP`. [I may
have found an answer to the long query problem.](
http://serverfault.com/a/285554/78197)

EHLOVader:
I couldn't think of a way to test it last night either.. But i was sure
that the syntax was correct. I just uncommented what was there.

I also confirm the settings loaded by calling `SHOW VARIABLES`

Testing unused index shouldn't be to hard. And after some `SLEEP`. [I may
have found an answer to the long query problem.](
http://serverfault.com/a/285554/78197)

cfoellmann:
Let's go with
```
slow_query_log      = 1
slow_query_log_file = /srv/log/mysql_slow.log
long_query_time     = 2
log_queries_not_using_indexes = 1
```
and
```sh
touch /srv/log/mysql_slow.log
touch /srv/log/mysql_error.log
```

EHLOVader:
There is definitely some permissions problems going on. The mysql user may not have access to the folder. I tried using the vagrant user and it isn't able to touch the files without sudo.

cfoellmann:
I will have a look tomorrow.
Was already wondering why sleep() did not show up in the log

EHLOVader:
I tried changing the user MySQL runs as for permissions to match the folder mapping and still can't get it to write to the logs. Error or slow queries.

Tried touching files, chmod, chown making mysql run as www-data.

Perhaps this is not possible.

cfoellmann:
There is no way to change ownership or permissions within a synced folder. SEE: http://jeremykendall.net/2013/08/09/vagrant-synced-folders-permissions/
But I still dont understand why mysql is not able to write to the log. Since the `log/` folder has the permission 0777 it should work, right?

cfoellmann:
@EHLOVader Can you update the PR to
my.cnf
```
log_error = /srv/log/mysql_error.log
#
# Here you can see queries with especially long duration
slow_query_log      = 1
slow_query_log_file = /srv/log/mysql_slow.log
long_query_time     = 2
log_queries_not_using_indexes = 1
```
for a start. This is the "current/new" syntax for mysql >= 5.5

https://github.com/EHLOVader/VVV/blob/mysql-logs/provision/provision.sh#L359
```sh
	touch /srv/log/mysql_slow.log
	touch /srv/log/mysql_error.log
```

I tried adding the `mysql` user to the `vagrant` group which didn't have any effect. But I am still not convinced that `select sleep(11);` results in a slow query log entry.

@jeremyfelt add a `in-progress` label, please

mikemcguire:
Ok I'll try to get a PR up tonight as I've already built it out, but I gotta run to a meeting. For posterity and understanding I'll post this here.
Ubuntu has a special security service called AppArmor that goes above regular file system permissions to lock down applications. It specifies what part of a file system any given application may have access to. 

You can see the default mysql configuration in this file
`/etc/apparmor.d/usr.sbin.mysqld`

There is an empty file generated for custom mysql configurations
`/etc/apparmor.d/local/usr.sbin.mysqld`

Adding these lines to the configuration and restarting AppArmor fixes the issue.
```
/srv/log/ r,
/srv/log/** rwk,
```

mikemcguire:
See https://github.com/Varying-Vagrant-Vagrants/VVV/pull/811/files

jeremyfelt:
Thanks for opening and working on this PR, @EHLOVader! I'm going to close it in favor of #811, which applies the AppArmor config.

PR #623, opened 2015-04-24
Update to latest vagrant versions

hubdotcom:
Is VVV only supported on 1.6.x or does this need updating?


cfoellmann:
WE should say >=1.6

hubdotcom:
Updated to alternative

jeremyfelt:
Sorry, missed this! I just updated the docs to say `1.7.x` for Vagrant. I think we're okay leaving it at the specific minor version, with patch version unspecified. I'm hesitant to show "or later" as we've been burned by new Vagrant releases before and I'd rather confirm first.

PR #625, opened 2015-04-24
Add default CPU cores

hubdotcom:
As per documentation:
http://docs.vagrantup.com/v2/virtualbox/configuration.html
http://parallels.github.io/vagrant-parallels/docs/configuration.html

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -12,7 +12,7 @@ Vagrant.configure("2") do |config|
   # Configuration options for the VirtualBox provider.
   config.vm.provider :virtualbox do |v|
     v.customize ["modifyvm", :id, "--memory", 1024]
-    v.customize ["modifyvm", :id, "--cpus", 2]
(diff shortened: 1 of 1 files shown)


cfoellmann:
I am not sure if we "can"/want to go with 2 cores. Some machines might be too weak to support the VM and proper performance for the host

hubdotcom:
In that case, should the VMware Fusion be adjusted to a single core too? It's good to have the setting explicit since users are aware they can modify this

khromov:
@hubdotcom You may be interested in the vagrant-faster plugin that can increase the specs automatically according to the performance specs of the host:
https://github.com/rdsubhas/vagrant-faster

hubdotcom:
nice recommendation!

hubdotcom:
Have adjusted the default cores to 1 on all installs

jeremyfelt:
Being explicit is good. I like the idea of sticking to 1 core by default for VirtualBox. I wonder if that impacted some of the speed results. :)

PR #629, opened 2015-04-24
Update version of phpMyAdmin

hubdotcom:
https://github.com/Varying-Vagrant-Vagrants/VVV/blob/master/provision/provision.sh#L593


cfoellmann:
lets remove the version here completely

hubdotcom:
Updated

PR #620, opened 2015-04-21
license link updated

michaelbeil:
should no longer 404

PR #630, opened 2015-04-24
Update broken license link and fix WP link

hubdotcom:
Addresses #620 and change make.wordpress.org to https (the link is not accessible over http)


michaelbeil:
I updated #620 and moved it to #643. Looks like you will need to rebase and squash this commit to include only the fix to the WP link.

jeremyfelt:
Looks like this was fixed elsewhere.

Issue #517, opened 2014-12-11
how to fix error "default: stdin: is not a tty" when using vagrant up


ricker07:
This is what i get when i run "vagrant up". i already do everything, line "sudo service mysql restart || true", vagrant hault, vagrant provision, sudo apt-get update, etc. but it doesn't work. please help me, i just want to update my mysql from 5.5 to 5.6 because i'm using laravel framework and the system i'm working on today is required to use mysql version 5.6.

Bringing machine 'default' up with 'virtualbox' provider...
==> default: Importing base box 'laravel/homestead'...
==> default: Matching MAC address for NAT networking...
==> default: Checking if box 'laravel/homestead' is up to date...
==> default: A newer version of the box 'laravel/homestead' is available! You currently
==> default: have version '0.2.0'. The latest is version '0.2.1'. Run
==> default: `vagrant box update` to update.
==> default: Setting the name of the VM: Homestead_default_1418261939885_87172
==> default: Clearing any previously set network interfaces...
==> default: Preparing network interfaces based on configuration...
    default: Adapter 1: nat
    default: Adapter 2: hostonly
==> default: Forwarding ports...
    default: 80 => 8000 (adapter 1)
    default: 3306 => 33060 (adapter 1)
    default: 5432 => 54320 (adapter 1)
    default: 22 => 2222 (adapter 1)
==> default: Running 'pre-boot' VM customizations...
==> default: Booting VM...
==> default: Waiting for machine to boot. This may take a few minutes...
    default: SSH address: 127.0.0.1:2222
    default: SSH username: vagrant
    default: SSH auth method: private key
    default: Warning: Connection timeout. Retrying...
    default: Warning: Remote connection disconnect. Retrying...
==> default: Machine booted and ready!
==> default: Checking for guest additions in VM...
==> default: Setting hostname...
==> default: Configuring and enabling network interfaces...
==> default: Mounting shared folders...
    default: /vagrant => /home/sedp/Homestead
    default: /home/vagrant/Projects => /home/sedp/Projects
==> default: Running provisioner: shell...
    default: Running: inline script
==> default: stdin: is not a tty
[...]

PR #633, opened 2015-04-28
fix-no-tty

cfoellmann:
fix suggested by @twisty7867 in #517

SEE: http://foo-o-rama.com/vagrant--stdin-is-not-a-tty--fix.html

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -186,6 +186,11 @@ Vagrant.configure("2") do |config|
     config.vm.synced_folder "www/", "/srv/www/", :owner => "www-data", :extra => 'dmode=775,fmode=774'
   end
 
+  config.vm.provision "fix-no-tty", type: "shell" do |s|
+    s.privileged = false
+    s.inline = "sudo sed -i '/tty/!s/mesg n/tty -s \\&\\& mesg n/' /root/.profile"
+  end
+
   # The Parallels Provider does not understand "dmode"/"fmode" in the "mount_options" as
   # those are specific to Virtualbox. The folder is therefore overridden with one that
   # uses corresponding Parallels mount options.


cfoellmann:
This works for me so far. Needs longer testing and more testers.

HongPong:
I tried putting this fix at line 212 and it doesn't seem to fix it, still get the error 2x on bootup, but I am new to vagrant and might be doing something wrong in the testing process here.

twisty7867:
@HongPong I'm curious to see your Vagrantfile and output. For me, it eliminates the error entirely *as long as it's the first provisioner that runs*. I am **obsessed** with getting zero red text.

HongPong:
sure, i think i tried that too (as in it was the first statement below 'provisioning' comments)
Please see https://gist.github.com/korgroup/ad904f8727627c1ee852
https://gist.github.com/korgroup/4d9cfc98deebc2b9adb6

However I also had a transient error where i could not go into the www folder when i was inside the vagrant shell, it spit out a bunch of ???? marks when i tried to ls the /srv/www . Maybe this went away when i made a new local shell window? 
It's not doing that www problem right now, but the errors are still occurring regardless. This is my first day with vagrant and it's pretty nifty but a little tricky to understand everything.

twisty7867:
Rather than cluttering up this PR with our discussion, I'd be happy to follow up with you via email if you want some help. I think the issue is that your VM was already provisioned and you just a `vagrant reload`. Run `vagrant provision` and you should see something like this in the output:
```
==> default: Running provisioner: fix-no-tty (shell)...
    default: Running: inline script
```
Once you see that, nevermore should you see the "not a tty" message. Depending on what your other provisioning scripts do, you might get a similar message from `dpkg-reconfigure`. The fix for that (and just a best practice in general) is to provide the answers up front using `dpkg-preconfigure` or `debconf-set-selections`.

HongPong:
thanks @twisty7867 that was the right advice, it did kick in and this patch seems to work quite well indeed.

jeremyfelt:
This is great stuff. Love getting rid of that error. :)

@twisty7867 Are you good with including this with VVV under the MIT License? It looks like the origin is straight from [your post](http://foo-o-rama.com/vagrant--stdin-is-not-a-tty--fix.html).

Thanks!

twisty7867:
Yes, feel free. I'll update my page to indicate that I consider the code snippet to be in the public domain.

jeremyfelt:
Great, thanks!

laite:
I'm getting an error with this commit when running `vagrant up` for the first time.

```
There are errors in the configuration of this machine. Please fix
the following errors and try again:

vm:
* The 'fix-no-tty' provisioner could not be found.
```

Vagrant 1.6.3 on Debian unstable (sid) x64. Building from current `master` branch works.

twisty7867:
Named provisioners require vagrant 1.7 or better. You can upgrade or remove the name and new type argument, leaving just the "shell" value instead.

michaelbeil:
VVV requires Vagrant 1.7.x: https://github.com/Varying-Vagrant-Vagrants/VVV/blame/develop/README.md#L55.

laite:
Thanks, it seems I noticed that too late. Sorry about that.

PR #644, opened 2015-05-08
Fix https WP links

hubdotcom:
Ref #630


jeremyfelt:
Read my mind earlier, thanks!

PR #646, opened 2015-05-14
fix typo on "helpful" header in readme

HongPong:
There was a typo in 'helpful' unless it's some obscure pun. Tiny change but reads better.


jeremyfelt:
Thanks @HongPong!

PR #656, opened 2015-05-31
Add support for VMWare Workstation

khromov:
The Vagrant VMWare Workstation plugin (for Windows) is compatible with Fusion images but you still have to specify `--provider=vmware_workstation`. However, the VVV Vagrantfile tries to load the default box in such case, so I added a conditional for the vmware_workstation provider so that it uses the Fusion box. 

Provisioned it, works without issues.


jeremyfelt:
Great, thanks @khromov!

PR #657, opened 2015-05-31
Fix phpMyAdmin permission issue on VMWare

khromov:
On VMWare, the default shared folder permissions are 777, which makes phpMyAdmin uncooperative, giving the following error message when trying to use it:

```
Wrong permissions on configuration file, should not be world writable!
```

This patch will disable permission checking in phpMyAdmin to get around this. The permission checking is a mechanism to avoid security issues on shared hosts, but since VVV is presumed to be a development environment this should cause no issues.

### DIFF
--- a/config/phpmyadmin-config/config.inc.php
+++ b/config/phpmyadmin-config/config.inc.php
@@ -140,3 +140,5 @@ $cfg['SaveDir'] = '';
  */
 
 $cfg['AllowUserDropDatabase'] = true;
+
+$cfg['CheckConfigurationPermissions'] = false;


jeremyfelt:
This is a little scary, but I think ok overall. I'm going to open another issue to document some of the possible security issues in VVV if configured with a public network.

Thanks @khromov!

jeremyfelt:
I went ahead and reverted this, only because it's still in the `develop` branch and has not been merged to `master`. Instead, we now have the ability to use a custom config file for phpMyAdmin via #688 that should be used for this instead. This prevents possible security issues on public networks.

Thanks again for the original ticket. :)

khromov:
@jeremyfelt Good call! :+1:

PR #665, opened 2015-06-08
bugfix/node-log-ignore

2ndkauboy:
ignore the npm-debug.log file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,6 @@
 # And a few of our web directories are important to share.
 /www/default/*
 !/www/default/index.php
+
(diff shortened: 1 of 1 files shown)


jeremyfelt:
Hi @2ndkauboy, thanks for the ticket!

Any ideas why the `npm-debug.log` file was created in the root directory? I don't think I've noticed that one yet and we don't have any local npm tasks that should put it there.

jeremyfelt:
I'm going to close this out pending reporter feedback. If this can be reproduced and is a file generated by VVV, we'll probably need a new pull request as this one is out of date.

PR #678, opened 2015-06-29
Remove duplicate line on Mailcatcher install

amchoukir:


jeremyfelt:
Good catch, thanks, @amchoukir!

PR #688, opened 2015-07-03
[PR] Allow for a custom phpMyAdmin configuration file

jeremyfelt:
If `config.inc.custom.php` exists in the `www/default/database-admin/` directory alongside the default `config.inc.php` file, it will be included so that our default configuration for phpMyAdmin can be overridden.

Props @grantnorwood for the initial approach in #588.


grantnorwood:
Thanks, Jeremy!  I appreciate all your hard work!

PR #691, opened 2015-07-03
[PR] Be explicit about defaults in XDebug configuration

jeremyfelt:
By being more explicit about the defaults currently configured for
XDebug, we can make it easier to submit future pull requests for
changes as only the values of individual configurations will be
changed.


jeremyfelt:
See #465 as well for inspiration.

PR #715, opened 2015-07-27
Allow symlinks in shared folders (in VirtualBox)

cfoellmann:
Not allowing symlinks in the first place is a security/stability safety. SInce this is for development it should be possible.

MIght need additional changes for Parallels + Fusion

---

New PR (replaces #592) to change destination branch to `develop`

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -14,6 +14,7 @@ Vagrant.configure("2") do |config|
     v.customize ["modifyvm", :id, "--memory", 1024]
     v.customize ["modifyvm", :id, "--natdnshostresolver1", "on"]
     v.customize ["modifyvm", :id, "--natdnsproxy1", "on"]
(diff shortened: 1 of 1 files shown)


cfoellmann:
scratch the feedback. doesnt work now

jeremyfelt:
Hi @cfoellmann, sorry I've never gotten around to testing this.

I'm okay having a solution only for VirtualBox if that's all we're able to support. Can you walk me through the best way to take advantage of this locally?

LoreleiAurora:
Symlinks that don't spread outside of a shared folder already work so personally i dont feel this adds anything we cant already do.

Rather than using symlinks i would use the local_dir in vvv-config.yml to move a site out of the www dir, and/or allow a Customfile in the site and set up another shared folder to map for example a plugin outside of www.

jeremyfelt:
I'm going to go ahead and close this out for now. If this is still needed, let's go ahead and discuss the problem and approach in a new issue and then go for another attempt.

Issue #725, opened 2015-08-19
Provision hangs on COMPOSER_HOME=/usr/local/src/composer composer self-update line


mariusvetrici:
Hello, 

today I ran into this issue: Upon reprovisioning the virtual machine, it started hanging up on this last message:
==> default: ack-grep already installed

I debugged the provision.sh script and seems that the script is actually hanging up on this line:
COMPOSER_HOME=/usr/local/src/composer composer self-update

I executed this line in the Ubuntu terminal (vagrant ssh) and it just hangs up - nothing happens.

I upgraded the virtual box and pulled the project from git, but still get the same issue.

Any clue to this?

Thank you in advance for any help.

mariusvetrici:
Seems that this issue fixes with this pull request:
https://github.com/Varying-Vagrant-Vagrants/VVV/pull/729

jeremyfelt:
Hi @mariusvetrici, thanks for the ticket and sorry for the delay in response.

I'm not able to reproduce this issue during re-provisioning at this point. I don't think I want to merge #729 in unless we're sure it resolves the issue at hand. Provisioning already runs as root, so sudo should not be necessary here.

I'm going to leave this open for now pending steps to reproduce. It's possible some other update has helped to resolve the issue.

PR #729, opened 2015-08-20
Fix composer selfupdate error

mariusvetrici:
The provision script tries to update composer through 'composer self-update' which throws this error:
[Composer\Downloader\FilesystemException]
Filesystem exception:
Composer update failed: the "/usr/local/bin/composer" file could not be written

Fix self update by prefixing all the calls to composer with sudo.

Commit refers to issue https://github.com/Varying-Vagrant-Vagrants/VVV/issues/725

### DIFF
--- a/provision/provision.sh
+++ b/provision/provision.sh
@@ -220,13 +220,13 @@ if [[ $ping_result == "Connected" ]]; then
 	# the master branch on its GitHub repository.
 	if [[ -n "$(composer --version --no-ansi | grep 'Composer version')" ]]; then
 		echo "Updating Composer..."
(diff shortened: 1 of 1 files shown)


prio101:
I am having this same issue .

slaFFik:
`sudo` might fix the problem :)

veeruborra:
sudo /usr/local/bin/composer self-update

jeremyfelt:
Thanks for the work on this PR, @mariusvetrici. I'm going to close this pending some further feedback on #725 as I have not been able to reproduce the issue locally. Once we can determine a set of reproducible steps, then we can determine the right approach here.

PR #731, opened 2015-08-21
Remove VVV Site Wizard from README, as it is no longer maintained.

bradp:


michaelbeil:
Great PR @bradp!

jeremyfelt:
Sad, but true. Happy to add it back in if maintenance fires back up. Thanks @bradp!

PR #734, opened 2015-08-25
Iron out wording of introductory note.

ericandrewlewis:
The introductory note is a bit repetitive — tried to boil this down to the essence.


jeremyfelt:
I dig it. Reduction + clarification = win. Thanks!

PR #738, opened 2015-08-28
Avoid browser warnings for "site URL does not match SSL certificate."

### LINKED ISSUE #737: TLS certificate does not specify common names, causing browser warnings
Weston added [HTTPS support](00d3a3e9068467c8363888a22fc687935d20bc36) a while ago :smile: Although default sites are still configured for HTTP (i.e. [`siteurl` and `home` are defined a la `http://src.wordpress-develop.dev`](https://github.com/Varying-Vagrant-Vagrants/VVV/blob/v1.1/provision/provision.sh#L470)), a user can change a site's options and HTTPS will work.

Sort of. Browsers will show warnings because the certificate authority is invalid and the site's URL does not match the certificate.

The certificate authority is invalid because this is a self-signed certificate, so we are the authority. A user can fix this browser warning by accepting the certificate. How this happens depends on the browser and OS. e.g. in OS X a user can add and trust the certificate to the Keychain Access application, and Chrome will respect this setting. However Firefox requires the certificate to be accepted (via a "security exception") in Firefox.

The site's URL does not match the certificate because when generating the cert a Common Name (`CN`) is not specified in the subject. 

We should consider stipulating the common names of all the default sites when creating the cert to avoid this browser warning.

I'm able to do this via the following:

```sh
openssl genrsa -out /etc/nginx/server.key 2048
openssl req -new -x509 \
            -key /etc/nginx/server.key \
            -out /etc/nginx/server.crt \
            -days 3650 \
            -subj /CN=*.wordpress-develop.dev/CN=*.wordpress.dev/CN=*.vvv.dev/CN=*.wordpress-trunk.dev
```

ericandrewlewis:
Fixes #737

### DIFF
--- a/provision/provision.sh
+++ b/provision/provision.sh
@@ -258,19 +258,19 @@ else
 	echo -e "\nNo network connection available, skipping package installation"
 fi
 
-# Configuration for nginx
+# Create an SSL key and certificate for HTTPS support.
 if [[ ! -e /etc/nginx/server.key ]]; then
 	echo "Generate Nginx server private key..."
 	vvvgenrsa="$(openssl genrsa -out /etc/nginx/server.key 2048 2>&1)"
 	echo "$vvvgenrsa"
 fi
-if [[ ! -e /etc/nginx/server.csr ]]; then
-	echo "Generate Certificate Signing Request (CSR)..."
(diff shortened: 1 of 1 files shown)


jeremyfelt:
Excellent. Thanks!

PR #739, opened 2015-09-01
Allow external access only when coming from host machine

westonruter:
Since the Vagrant host machine has the [IP address of `192.168.50.1` as the gateway](https://github.com/Varying-Vagrant-Vagrants/VVV/blob/8098632e255d05cea2dfc03e512a8a86cffe891f/Vagrantfile#L97-L111), this IP address can be explicitly supplied when creating the `external` user, instead of a wildcard `%`. This seems more secure.

### DIFF
--- a/database/init.sql
+++ b/database/init.sql
@@ -12,4 +12,4 @@ GRANT ALL PRIVILEGES ON `wordpress_unit_tests`.* TO 'wp'@'localhost' IDENTIFIED 
 
 # Create an external user with privileges on all databases in mysql so
 # that a connection can be made from the local machine without an SSH tunnel
-GRANT ALL PRIVILEGES ON *.* TO 'external'@'192.168.50.4' IDENTIFIED BY 'external';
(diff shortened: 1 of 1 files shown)


westonruter:
Or maybe this is unnecessary because it is a `private_network`?

johnpbloch:
@westonruter It's only private through the default network interface. Anybody can add public interfaces in the Customfile, and some providers don't allow any control over network restrictions. This would require some extra work to get it to play nicely with hyperv. Vagrant doesn't manage hyperv networking at all, so a private network can never be assumed, nor can the host's address. For example, my hyperv boxes get a DHCP lease directly from my router. This means the host machine could be connecting from any address in the router's range of addresses that it leases.

jeremyfelt:
Yeah, I think we should keep it at `%`. In a perfect world, we would detect the right IP address to use here, but we can't make many assumptions.

Maybe it would be worth it to have a security doc with a guide to how exposed your VM may or may not be depending on your config.

PR #743, opened 2015-09-09
Add Adapter and IP list to default page

### LINKED ISSUE #730: Create new page to display public_network IP address
Lots of great discussion in #566 helped me understand how to use the **public_network** property properly.

@jpry's notes about `vagrant ssh` / `ifconfig` to determine the public_network IP address are great

I'm wondering if we could create a page at something like http://vvv.dev/public-ip/ which could display either:

> This site is not available on any public IP, see how to configure public_network

or

> Your public IP is: <display IP address here>

I'm happy to look into this but I wanted to float the idea first.

salcode:
Add adapters with assigned IP addreses to output on the default page

Fixes #730

### DIFF
--- a/www/default/index.php
+++ b/www/default/index.php
@@ -10,12 +10,18 @@ if ( file_exists( 'dashboard-custom.php' ) ) {
 }
 
 // Begin default dashboard.
+
+// Allow cross origin access, for Ajax check for IP addresses.
+header( 'Access-Control-Allow-Origin: *' );
 ?>
 <!DOCTYPE html>
 <html>
 <head>
 	<title>Varying Vagrant Vagrants Dashboard</title>
 	<meta name="viewport" content="width=device-width, initial-scale=1.0">
+	<style>
+		.hide-me { display: none; }
+	</style>
 </head>
 <body>
 <ul class="nav">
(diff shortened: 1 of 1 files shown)


salcode:
I wanted to keep this first pass simple.

I think it makes sense to filter out the __eth0__ adapter since on my Mac with Virtualbox, the relevant IP addresses are always assigned to __eth1__ or higher.

jeremyfelt:
Oooh, slick. Sorry for taking 27 days to see this. :)

I'm going to mess around with it, but I like the idea.

salcode:
@jeremyfelt Great, I'd love another set of eyes on this and some fresh ideas.  Thanks.

salcode:
I've updated this PR to output a list of the IPs and then test each one with an AJAX call.

While this will detect localhost (127.0.0.1), it does not detect your local IP address (e.g. 192.168.50.1).  

The User's local IP address is particularly relevant if you're using Port Forwarding rather than the Public Network option.  It appears you can detect your local IP address from the browser in Chrome and Firefox, see http://net.ipcalf.com/, which would be great but adds complexity.  I think it makes sense to get this base functionality first and then look at adding local IP address detection in a future commit.

tomjn:
I like this idea too, but it conflicts with the new VVV 2 changes :(

tomjn:
I'm closing this out as the dashboard was moved to a separate repo

PR #659, opened 2015-06-01
Feature/rewrite

cristovaov:
Hi there,

I want to submit the following enhancements to VVV to the provisioning script. 
I believe putting everything in functions allows for better structure and modularity/flexibility of the provisioning flow as certain steps can more easily be optional.

Other changes include:
- double quoting of values/arguments for better readablitiy
- simplified output of the installed/not installed packages
- reviewed function to run as vagrant user

Along commits, the script has been checked against shellcheck and incorporates latest changes (a02654c1b94457de6808b3d1b96ea082f6ffb1fe). It didn't give me any merge conflicts locally from feature branch to develop branch.

By uncommenting line 730 and line 776, you can run the script with bash debugging features on.

I'd also would like to discuss further enhancements at a later point.

Cheers,
-Cris.

Issue #474, opened 2014-10-13
Allow selecting which sites to make active


tnorthcutt:
I keep a number of sites (~20 right now) on one VVV install. Most of those, however, are only around for the rare instance when I need to work on them, and 90%+ of the time, I don't need them available. To speed up `destroy`, `halt`, `up`, `provision`, etc. it would be great to have a way to indicate which sites should be loaded at any given time. Perhaps a simple text file that `provision.sh` can look for, and if found, would then use to load only the sites explicitly defined?

I'm not sure if that's the best approach (and if it is, how exactly to implement that), but if given some direction, I'd be happy to submit a PR to implement this functionality.

jeremyfelt:
It is now possible to selectively enable and disable sites through a YAML config file after the merge of #980. :tada:

PR #756, opened 2015-10-04
Feature/wp split

cristovaov:
Moved WP-CLI and WP installs to a separate script. As the installs are performed not as root during provisioning, moving them to a separate script seems to be more straightforward. Provision rule without privileges in Vagrantfile added. WP Develop was made optional by comment.


jeremyfelt:
Hi @cristovaov, thanks for the PR!

Let's focus on handling the provision rewrite through #659 first and then see about splitting out the WP functions through something like #474.

I'm going to make an attempt to finish #659 off this week.

PR #759, opened 2015-10-07
Fix the HyperV Provider

### LINKED ISSUE #758: vagrant up Fails with SMB Error
On Mac OSX 10.11 (El Capitan), fresh install of Vagrant, fresh install of  VVV, I ran into an error trying to start VVV. 

`vagrant up` produces this error. 

```
SMB shared folders are only available when Vagrant is running
on Windows. The guest machine can be running non-Windows. Please use
another synced folder type.
```
After digging through the VagrantFile I commented line 232 that sets up smb for the file system which fixes it and am able to start VVV. 

```
if ! options[:type]
    #options[:type] = "smb"
end
```

I tried to install Samba, but with no avail, still haven't figured that out yet.

ericmann:
Though the SMB config is wrapped in a conditional statement, Vagrant's built-in Ruby interpreter seems to be ignoring scope here. Using `config` in the scope shouldn't be an issue, but apparently is. Instead we'll use the explicitly-scoped `override` version to make sure nothing bubbles out unexpectedly.

Fixes #758

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -227,7 +227,7 @@ Vagrant.configure("2") do |config|
   config.vm.provider :hyperv do |v, override|
     override.vm.synced_folder "www/", "/srv/www/", :owner => "www-data", :mount_options => ["dir_mode=0775","file_mode=0774","forceuid","noperm","nobrl","mfsymlinks"]
     # Change all the folder to use SMB instead of Virtual Box shares
-    config.vm.synced_folders.each do |id, options|
+    override.vm.synced_folders.each do |id, options|
       if ! options[:type]
         options[:type] = "smb"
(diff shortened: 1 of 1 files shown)


ericmann:
Note, I've verified this on both Mac with VirtualBox and Windows with Hyper-V. It resolves the provisioning error on Mac without blocking provisioning on Windows.

lkwdwrd:
Tested now on both Windows with Hyper-V and my mac with Virtual Box. Normally I would let @jeremyfelt  merge, but this is causing too many issues.

jeremyfelt:
:+1: Thanks for catching this @ericmann and @lkwdwrd!

PR #769, opened 2015-10-22
Allow opcache to be easily enabled/disabled

thebrandonallen:
As it stands, OPCache doesn't even work since the module isn't loaded. Here we load the extension, and move the opcache.ini to the `mods-available` folder. This allows it to be enabled/disabled from the command-line, without a having to reprovision. I've also added opcache_(on/off) scripts, but these may not be wanted, as OPCache will likely be switched on/off less than XDebug.

### DIFF
--- a/config/homebin/opcache_off
+++ b/config/homebin/opcache_off
@@ -1,0 +1,3 @@
+#!/bin/bash
+sudo php5dismod opcache
+sudo service php5-fpm restart
--- a/config/homebin/opcache_on
+++ b/config/homebin/opcache_on
@@ -1,0 +1,3 @@
+#!/bin/bash
+sudo php5enmod opcache
+sudo service php5-fpm restart
--- a/config/php5-fpm-config/opcache.ini
+++ b/config/php5-fpm-config/opcache.ini
@@ -1,3 +1,7 @@
+; configuration for php ZendOpcache module
+; priority=05
+zend_extension=opcache.so
+
 [opcache]
 ; Determines if Zend OPCache is enabled
 opcache.enable=1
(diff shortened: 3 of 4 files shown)


cfoellmann:
:+1:

jeremyfelt:
Thanks for the ticket and patch, @thebrandonallen. 

OpCache should be enabled by default. The organization of the config files is a little strange, but the extension is loaded in `/etc/php5/fpm/conf.d/05-opcache.ini` and the custom config is in `/etc/php5/fpm/conf.d/opcache.ini`. You can see OpCache in action at http://vvv.dev/opcache-status/opcache.php, where a list of cached scripts and hit/miss statistics can be seen.

I would be happy if there was a nicer way to enable this. It may be that the `opcache.ini` should live somewhere else. The `05-opcache.ini` was auto-created by PHP (I think), and I recall there being some problems when shifting from PHP 5.4 to 5.5. I may also be recalling incorrectly. :)

I don't think we need a flag to turn it off/on. I'm happy to be swayed there, but I don't think I've ever run into a case where it would be beneficial to turn it off.

thebrandonallen:
@jeremyfelt 

Everything you said is correct. The problem I was having at the time I wrote the patch, was that I had just destroyed and re-provisioned my VVV install, and upon visiting `opcache-status` I saw that OPCache wasn't active. PHP 5.5 activates OPCache by default by adding a file to `mods-available` with the contents setting loading priority to 5, and loading the opcache.so Zend module. The Debian commands `php5enmod`/`php5dismod` had no effect, and I had to load the OPCache Zend module in VVV's opcache.ini  to get it to load.

So, I wrote the above patch. Comfirmed it worked on an existing install. I destroyed and re-provisioned without the patch, and confirmed OPCache was disabled. I then applied the patch, provisioned, and confirmed it worked. Once more I destroyed, then re-provisioned with the patch, and confirmed OPCache was enabled.

After all of that, I can't reproduce the original issue. After two destroys, OPCache is enabled every time, and can be enabled/disabled with the commands `php5enmod`/`php5dismod`. Before I wrote the patch, the commands weren't doing their job.

I'm going to leave this open for now in case someone wants to champion the opcache_on/off flags, but since turning on/off OPCache is not something I even do that often, running two shell commands as opposed to one is pretty inconsequential. They were pretty much copy and paste, I figured "why not?"

jeremyfelt:
Going to close this out as we haven't had a repeat issue. Please feel free to re-submit the proposal if it comes up again in the future.

PR #781, opened 2015-11-03
Simplify private network overrides

johnpbloch:
This is ostensibly an improvement for hyperv providers, but carries extra benefits for all providers.

Hyper-V does not know how to make private networks, so the extra IP address configured by VVV is not needed (and when used with vagrant ghost, results in duplicate hosts entries that point to a dud address). This unsets the private network for the hyperv provider.

Additionally, this makes it much easier and more forwards-compatible to modify the VM ip address. Currently, if you need to change the vm's network (e.g. because your network is using the 192.168.50.* subnet), you must edit `Vagrantfile` directly. With an id for the network config, the ip can be overridden in the `Customfile` in one line:

```rb
config.vm.network :private_network, id: "vvv_primary", ip: "10.10.10.10"
```

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -119,7 +119,7 @@ Vagrant.configure("2") do |config|
   # should be changed. If more than one VM is running through VirtualBox, including other
   # Vagrant machines, different subnets should be used for each.
   #
-  config.vm.network :private_network, ip: "192.168.50.4"
+  config.vm.network :private_network, id: "vvv_primary" ip: "192.168.50.4"
 
   # Public Network (disabled)
   #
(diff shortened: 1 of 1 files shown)


jeremyfelt:
> With an id for the network config, the ip can be overridden in the Customfile in one line:

What! That's crazy talk. I had no idea the `id` parameter existed. Thanks!

johnpbloch:
@jeremyfelt I know, right?? I didn't know either until this morning when I was trying to get this working another way! It took me an hour digging through undocumented ruby in the vagrant repo to find that little gem! [Proof!](https://github.com/mitchellh/vagrant/blob/73edde73565144653ca8cceadd12e9885909d8f2/plugins/kernel_v2/config/vm.rb#L243)

PR #783, opened 2015-11-06
Set the WP_CORE_DIR path so phpunit tests are run against WP trunk

danielbachhuber:
Otherwise, test install scripts will default to using WordPress latest, which is incompatible with WP trunk tests.

From https://github.com/wp-cli/wp-cli/issues/2129#issuecomment-154356360


jeremyfelt:
:+1: Thanks @danielbachhuber!

PR #797, opened 2015-12-07
Update README.md

jonathanbossenger:
Added recommendation to run cmd window as Administrator, this was causing me an issue when trying to write to the hosts file


jeremyfelt:
Thanks @jonathanbossenger. I agree that this is a tricky spot and deserves a note. I'll merge this in via #871 with some other README changes as I've let this PR get out of date.

PR #805, opened 2015-12-28
Adding PHP FPM status page

f4bsch:
The PHP FPM status page lists active requests, this is useful to debug running wp-crons.
Few people might use it, its requires only these few changes.
I'm using it to track concurrent cronjob workers that re-spawn themselves to process big data in background.


jeremyfelt:
Good stuff, I dig it. Thanks @f4bsch!

PR #806, opened 2015-12-29
Provisioner names

coreymckrill:
Problem: Working on a **provision-post.sh** script, every time you want to test it, you have to run _all_ the provisioners.

Solution: Use the `--provision-with` switch to specify which provisioner(s) to run. Per the Vagrant [documentation](https://docs.vagrantup.com/v2/cli/provision.html):

> `--provision-with x,y,z` - This will only run the given provisioners. For example, if you have a `:shell` and `:chef_solo` provisioner and run `vagrant provision --provision-with shell`, only the shell provisioner will be run.

Since all of VVV's provisioners are `:shell`, this doesn't help much. However, if each one is named, you can do something like this: `vagrant provision --provision-with post`


jeremyfelt:
Thanks for the PR @coreymckrill, and for the patience. :) I've looked at this a 100 times. Let's do it!

coreymckrill:
:boom:

PR #808, opened 2016-01-06
Update phpunit to 4.8.x

coreymckrill:
The latest version is actually 5.1.x, but it requires PHP 5.6+. Updating to 4.8.x at least gets us to a version that has [available documentation](https://phpunit.de/manual/4.8/en/index.html).

I ran the WordPress test suite with this update and it still completes successfully. I'm not sure what other testing needs to be done to vet this change.


jeremyfelt:
Thanks @coreymckrill!

PR #813, opened 2016-01-19
update deprecated `parallels.optimize_power_consumption = false`

eriktrom:
- use `parallels.customize ["set", :id, "--longer-battery-life", "off"]` in its place


jeremyfelt:
Thanks @eriktrom!

PR #815, opened 2016-01-22
Switch from hostsupdater to ghost for hosts

morganestes:
Based on discussion in #606, and inspired by a (recently fixed)
[bug in vagrant-hostsupdater](https://github.com/cogitatio/vagrant-hostsupdater/commit/8b464369016d880dba4936428424c038042488f7) that removes hosts unexpectedly, I would
like to see vagrant-ghost recommended as the plugin for managing hosts.

The two are slightly different in configuration, but this initial change
can keep the existing logic and simply swap out the config settings names
and remove options that don't exist in ghost.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -106,6 +106,13 @@ Vagrant.configure("2") do |config|
 
     return hosts
   end
+
+  if defined?(VagrantPlugins::Ghost)
+    # Pass the found host names to the Ghost plugin.
+    config.ghost.hosts = get_vvv_hosts
+  elsif defined?(VagrantPlugins::HostsUpdater)
+    config.hostsupdater.aliases = get_vvv_hosts
+    config.hostsupdater.remove_on_suspend = true
   end
 
   # Private Network (default)
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -78,17 +78,21 @@ Vagrant.configure("2") do |config|
 
   # Local Machine Hosts
   #
-  # If the Vagrant plugin hostsupdater (https://github.com/cogitatio/vagrant-hostsupdater) is
-  # installed, the following will automatically configure your local machine's hosts file to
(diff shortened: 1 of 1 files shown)


cfoellmann:
How about using conditional logic to support both?

morganestes:
@cfoellmann Can you clarify what you mean by supporting both? We can't have both of them operating at the same time since they would just fight over the hosts file.

I suppose we could have an `if…else` block that checks for both plugins, but which one would take precedence? We'd have to stop processing with the first one we list that's installed, and if someone has them both installed we'd be making the decision for them to use the one we list first.

cfoellmann:
Like you said. `if...else...end`
Use ghost if present but fallback to the hostsupdater. But we need this to not break the installs that only update.

morganestes:
Makes sense to not make a breaking change when it's not necessary.

I extracted the code to grab the hosts into its own function to maximize reuse, then use it in the conditional to pass the array to each plugin.

lkwdwrd:
I just merged this into my windows install to test since I'm using Hyper-V which doesn't support the static IP. Ghost should support the dynamic IP provided Hyper-V where I was using a modified version of the hostsupdater plugin before. I will report back on the success of the test when I have some data, and if I forget, please ping me :).

lkwdwrd:
@morganestes - needs a bit of tweaking to get proper paths. We could make this function a little more general for reading files by lines, meaning Customfile users could make more interesting use of it, but here is a [simple fix](https://github.com/lkwdwrd/varying-vagrant-vagrants/commit/a0467b55c1972944a0db1e28760360a1b5d26200). I tried to PR your branch from mine, but it wont let me select your fork as the base branch for some weird reason.

jeremyfelt:
Thanks for all of the work on this PR. I'm going to close it for now as it needs some work to be mergeable and the original concerns that initiated #606 have been addressed.

PR #840, opened 2016-02-29
possible false positive in package_check

keegoid:
### commit comment

use apt-cache policy instead of dpkg -s to check if pkg is NOT installed
sometimes dpkg -s will find Version information for packages that have
been removed, resulting in a false positive

### reasoning

1. If we check that a package exists and get a false positive, that package will never get installed.
1. If we check that a package doesn't exist and get a false positive, apt-get will attempt to install it and notice that it is already installed.

#### remove virtualbox-5.0 and check if it still exists
```bash
sudo apt-get remove virtualbox-5.0
dpkg -s virtualbox-5.0 2>&1 | grep 'Version:'
```

This results in a false positive due to 'Version:' being found in several places.

```bash
Version: 5.0.14-105127~Ubuntu~trusty
Config-Version: 5.0.14-105127~Ubuntu~trusty
Python-Version: 2.7
```

#### check with apt-cache policy instead

```bash
apt-cache policy virtualbox-5.0
```

The result is not a false positive.

```bash
virtualbox-5.0:
  Installed: (none)
  Candidate: 5.0.14-105127~Ubuntu~trusty
```

### DIFF
--- a/provision/provision.sh
+++ b/provision/provision.sh
@@ -159,15 +159,15 @@ package_check() {
   local package_version
 
   for pkg in "${apt_package_check_list[@]}"; do
-    package_version=$(dpkg -s "${pkg}" 2>&1 | grep 'Version:' | cut -d " " -f 2)
-    if [[ -n "${package_version}" ]]; then
+    if [[ -n "$(apt-cache policy ${pkg} | grep 'Installed: (none)')" ]]; then
+      echo " *" $pkg [not installed]
+      apt_package_install_list+=($pkg)
+    else
+      package_version=$(dpkg -s "${pkg}" 2>&1 | grep 'Version:' | cut -d " " -f 2)
       space_count="$(expr 20 - "${#pkg}")" #11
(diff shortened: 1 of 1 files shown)


keegoid:
This may be even more reliable than `apt-cache policy` and `dpkg -s`:

```bash
program_exists() {
   local ret='0'
   command -v $1 >/dev/null 2>&1 || { local ret='1'; }

   # fail on non-zero return value
   if [ "$ret" -ne 0 ]; then
      return 1
   fi

   return 0
}
```

It doesn't result in a false positive after virtualbox-5.0 is removed.

keegoid:
Perhaps a combination of the two will work better.

```bash
if [ "$(dpkg -s ${1} 2>&1 | grep 'Version:')" ]; then
   [ -n "$(apt-cache policy ${1} | grep 'Installed: (none)')" ] && return 0 || return 1
else
   return 0
fi
```

jeremyfelt:
Thanks for the excellent pull request, @keegoid. This makes sense.

Issue #834, opened 2016-02-21
Upgrade PHP to 5.6.x or 7.0.x


jeremyfelt:
We're way behind right now with PHP 5.5.x. We should do better.

So what's the right answer? I lean toward PHP 7, but am definitely willing to hear reasons for sticking with PHP 5.6.

cfoellmann:
I would love to have php 7 as the default.

GaryJones:
PHP 7, but with phpbrew (another ticket already open) for those who wish to change down per domain to match production setups. Even if someone isn't using any of the new features from PHP 7, there's no harm in having a faster localhost.

itsazzad:
Better if would be selectable per project. Like project X with PHP5.6, Y with PHP7.

slaFFik:
Agree with @itsazzad : selectable PHP version.

16nsk:
Agree with selectable. Only supported versions. Once 5.5 is off, it should not be present in VVV as well.
Who needs older versions could always just checkout a previous version of VVV that supported that particular version.

jjeaton:
Another vote for phpbrew. While I would love PHP7 as the default, it'll be a long time before the production environments for local sites are all up to PHP7.

LoreleiAurora:
Another vote for PHP7 with phpbrew

neilgee:
Selectable version would be great

rowatt:
As much as I would love to have access to PHP7 in VVV, I think that sticking with PHP5.5 might be the right answer until there's a way to make it user selectable (ideally per project, but for the whole VVV install would be OK).

Why?

If I'm developing for a site which is PHP5.5 in production, then running PHP7 (or 5.6) in VVV isn't going to pick up any issues relating to my code (or plugins, libraries etc) using PHP5.6 or PHP7 features. While I am sure nobody here (except me :-O) ever pushes code directly from VVV on localhost to production, having things not work on a staging environment but work perfectly on VVV is still a pain to debug.

On the other hand, if a host is PHP5.6 or PHP7, it's far less likely to cause problems if I've developed and tested on PHP5.5. Not impossible but less likely I think.

The ideal is for VVV to match a production environment as closely as possible, but until that's possible with a choice of PHP version in VVV, I'd suggest that being conservative is the right way to go (alas).

GaryJones:
Let me offer a counter-argument to rowatt's case.

[...]

PR #844, opened 2016-03-08
PHP7.0

LoreleiAurora:
Updates PHP from 5.5 to 7.0 as per #834


jeremyfelt:
Thanks @LoreleiAurora! This is great. I'm going to do some testing before merging, but things look like they're on the right track.

jeremyfelt:
This worked out **really** well. I'm going to merge now and then adjust a couple things. I think we can switch to a more generic `php-config` or `php-fpm-config` directory in the VVV structure to make future upgrades slightly better.

PR #846, opened 2016-03-10
Allow unit test DB config to be supplied via environment vars

westonruter:
Based on our conversation here: https://twitter.com/jeremyfelt/status/707751217187819521

With this in place, you'd be able to `brew install mysql`, create the `wordpress_unit_tests` database:

```sql
CREATE DATABASE IF NOT EXISTS wordpress_unit_tests CHARACTER SET utf8;
```

And then add the following lines to your host machine's `.bash_profile`:

```php
export WP_TESTS_DB_HOST=localhost
export WP_TESTS_DB_NAME=wordpress_unit_tests
export WP_TESTS_DB_USER=root
export WP_TESTS_DB_PASSWORD=''
```

Assuming you also have PHPUnit installed (`brew install phpunit`), you show now be able to run `phpunit` from a terminal on the host machine and execute the unit tests as fast as possible.

### DIFF
--- a/config/wordpress-config/wp-tests-config.php
+++ b/config/wordpress-config/wp-tests-config.php
@@ -23,14 +23,22 @@ define( 'WP_DEBUG', true );
 // These tests will DROP ALL TABLES in the database with the prefix named below.
 // DO NOT use a production database or one that is shared with something else.
 
-define( 'DB_NAME', 'wordpress_unit_tests' );
-define( 'DB_USER', 'wp' );
-define( 'DB_PASSWORD', 'wp' );
-if ( file_exists( '/vagrant' ) ) {
+if ( getenv( 'WP_TESTS_DB_HOST' ) ) {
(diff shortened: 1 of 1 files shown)


westonruter:
@jeremyfelt in my testing, I found that running the entire test suite inside vagrant took around 3 minutes, whereas running the test suite completely from host machine took 2.5 minutes. Does this difference in performance align with what you've seen?

jeremyfelt:
@westonruter Hmm. I think the WP core tests have gotten quite a bit quicker. The last time I ran them inside the VM, it took ~8 minutes while my local environment was around 3 minutes. Just now though, the VM tests took less than 2 minutes. I think I might be going crazy.

westonruter:
I just published a writeup comparing running PHPUnit entirely in the VM, PHP on host but DB in VM, and then entirely on host without VM at all: https://make.xwp.co/2016/03/10/running-phpunit-tests-outside-vvv/

PR #850, opened 2016-03-10
Fixes #849: Correctly find `vvv-init.sh` scripts with spaces in path.

### LINKED ISSUE #849: Spaces in path to `vvv-init.sh` scripts cause provisioning to fail
If a custom site has a space in the path to its `vvv-init.sh` script, such as, for instance `/srv/www/a space/vvv-init.sh`, then the `provision.sh` script fails to load it. This means the site cannot be provisioned with `vagrant provision`.

meitar:


jeremyfelt:
Thanks for the pull request, @meitar!

PR #635, opened 2015-04-30
Add PHPBrew to VVV

ndisidore:
## Why
This is my initial attempt at adding PHPBrew to VVV. Over the course of development, there were times where I more closely wanted to mimic our production environment, which may not be the latest version of PHP, may or may not have memcached installed, etc. Rather than just install a single version of PHP, I stumbled upon PHPBrew and thought it would be beneficial to integrate into VVV. PHPBrew allows you to easily build php versions on the fly, and turn on/off extensions like memcached, imagemagic, and APC.

More information on PHPBrew, along with it's command list, can be found here: https://github.com/phpbrew/phpbrew

## How to use
This installs phpbrew to /opt/phpbrew and registers a new binary, `phpswap`, that is a wrapper for phpbrew functionality. Switching PHP versions is as simple as issuing a single command from terminal
`phpswap 5.3.28`
This looks for the version installed locally, and, it doesn't find it, grabs the source and builds from scratch with the set list of extensions.
Restart the fpm process for the currently running version of PHP with `phpswap fpm-restart`
You can switch the version persist across reboots by using the -s flag, e.g. `phpswap -s 5.3.28`
Switch back to the system version with `phpswap system`.

## Notes
* Even though the script executes `phpbrew use/switch $PHP_VER`, it doesn't seem to register unless you run the command separately, e.g. `phpswap 5.5.20 && phpbrew use 5.5.20` or `phpswap -s 5.5.20 && phpbrew switch 5.5.20`
* After enabling/disabling extensions, issue `phpswap fpm-restart` for the changes to take effect on the web server

PR #854, opened 2016-03-18
PHPBrew + PHP7

schlessera:
Update to #635 :

This is an updated PR for `feature/phpbrew`, to integrate latest changes to `develop` and make it work with PHP7.


cfoellmann:
@jeremyfelt I would like to blind merge this for testing

GaryJones:
Usage here is as simple as SSH-ing into vagrant, then `phpswap -s 7.0.4`. 

`phpswap` appears to be a wrapper for `phpbrew`. If phpbrew doesn't have a build for the chosen version, then it will automatically build it.

However, @schlessera, I'm getting the following error on a freshly provisioned system (VirtualBox 5.0.16, vagrant 1.8.1, this PR branch):

screen shot 2016-03-28 at 18 06 38

`php -v` also still shows 5.5.9, so the OS version of PHP has not changed, as well as not telling FPM to use it.

schlessera:
@garyj This happens when a non-phpbrew version of PHP is active. The script normally checks for the PHP version and only kills the relevant php-fpm. However, with the non-phpbrew version, it is not able to do this.

A possible solution would be to killall php-fpms, but I did not want to do this for this pull request, as this issue is not relevant to PHP7, but rather a genetal weakness of the phpswap script.

Just do a "killall php-fpm", and then you'll be able to switch versions.

GaryJones:
Needed `sudo killall php-fpm`, but then `phpswap -s 7.0.4` (which by now was already built) worked straight away.

`php -v` still gives 5.5.9, but the local sites are being delivered with 7.0.4 :+1: I did a `phpbrew switch php-7.0.4` to tell the OS which PHP should be default.

jeremyfelt:
> I would like to blind merge this for testing

@cfoellmann +1 I'd be okay with you merging this into the feature branch whenever you'd like. I'll pay closer attention soon. :)

ocean90:
I'm getting a 502 error when I try to access the sites after switching the PHP version. These were my steps:

```
$ git clone git@github.com:Varying-Vagrant-Vagrants/VVV.git
$ git fetch origin pull/854/head:phpbrew
$ git checkout phpbrew
$ vagrant up
$ vagrant ssh
$ phpswap -s 7.0.8
$ phpbrew switch php-7.0.8
```

`$ ps aux | grep php` returns

```
root     13023  0.0  0.6 192208  6424 ?        Ss   10:15   0:00 php-fpm: master process (/opt/phpbrew/php/php-7.0.8/etc/php-fpm.conf)
www-data 13024  0.0  0.5 192208  5788 ?        S    10:15   0:00 php-fpm: pool www
www-data 13025  0.0  0.5 192208  5788 ?        S    10:15   0:00 php-fpm: pool www
````

Running `sudo service php5-fpm start` makes the sites available, but with PHP 5.5.9. What am I missing here?

```
$ ps aux | grep php
root     13023  0.0  0.6 192208  6424 ?        Ss   10:15   0:00 php-fpm: master process (/opt/phpbrew/php/php-7.0.8/etc/php-fpm.conf)
www-data 13024  0.0  0.5 192208  5788 ?        S    10:15   0:00 php-fpm: pool www
www-data 13025  0.0  0.5 192208  5788 ?        S    10:15   0:00 php-fpm: pool www
root     13048  0.0  1.8 434764 18612 ?        Ss   10:18   0:00 php-fpm: master process (/etc/php5/fpm/php-fpm.conf)
www-data 13051  0.0  2.3 438040 24308 ?        S    10:18   0:00 php-fpm: pool www
www-data 13052  0.1  3.3 438712 34332 ?        S    10:18   0:00 php-fpm: pool www
```

**Comment by schlessera** [comment]
You should only need the `phpswap` command, without the additional `phpbrew` step.

However, the first time around, you might need to kill the php-fpm process by hand.

**Comment by ocean90** [comment]
@schlessera I did the `phpbrew` step only to check if it fixes the issue, it doesn't. php-fpm processes were killed multiple times. As you can see in the output of `ps aux | grep php` phpbrew is the only one which is running.

**Comment by schlessera** [comment]
@ocean90 I just tried to build a fresh system with this branch, which failed as well.

The issue on my build now was that php was listening on a different socket than nginx. A quick fix is to edit the `/opt/phpbrew/php/php-<version>/etc/php-fpm.d/www.conf` file, and make the `listen` clause listen to the old v5 socket:
```
listen = /var/run/php5-fpm.sock
```

As VVV has since switched to PHP7 as default version, I think that the `phpbrew/phpswap` integration needs to revised to match that new environment.

**Comment by mishtershmart** [comment]
I can't seem to get this working. @schlessera, I attempted to follow your suggestion but there was no php-fpm.d dir;

````
vagrant@vvv:/opt/phpbrew$ ls -lah /opt/phpbrew/php/php-5.6.25/etc/
total 128K
drwxrwxr-x 2 vagrant vagrant 4.0K Oct  2 23:09 .
drwxr-xr-x 9 vagrant vagrant 4.0K Oct  2 23:05 ..
-rw-rw-r-- 1 vagrant vagrant  23K Oct  3 00:52 php-fpm.conf
-rw-r--r-- 1 vagrant vagrant  23K Oct  2 23:05 php-fpm.conf.default
-rw-rw-r-- 1 vagrant vagrant  71K Oct  2 23:09 php.ini
````

I replaced the `listen` clause in the *php-fm.conf* file and when I run the `php -v` command, it reports the desired version;

````
vagrant@vvv:/opt/phpbrew$ php -v
PHP 5.6.25 (cli) (built: Oct  2 2016 23:01:26)
Copyright (c) 1997-2016 The PHP Group
Zend Engine v2.6.0, Copyright (c) 1998-2016 Zend Technologies
    with Xdebug v2.4.1, Copyright (c) 2002-2016, by Derick Rethans
vagrant@vvv:/opt/phpbrew$
````

However, when I dump `phpinfo()` on the site, the old PHP version is still running. Any thoughts on what I'm doing wrong?

**Comment by schlessera** [comment]
@mishtershmart So, you managed to get the version switching to work, but your web server was still using the original version?

Did you use the `-s` switch with `phpswap` ?

**Comment by mishtershmart** [comment]
@schlessera: I mustn't have been, as that seems to have done the trick. The problem that I'm now getting is a missing CURL extension.

**Comment by sudar** [comment]
Now that VVV uses PHP 7 by default, it would be really nice to get PHPBrew support built-in so that we can easily switch PHP version for old projects that are still struck with PHP 5.x in production.

I am going to merge latest `develop` branch into this branch locally and provision a new vagrant instance to test this branch (and report any issues). Meanwhile let me know if anything else can be done to get more love for this ticket :)

**Comment by tomjn** [comment]
Is there any benefit to `phpbrew` we don't already have besides giving PHP5.2 support? If so should this not be built as a utility? Or as a PR to the main core utility?

**Comment by tomjn** [comment]
I've discussed this with @LoreleiAurora and we agree we don't want to support unsupported versions of PHP. With that in mind we won't put barriers in the way

My recommendation is to create a new VVV utility named `phpbrew`, which provides `php55`/`php54` etc and all the extensions. Users could then include it in their vvv-custom.yml like this:

```YAML
utilities:
  core:
    - memcached-admin
    - opcache-status
    - phpmyadmin
    - webgrind
    - trusted-hosts
    - php56
  phpbrew:
    - php54
utility-sources:
  phpbrew: https://github.com/example/vvv-phpbrew.git
```

Where the `vvv-phpbrew` git repo contains folders such as `php54` and files such as `php54/provision.sh` that use PHPBrew to install PHP 5.4 etc.

Then to make use of php 5.4 in a site, you'd do it way the way PHP 5.6/7/7.1 etc work via nginx-upstream. [See here for how that works](https://github.com/Varying-Vagrant-Vagrants/vvv-utilities/blob/master/php71/provision.sh#L145)

Also see here for how php7.1 works in the main VVV utility. The only required file is `provision.sh`, and it knows to load it because `php71` is mentioned in `vvv-config.yml`, so it runs `php71/provision.sh`. The other files are there for the provisioner scripts use


If anybody does this, I'm happy to update the docs to reference it as an example, but this functionality doesn't belong in VVV core

Issue #788, opened 2015-11-20
Update NodeJS version to most recent LTS (4.2.x)


morganestes:
During a provision, one of the packages complains about having an outdated version of `node`. Normally that's not a major blocker, but since the regular `nodejs` package installs v.0.10.x and Node LTS stable is now at 4.2.2, I think moving to that  version is safe for general use. I'm open to other ideas, including installing v0.12.x, if there's a compelling reason not to go with 4.x.

Following the guidance at https://nodejs.org/en/download/package-manager/#debian-and-ubuntu-based-linux-distributions, I've tested swapping out the existing `nodejs` sources from Ubuntu to NodeSource and it seems to be working okay. Next up is updating the provisioning script to install it properly.

cfoellmann:
sounds good. Just send a PR so we can test it.

ntwb:
I've just come across this myself...

I'll take a look at the provisioning script to see what can be done to both provision and upgrade existing NodeJS installs next week sometime(hopefully)

FYI: The original NodeJS was switched to Chris Lea's Launchpad PPA in https://github.com/Varying-Vagrant-Vagrants/VVV/commit/803b7d8c9cf14cd794829d5ce860ec1b5577d425

morganestes:
I haven't had much luck getting it to work as part of the provisioning script for upgrade/change, but can get it to work for fresh installs. I don't think we should require folks to `destroy` just to swap sources, so I'm still looking for better solutions.

So far, I've tried two approaches: 1) dropping in the script directly from https://raw.githubusercontent.com/nodesource/distributions/master/deb/setup_4.x into `provision.sh` and letting it do the work, and 2) adding the PPA to the custom sources, adding the keys to `provision.sh`, and letting the normal provision process use the new deb.nodesource.com release.

ericandrewlewis:
How would a PHP (or another system package) upgrade be handled in VVV?

morganestes:
@ericandrewlewis I believe you can add it to the provision process using `provision-post.sh` to install your desired version, or through `apt-get` in the shell once it's provisioned. Either one would require Ubuntu to have an updated version in their repo, or a custom PPA set up to pull from.

From https://askubuntu.com/questions/109404/how-do-i-install-latest-php-in-supported-ubuntu-versions-like-5-4-x-in-ubuntu-1:
```
sudo add-apt-repository ppa:ondrej/php5-5.6
sudo apt-get update
sudo apt-get upgrade
[...]

PR #779, opened 2015-11-02
Switch to NodeSource for NodeJS packages

thebrandonallen:
Some of the WP node modules and their dependencies are beginning to complain about our outdated versions of NodeJS and npm. NodeSource seems to be the new standard for up-to-date NodeJS packages.

I've taken the conservative approach and chosen NodeJS v0.12.

Note, if someone runs `apt-get update` after a `vagrant provision` without first running `vagrant destroy`, apt-get will complain about not having the appropriate key for the nodesource packages. I'm not sure what the appropriate way to handle this. It may just be adding a something to the release notes saying to run `wget -qO- https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add -` if one experiences this issue.

PR #863, opened 2016-04-05
Add nvm

### LINKED ISSUE #562: Use nvm instead of nodejs
Out of the box, after a fresh provisioning, each time I need to use npm I have to add sudo before.

I finally removed nodejs package and installed nvm, it's like rvm for node (https://github.com/creationix/nvm)

Everything works without sudo now

Any reason not to switch in the provision script?

ntwb:
Initial support using nvm to allow alternate NodeJS versions to be installed, 
• https://github.com/creationix/nvm#manual-install

• By default no additional NodeJS versions are installed and the default installed NodeJS `system` version is used, currently this is `0.10.x`, #779 updates this to `0.12.x`.

• To install additional NodeJS versions use `nvm install 4`, or `nvm install 5` for the latest v4.x or v5.x branches respectively, 

• To list installed NodeJS versions use `nvm list`

• To switch to an installed NodeJS version use `nvm use 4`, or `nvm use 5` for the current installed v4.x or v5.x branches respectively

Fixes #562 Use nvm instead of NodeJS, see also:
• #779 Switch to NodeSource for NodeJS packages
• #788 Update NodeJS version to most recent LTS (4.2.x)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 * Add support for VMWare Workstation as a provider. See [656](https://github.com/Varying-Vagrant-Vagrants/VVV/pull/656).
 * Add MailCatcher to default provisioning. See [#632](https://github.com/Varying-Vagrant-Vagrants/VVV/pull/632).
 
-## Bug fixes and enhancements
+### Bug fixes and enhancements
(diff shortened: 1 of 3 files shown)


jeremyfelt:
We're close on merging this, but I want to gather some decisions in #562 first.

jeremyfelt:
Now's a good a time as any. :) Let's try it out. Thanks @ntwb!

Issue #800, opened 2015-12-16
Update the Virtualbox guest addition


Mte90:
http://kvz.io/blog/2013/01/16/vagrant-tip-keep-virtualbox-guest-additions-in-sync/

I've followed that guide to update the VB guest addition but is a simple plugin for vagrant that check on update on `vagrant up`.
Maybe is possible integrate in vvv?

fxbenard:
+1

cfoellmann:
Isn't this something we can only add to the readme as a recommendation?
The vagrant plugins are local and independent from Vvv or any other vagrant configuration

fxbenard:
@cfoellmann You're right, let's keep plugins apart from vvv. The recommendation will be a good move.

jeremyfelt:
Yeah, I think a README addition could work. I've never needed to update VB guest additions myself, but I have heard from others that they help.

Does this conflict at all with the updating of the ubuntu/trusty64 box through `vagrant box update`?

Mte90:
Not create conflict i am using that solution from months without problem.

16nsk:
I agree with @cfoellmann . I've been using that vagrant plugin for a few years and didn't have any problems.

It should be just a __recommendation__ in the VVV Readme.

meitar:
I was just hit by this issue today and found the recommended `vagrant-vbguest` plugin helpful, so I added it to VVV's recommended list of plugins in the README.

Also, sorry for the commit-spam, I had a brain fart and forgot to `git add` before I `--amend`ed.

jeremyfelt:
Documentation was added via #875.

PR #875, opened 2016-04-15
Add vagrant-vbguest plugins recommendation to the README, group optional plugin steps

meitar:
As per the discussion in issue #800, some people find it useful to use the `vagrant-vbguest` plugin for Vagrant to keep VirtualBox Guest Additions versions synchronized between the host and guest. Doing so can sometimes prevent subtle shared folder errors and doesn't seem to interfere with VVV's own operation.

This commit adds that plugin as a recommended installation step to the README. This commit also groups the three recommended plugins into a single logical step called "Optionally, install some convenient Vagrant plugins." Prior to this commit, each plugin was listed as its own step in the "first `vagrant up`" section of the README.


jeremyfelt:
Thanks for the PR, @meitar! I think there's an opportunity to continue improving these instructions and this is a good addition.

PR #877, opened 2016-04-17
fix some ShellCheck code linter warnings

keegoid:
- make code more standard and readable
- insert space around code inside $(( ... ))
- define variables within functions as local

Shell is special, even variables declared inside a function and inside
another block, appearing to have limited scope, will still be available
after the function call. Define as local to avoid potential issues.

### DIFF
--- a/provision/provision.sh
+++ b/provision/provision.sh
@@ -163,25 +163,32 @@ not_installed() {
   fi
 }
 
+print_pkg_info() {
+  local pkg="$1"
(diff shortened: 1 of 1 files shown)


jeremyfelt:
Thanks @keegoid, this looks great!

I've been meaning to look into ShellCheck some more. It may be nice to setup Travis to run a lint against VVV for every PR to help with testing and merging.

PR #881, opened 2016-04-21
Update base box for VMWare

khromov:
This PR changes the base box for VMWare Fusion / Workstation. 

The [old box](https://vagrantcloud.com/netsensia/boxes/ubuntu-trusty64) has not been updated for 2 years and has some very old packages. The [new box](https://atlas.hashicorp.com/cbednarski/boxes/ubuntu-1404) was updated more recently (currently with 14.04.3 about 9 months ago).  

I also found the old box had some problems with HGFS (shared folder) initiation that would cause it to hang on provision intermittently (Related issue: https://github.com/mitchellh/vagrant/issues/6775 )

The VVV provisioning was tested and successful on the new box.


jeremyfelt:
Thanks for the PR @khromov! I'm more than happy to merge this in, though I haven't used the VMWare configuration at all myself. I'd like to get a +1 if possible from others using VMWare.

My only concern is that a box change is probably considered breaking in that someone would need to back-up all of their work before provisioning.

ntwb:
This PR needs rebasing

jeremyfelt:
Ok, it looks like the next version of VVV is going to be best with a full destroy and provision, so it would be the right time to update the box if necessary.

Both the current box and the new box have about 10,000 downloads through https://atlas.hashicorp.com/boxes/. There are a few others that have hundreds of thousands of downloads. Is it possible that another box is the right answer for the long term? https://atlas.hashicorp.com/geerlingguy/boxes/ubuntu1404 looks like it may have some promise.

I'm not opposed to using the suggested box, but I want to make sure we're getting this right for the next couple years.

Also - we may need a new PR that just replaces the box name. I can cherry-pick if needed, but a clean PR would be helpful.

ntwb:
> Ok, it looks like the next version of VVV is going to be best with a full destroy and provision, so it would be the right time to update the box if necessary.

Sounds good to me 👍 

What about https://atlas.hashicorp.com/geerlingguy/boxes/ubuntu1604 ?

jeremyfelt:
I'm going to close this PR and defer to #1074 as the new issue for tracking the state of this. I think we need a new PR that uses the geerlingguy box and is based against a more recent version of the `develop` branch.

PR #896, opened 2016-05-05
Docker Support

ericmann:
This change adds Docker support to VVV and builds in containers for PHP 5.3 through 7.0, pre-configured for WordPress sites (thanks to the amazing work of [10up](http://10up.com)). The containers are each configured to run as distinct PHP back-ends, each listening on a distinct port:
- 5.3 on port 9053
- 5.4 on port 9054
- 5.5 on port 9055
- 5.6 on port 9056
- 7.0 on port 9070

Each is then configured as an available upstream in Nginx (alongside the stock `php` upstream provided during provisioning as well). By default, sites will use the regular php process, but developers can:
1. Define a custom back-end in their config by adding `set $default_backend php54;` (for PHP 5.4) to the site's `server` block.
2. Dynamically select their PHP backend by specifying an environment cookie during the request (the README is updated with bookmarklets that do this for you). The cookie is read by Nginx and will override the default backend specified in the configuration file.

**Note:** There is no PHP 5.2 container as that version pre-dates FPM availability, but the containers provided give broad support for proactive compatibility testing during development.

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,6 +180,29 @@ A bunch of stuff!
 1. [NodeJs](https://nodejs.org/)
 1. [grunt-cli](https://github.com/gruntjs/grunt-cli)
 1. [Mailcatcher](http://mailcatcher.me/)
+1. [Docker](https://www.docker.com/)
+
+### Dockerized PHP
+
+VVV ships with PHP built-in by default and, with no changes from you, will use that stock version of PHP just fine. But in case you need to test your code with _other_ versions of PHP, we've also added Docker-wrapped installations of all the recent major versions as well:
+* 5.3
+* 5.4
+* 5.5
+* 5.6
+* 7.0 - this is installed normally as well
+
+If you want your site to use a specific PHP backend (read: version), then add `set $default_backend phpXX;` to the `server` block of that site's Nginx config file. The XX in this string stands for the version number you want to use (i.e. 5.3 would be `php53`, 5.5 would be `php55`, 7.0 would be `php70`). Your site will now use _that_ version of PHP by default!
+
+#### Cookie Control
+
+We've also set up a tricky Nginx cookie map so you can swap the back-end powering a site dynamically. Click and drag the following bookmarklet links to your browser's toolbar, and you can click them to dynamically change which version of PHP will deliver the site:
+
+* <a href="javascript:(function() {document.cookie='env='+''+';path=/;';})()">PHP</a> - Resets you back to stock PHP
+* <a href="javascript:(function() {document.cookie='env='+'53'+';path=/;';})()">PHP53</a>
+* <a href="javascript:(function() {document.cookie='env='+'54'+';path=/;';})()">PHP54</a>
+* <a href="javascript:(function() {document.cookie='env='+'55'+';path=/;';})()">PHP55</a>
+* <a href="javascript:(function() {document.cookie='env='+'56'+';path=/;';})()">PHP56</a>
+* <a href="javascript:(function() {document.cookie='env='+'70'+';path=/;';})()">PHP70</a>
 
 ### Need Help?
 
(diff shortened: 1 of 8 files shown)


lkwdwrd:
To be very clear about what this adds - you get the ability to run different sites on different PHP versions _side by side_ with each other. And further more with the cookie mapping you can test multiple versions of PHP on the same site with the click of a bookmarklet.

Also worth noting, this requires the 64-bit OS. Switching to the 32-bit OS (precise32, trusty32) wont work. I know some end users have done this to fix some stability issues with their machines.

jeremyfelt:
This is pretty freakin cool, thanks @ericmann. 👏 

Everything installed well during provisioning and I was able to flip back and forth by changing the backend in the config for `http://vvv.dev/phpinfo/` without much effort. It looks like it would be even easier to do the cookie switching for the WP sites. The changeset itself is very straight forward, so it's pretty easy to follow. 💯 

A few quick thoughts/questions/first impressions:

* How does remote debugging play into all of this? Is the xdebug configuration shared or is that up to the individual PHP docker images?
* How does this work with command line tools such as `phpunit` and `wp`? It would be pretty fantastic to run a suite of tests against multiple versions.
* It appears that PHP's configuration for each of these is deployed inside the Docker container. Does VVV have any control over that configuration or is it shipped as part of the upstream package?
* We have quite a bit of progress on PHPBrew in tickets such as #854. It would be interesting to compare some pros/cons of each approach.
* My one immediate hesitation is that the docker images for each version of PHP seem to always be running in the background, even when not in use. Maybe a command similar to PHPBrew could make sense where it makes a single image available at a time as **the** backend.

Would like to hear others thoughts on this as well.

> There is no PHP 5.2 container as that version pre-dates FPM availability

I believe it's possible with a `php-fpm` patch, but seems uuuuuughhhly and unfriendly.

ericmann:
> How does remote debugging play into all of this? Is the xdebug configuration shared or is that up to the individual PHP docker images?

I haven't yet played with remote debugging across the Docker instances yet, but the xdebug configuration would need to be per-container. Each container is completely isolated, so this would enable you to have xdebug on in one while off in the others, but again, not something I've set up just yet.

> How does this work with command line tools such as phpunit and wp? It would be pretty fantastic to run a suite of tests against multiple versions.

The commands change a tiny bit due to the nature of using a container. On the one hand, you _could_ `docker exec` into a new container in Bash, then `cd` into the project directory and run commands directly. On the other hand, I've had some success with aliasing `wp` and `phpunit` on a per-project basis to run an `exec` command directly. I'll look up what I did to make that happen, but it _does_ enable running test suites against multiple PHP configurations.

> It appears that PHP's configuration for each of these is deployed inside the Docker container. Does VVV have any control over that configuration or is it shipped as part of the upstream package?

Right now, it's using just the default configuration, but we _could_ map a VVV-controlled config into the containers to give external control. Do you have a specific config file (that would work with all of the versions) you'd like mapped in?

> We have quite a bit of progress on PHPBrew in tickets such as #854. It would be interesting to compare some pros/cons of each approach.

Biggest con to PHPBrew is that you won't be running that in production - but in quite a few cases, you _would_ be running a Docker container to wrap PHP. If we're seeking at all to retain dev/production parity, that's a big concern. A big pro to Docker is that you can run _multiple_ containers at the same time for different projects. If you've got 5 client projects, each targeting a different PHP version in production, swapping PHPBrew multiple times when switching focus is a hurdle to productivity - Docker would let you set up each to use the right version of PHP and just go.

> My one immediate hesitation is that the docker images for each version of PHP seem to always be running in the background, even when not in use. Maybe a command similar to PHPBrew could make sense where it makes a single image available at a time as the backend.

The containers aren't that heavy when they're idling ... what's your concern here?

> >There is no PHP 5.2 container as that version pre-dates FPM availability

> I believe it's possible with a php-fpm patch, but seems uuuuuughhhly and unfriendly.

Oh, there are workarounds, but all of them are ugly, difficult to maintain, and held together with naught but duct tape and prayers.

jeremyfelt:
Thanks for the effort on creating and documenting this PR, @ericmann. I appreciate the possibilities here and the work!

I'm going to close this in favor of providing PHPBrew as the default approach via VVV-Utilities. In the newer provisioning configuration slated for 2.0.0, it will be possible to disable this and replace it with some custom, like Docker support.

PR #898, opened 2016-05-09
Add php7.0-soap

GaryJones:
The SOAP module may be common enough to be part of the default install?


jeremyfelt:
I run the other way when I hit a SOAP project, but it's nice to have when you need it. 👍 

Thanks, @GaryJones!

PR #909, opened 2016-05-24
Enable query cache by default

jdevalk:
Makes it about a gazillion times faster.


rene-hermenau:
:+1: Very useful

jeremyfelt:
Let's give it a shot. :) Thanks @jdevalk!

PR #925, opened 2016-06-14
query_cache_* entries should be under [mysqld]

weskoop:
query_cache_* entries should be under [mysqld] not under [mysql]. the current way is throwing errors on provision


jeremyfelt:
Oops! Thanks @weskoop!

PR #926, opened 2016-06-15
[PR] Tracking new contributors for README

jeremyfelt:


jeremyfelt:
Went through the commit log through October 2015 and I don't think we're missing anyone.

Issue #876, opened 2016-04-16
Update from MySQL 5.5 to MariaDB 10.0


ottok:
If I'd write a pull request that updates MySQL 5.5 to MariaDB 10.0 (from official Ubuntu repositories), would the community here accept it?

Motivation:
- Many are upgrading from Oracle MySQL to truly open source MariaDB by original creators of MySQL. For example Fedora/RedHat and SUSE ship by default only MariaDB since a few years back. MariaDB has been available in Ubuntu since 14.04 and in Debian since version 8. Debian is planning to have MariaDB as default in next release, Ubuntu might follow.
- Projects like [XAMPP](https://www.apachefriends.org/) already has MariaDB as default.
- Automattic, the company behind WordPress, is a sponsor of the [MariaDB Foundation](http://mariadb.org) and much of the new WordPress infrastructure builds on MariaDB
- Techincal benefits listed at https://seravo.fi/2015/10-reasons-to-migrate-to-mariadb-if-still-using-mysql

Why version 10.0 and not 10.1?
- Official Ubuntu repositories have MariaDB 10.0, so for maintenance in long term it is best to stick to it and upgrade when Ubuntu/Debian also upgrade.

I want to have a thumbs up before I invest a lot of time in developing and testing.

ottok:
Any comments from VVV developers? Will you merge the PR if I implement this?

johnjamesjacoby:
+1 for MariaDB

moorscode:
I would love to have MySQL updated from 5.5 to anything at this point.
Though it seems to be harder to change to MariaDB if we first go to MySQL 5.7 in #930 so this would have to be a consideration when looking at that PR.

It seems that going for MySQL 5.7 would be the easiest way to pick, but going for MariaDB the most future-compatible one. Though I wouldn't mind to hear some counter arguments for switching to MariaDB, just to get an idea of what they would be.

I can imagine that a lot of sites are still running on MySQL and might get updated to 5.6 or 5.7 and with the internal choices growing apart this would not fair well on having a reproduction of the production server as development environment. Though this is just speculation and some logic thrown around :)

jeremyfelt:
I've created https://github.com/Varying-Vagrant-Vagrants/VVV/issues/876 to discuss this further.

moorscode:
@jeremyfelt you mean https://github.com/Varying-Vagrant-Vagrants/VVV/issues/949

lukecav:
[...]

PR #930, opened 2016-06-27
Update MySQL to 5.7

### LINKED ISSUE #928: Upgrade MySQL to 5.6
One of the minimum requirements for WordPress is MySQL 5.6.

Ubuntu 14.04 still has the default version at 5.5 but 5.6 is available in the default repositories.
I changed provision/provision.sh line 59 from mysql-server to mysql-server-5.6 then did a vagrant provision to get it updated.
So far everything is running well without any problems.

LoreleiAurora:
Fixes #928

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,7 +180,7 @@ A bunch of stuff!
 1. [WordPress Trunk](https://core.svn.wordpress.org/trunk/)
 1. [WP-CLI](http://wp-cli.org/) (master branch)
 1. [nginx](http://nginx.org/) ([mainline](http://nginx.com/blog/nginx-1-6-1-7-released/) version)
-1. [mysql](https://www.mysql.com/) 5.5.x
+1. [mysql](https://www.mysql.com/) 5.7.x
 1. [php-fpm](http://php-fpm.org/) 7.0.x
 1. [memcached](http://memcached.org/)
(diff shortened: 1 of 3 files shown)


johnjamesjacoby:
See also #876 to consider using MariaDB instead.

moorscode:
When trying these changes in a test vagrant I get the following output:

```
==> default: GPG error: http://ppa.launchpad.net trusty InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 4F4EA0AAE5267A6C
```

Which breaks the installation and leaves the Vagrant without MySQL package being installed (among other packages).

jeremyfelt:
I've created https://github.com/Varying-Vagrant-Vagrants/VVV/issues/876 to discuss this further.

moorscode:
#949 instead of 876.

ottok:
Please avoid using PPAs. Stick to the packages in official Ubuntu repositories for best quality, security are reliability.

RobertoSnap:
We have formerly used [PHP 7 Rasmus vagrant](https://github.com/rlerdorf/php7dev) for our local development enviroments. Decided to switch to VVV. Had to use this pull so we could import our databases with utf8mb4_unicode_ci collation.

PR #937, opened 2016-07-13
Add Elasticsearch 2.3.4

igmoweb:
- New Elasticsearch setup in provision
- Start Elasticsearch servic on startup
- Added Elasticsearch default config file
- Added Elasticsearch forwarded ports

Elasticsearch can be really useful to develop large sites with complex queries, so why don't add it to VVV? There's actually some plugins for WordPress.

Make a provision and then:

From guest VM: curl 127.0.0.1:9200
From host: curl or open your browser on http://192.168.50.4:9200/

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -148,6 +148,7 @@ Vagrant.configure("2") do |config|
   # Please see VVV and Vagrant documentation for additional details.
   #
   # config.vm.network "forwarded_port", guest: 80, host: 8080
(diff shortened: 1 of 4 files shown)


jeremyfelt:
Thanks for the PR @igmoweb! I don't want to add much additional weight to the current configuration. However, I would like to start encouraging more custom/official extensions of the default provisioner. I'm working on a proposal at the moment that may help clarify how we can go about this. For now, I think the best place for an ES configuration is in `provision-custom.sh`.

PR #967, opened 2016-08-20
support provisioning scenarios without wget

diddledan:
Specifically this allows to test the network with curl as a
fallback from wget if it's unavailable on the installed system.

Curl is more likely to be installed in a very minimal system,
such as those created for Docker.

This is required for my Docker-supporting PR #968 .


LoreleiAurora:
This would require the function to be copied into every provisioner in the new provisioning system, why not just add wget to the list of packages to be installed?

jeremyfelt:
Thanks for the PR, @diddledan. I hate to disappoint, but I've closed #968. Providing `curl` and `wget` doesn't seem necessary now as all of the other provisioners should support `wget`. Thank you for your effort!

PR #968, opened 2016-08-20
Docker support

diddledan:
Requires my previous PR #967.

Support Docker and Native Docker on all platforms (Windows, macOS, and Linux).
Update readme with details of how to prevent vagrant from creating a new VM on Windows and macOS when you have native-docker installed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,47 @@ FROM phusion/baseimage
 
 RUN rm -f /etc/service/sshd/down
 
-RUN apt-get update && apt-get install -y sudo
+RUN echo "mysql-server mysql-server/root_password password root"       | debconf-set-selections && \
+    echo "mysql-server mysql-server/root_password_again password root" | debconf-set-selections && \
+    echo "postfix postfix/main_mailer_type select Internet Site"       | debconf-set-selections && \
+    echo "postfix postfix/mailname string vvv"                         | debconf-set-selections && \
+    apt-get update && apt-get install -y \
+        colordiff \
+        dos2unix \
+        gettext \
+        git \
+        graphviz \
+        imagemagick \
+        memcached \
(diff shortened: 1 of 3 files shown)


d3v1an7:
Hey @diddledan, thanks for you work on this.

I've just tried this locally, in conjunction with #967 and I had some issues.
Namely, both `wget` is used quite a bit in provision.sh (more than just the connection test) and `rysnc` is also missing from phusion/baseimage.
I was also getting a bunch of other errors during provision (missing/invalid packages etc).

I rolled provision.sh back to HEAD, then make the following changes to your Dockerfile:
```
# baseimage by default is using 
FROM phusion/baseimage:0.9.18
```
and:
```
RUN apt-get update && apt-get install -y sudo wget rsync
```

Also, for README.md, I believe `CustomFile` should be `Customfile`?

With those changes, I was able to successfully run `vagrant up`, but there must have been a few other 'silent' fails during the provision, as the default URLs are all inaccessible (resolves to correct IP, but returns no data).

Did you have any other changes to get this up and running on your end?

diddledan:
I've tweaked the Dockerfile in 897c116.
Also fixed the README.md (correcting the typo of Customfile) in 23a8905.

I can successfully provision under Linux-native Docker now, but there seem to be problems with macOS+boot2docker support (the older method). I still need to verify Windows-native and Windows+boot2docker in addition to macOS-native. (I'm running macOS Sierra developer beta right now which appears to have broken macOS-native Docker for now, which is why I haven't been able to test it properly there)

jeremyfelt:
I've gone back and forth on this a bit, but the maintenance cost of having Docker as another provider supported by VVV seems like it will be too high. Sorry to keep you hanging on this pull request for so long, and thanks for the effort you've put into it.

PR #976, opened 2016-08-21
Add SSH2 PHP Extension

### LINKED ISSUE #923: Add libssh2-1-dev and libssh2-php modules
This is a continuation of #836, which was created before VVV used PHP 7. Now that we do use PHP 7, the libssh2-php module is not directly available and must be built from source.

I'd rather not build anything from source during provisioning, but we should have an issue open to track inclusion at some point.

I'd also like to gather some use cases for these modules so that we know it makes sense to add them in default provisioning and not as a custom configuration.

LoreleiAurora:
Fixes #923

### DIFF
--- a/provision/provision.sh
+++ b/provision/provision.sh
@@ -49,6 +49,7 @@ apt_package_check_list=(
   php7.0-imap
   php7.0-json
   php7.0-soap
+  php7.0-ssh2
   php7.0-xml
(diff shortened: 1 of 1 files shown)


jeremyfelt:
Excellent, I'm not sure why I wasn't able to find the right name for it the other day. Thanks!

PR #977, opened 2016-08-22
Add BCMath PHP Extension

### LINKED ISSUE #944: Box doesn't include PHP's BC Math Module?
I wasn't able to use any of [PHP's BC Math functions](http://php.net/manual/en/ref.bc.php) when running the latest version of VVV (it just throws PHP errors). Is this module not included? If not, why?

LoreleiAurora:
Fixes #944

### DIFF
--- a/provision/provision.sh
+++ b/provision/provision.sh
@@ -41,6 +41,7 @@ apt_package_check_list=(
   php-imagick
   php-memcache
   php-pear
+  php7.0-bcmath
   php7.0-curl
   php7.0-gd
   php7.0-mbstring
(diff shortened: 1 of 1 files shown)


LoreleiAurora:
@khromov It appears that way but the ssh2 module is already in master, in that exact position. 😕

khromov:
I see. There's something really funky going on in this repo. One of my pull requests suddenly got unrelated commits added to it. :weary:

jeremyfelt:
Thanks @LoreleiAurora! And thanks @cfoellmann for testing! :)

Issue #482, opened 2014-10-29
Move to Git?


aaroncampbell:
I don't want to start a holy war, but after having moved most of my workflow to git, when I try to do core dev on VVV it's painful to be stuck in SVN. What is the consensus (or lack thereof) for moving to using git for wordpress-develop (and maybe wordpress-trunk)?

aaroncampbell:
> If the goal of this is simply to make diffing and patching easier on folks unable or unwilling to learn svn, then just say that. 
"The goal of this is simply to make diffing and patching easier on folks." At least in my opinion that's the case. I think that VVV is supposed to make developing with and on WordPress as easy as possible, and I think git helps do that by letting people use modern tools they are familiar with and like.

ericandrewlewis:
Drew said:
> My feeling is that as long as core continues to be versioned with svn, VVV -- considered to be the officially recommended vagrant package for core development -- should follow that lead. 

Whether or not there's any official recommendation by the WordPress core team, the VVV community should define and serve its userbase independently. I think the 80/20 rule should apply here. 

The fact that WordPress core primarily uses svn doesn't apply to this discussion. The project's reasons for staying on svn, as I understand, are a matter of priority and backwards compatibility. That's a separate issue from the goal of VVV: providing the greater contributor-base with a VM to use for WordPress core development (among other things).

I think we can all agree svn is the past and git is the future. I'm personally unwilling to learn any more about svn than I already do, unless out of necessity, because it is an archaic technology. We should push contributors to use modern, familiar, and better tools.

jeremyfelt:
I totally agree that git would be the more pleasant default. Speed continues to be my primary concern with switching before having a configurable VVV. There's just no easy way for a user to decide which sites should be created in advance. Ideally, I could tell provisioning to skip everything except for the develop repo, and then check it out with git rather than svn.

With my unshaped 34Mbps down connection at home:
* 34 seconds for `git clone git://develop.git.wordpress.org/`
* 14 seconds for `svn checkout https://develop.svn.wordpress.org/trunk`

If I limit that connection to 5Mbps:
* 5 minutes, 23 seconds for `git clone git://develop.git.wordpress.org/`
* 49 seconds for `svn checkout https://develop.svn.wordpress.org/trunk`

[...]

PR #978, opened 2016-08-23
Move to using GIT as the default for the WordPress Develop repository…

aaroncampbell:
…. See #482


LoreleiAurora:
As the provisioning scripts are now in separate repositories this issue should be against [Varying-Vagrant-Vagrants/vvv-wordpress-develop](https://github.com/Varying-Vagrant-Vagrants/vvv-wordpress-develop).

Issue #210, opened 2013-11-24
Auto-site setup all the things


simonwheatley:
![43272801](https://f.cloud.github.com/assets/233434/1609507/75a3bc2a-5548-11e3-867f-1fb68274dc93.jpg)

Should we move all the sites in the `www` directory to use the auto-site setup method? Dog fooding, and all that. Plus it'd give handy built in examples for people.

Should we even consider moving some of those sites out into their own repos, allowing VVV to lose a little bloat (so we don't have `wordpress-default` and `wordpress-develop` and `wordpress-trunk`)?

jeremyfelt:
The dream has been realized in #980! :taco:

PR #980, opened 2016-08-23
Split provisioners into external git repos

### LINKED ISSUE #313: Ability to disable installation of default WP sites

### LINKED ISSUE #474: Allow selecting which sites to make active

### LINKED ISSUE #753: Add site without re-provisioning existing site
Not sure if this is a support or feature request, but I want to be able to provision new sites without running the complete provisioning process on existing sites? Mainly I'm trying to speed up the process and avoid SVN merge errors. Is this possible?

### LINKED ISSUE #771: Folder structure for auto site setups is inflexible
In the Meta Environment, we've run into [an issue](https://github.com/WordPress/meta-environment/issues/9) where users have to clone the repo to a specific folder name in order for provisioning to work.

We have a folder structure like:

`{vvv_root}/wordpress-meta-environment/wordcamp.dev/logs`
`{vvv_root}/wordpress-meta-environment/wordcamp.dev/provision`
`{vvv_root}/wordpress-meta-environment/wordcamp.dev/public_html`

That keeps things clean and organized, but it prevents us from using `{vvv_path_to_folder}` in our Nginx conf files, because the resulting path would be 

`/srv/www/wordpress-meta-environment/wordcamp.dev/provision/`
 instead of 
`/srv/www/wordpress-meta-environment/wordcamp.dev/`

So we've just hardcoded `/srv/www/wordpress-meta-environment` for now. That works if the user clones the directory with that name, but it breaks if they choose a custom folder name.

There are two approaches I can think of to improve the situation for all auto-site-setups:

1) Provide a way for auto-site-setups to customize the replacement value of `vvv_path_to_folder`.

2) Provide a way for auto-site-setups to hook into `provision-post.sh`.

I'm thinking that there'd be a default `/provision/provision-post.sh` file that would `find /srv/www -maxdepth 5 -name 'provision-post.sh'`, and `source` the results.

This seems like it might be the better approach, because it would allow for a lot of other customization  as well, beyond just this situation.

With something like this in place, we could just setup out own `{wme_path_to_folder}` placeholders and do our own `sed` replace in our child `provision-post.sh`.

There'd be some back-compat issues, though, since some users will have already created their own provision-post.sh files (#48).

Any thoughts on the above approaches, or better ideas?

### LINKED ISSUE #851: `Customfile` from any folder, not just Vagrantfile path (or alternative)
Scenario:
We would like to be able to execute certain actions during a `vagrant up`. Different project may have their own set of tasks to perform. Ideally, these tasks would be defined in the project's git repository.

At the moment this can be accomplished by making use of the `Customfile` and adding tasks in here. However, this is not part of the project and when different projects require tasks it needs to be added to `Customfile` outside of the project repo.

Suggestion would be to allow multiple `Customfile`s to be created in the Vagrantfile's directory tree. 

Example:

```
Dir.glob('**/Customfile').each do |f|
    eval(IO.read(File.join(vagrant_dir,f)), binding)
end  
```

vs

```
if File.exists?(File.join(vagrant_dir,'Customfile')) then
    eval(IO.read(File.join(vagrant_dir,'Customfile')), binding)
end
```

This can be unstable and potentially a risk of allowing arbitrary `Customfile`s to be defined in projects.  

Open to suggestions.

LoreleiAurora:
# Tasks:
- [x] YAML config file to control VVV (Fixes #313)
- [x] Allow selecting which sites to make active (Fixes #474)
- [x] Ability to provision sites individually (Fixes #753)
- [x] Split provisioning of WP sites (Fixes #611 #210)
---
- [x] Ability to include `Customfile` from any folder (Fixes #851)
- [x] Pass site name & root dir to vvv-init scripts
- [x] Pass site name & root dir to nginx config (Fixes #771)
- [x] Allow a site to be stored outside the VVV dir on the host machine
# YAML config file to control VVV

The YAML config file will allow inclusion of external provisioning scripts from git repos and disabling of provisioning of specific sites (at a later date it can be used to configure other aspects of VVV).

To use a default option for any site parameter simply miss it out of the site block.

``` YAML
sites:
  # All options defaults that can be used in a site config
  wordpress-default:
    # The git url for this site
    # e.g. https://github.com/Varying-Vagrant-Vagrants/vvv-wordpress-default.git
    repo: ''
    # The directory on the vm where the site lives
    # e.g.  '/srv/www/wordpress-default'
    vm_dir: /srv/www/{SITE NAME}
    # The directory on the host where the site lives
    # e.g.  '/PATH TO VVV/www/wordpress-default'
    local_dir: vagrant_dir/www/{SITE NAME}
    # Weather to use the sites Customfile if one is found
    allow_customfile: false 
    # Weather to run this sites provisioner
    skip_provisioning: false

    # If you are only providing a repo url to your site config 
    multisite: https://github.com/Varying-Vagrant-Vagrants/vvv-multisite.git

    # Site not stored in git accepting the default options
    site_name:
```
# Ability to provision sites individually

Sites can be provisioned individually using the following vagrant command
`vagrant provision --provision-with site-{sitename}`
e.g. `vagrant provision --provision-with site-wordpress-default`
# New Variables

| Bash Variable | Nginx Variable | Example |
| --- | --- | --- |
| ${VVV_PATH_TO_SITE} | {vvv_path_to_site} | /srv/www/wordpress-default |
| ${VVV_SITE_NAME} | {vvv_site_name} | wordpress-default |

Path to site uses the vm_dir as defined in the config as apposed to the path to the dir in which the nginx config or shell script was found. This resolves the issue listed in #771.


LoreleiAurora:
@jeremyfelt needs:feeback, needs:testing

mbijon:
This is excellent for granularity, resources, and general control-freak moments. All these changes get a +1 from me (I haven't tested yet, but several of these things are great ideas).

It's a huge batch of changes though. So I'll say something I always say when I like something and want to see it in prod... if there's any resistance please break it up and include whatever possible. Let's only delay pieces that are disputed or buggy, not everything g.

jeremyfelt:
I haven't had time to dig in yet, but I want to leave a note of general support for this. I'll take a closer look over the next few days. Thank you @LoreleiAurora!

cfoellmann:
how about changing the dashboard like so?
```yaml
---
default:
  dashboard:
    repo: https://github.com/topdown/VVV-Dashboard.git
---
sites:
  wordpress-default:
    repo: https://github.com/Varying-Vagrant-Vagrants/vvv-wordpress-default.git
    vm_dir: '/srv/www/wordpress-default-provisioner'
[...]
```

cfoellmann:
@LoreleiAurora the timer for the provisioning is shown before the start of the provisioning of the sites. Not at the very end of the process

LoreleiAurora:
@cfoellmann i forgot the timer existed, as each site is called as a separate vagrant provisioner i'm not really sure how we could keep it. Unless anyone else has any ideas?

cfoellmann:
maybe add the counter to all provisioners? I can add them together. 😸

LoreleiAurora:
@cfoellmann if you want to try that you can.
If you have you have push on this repo you also have push on LoreleiAurora:split-provisioners

cfoellmann:
@jeremyfelt can you add some repos for multisite?

- vvv-wpmu-subdirectory

- vvv-wpmu-subdirectory-trunk

- vvv-wpmu-subdomain

- vvv-wpmu-subdomain-trunk

cfoellmann:
@LoreleiAurora is it right that if I copy my "sites" (like one folder from vvv-multisite) manually to www it won't provision it with this version?

LoreleiAurora:
@cfoellmann you need to add `wpmu-subdirectory-trunk.dev:` to the sites list in vvv-config.yml

jeremyfelt:
Ok! Now that 1.4.0 is released, let's have some fun. :smile: 

I tested this out and things are absolutely wonderful. I'm going to merge now and then we can iron out a couple things once it's in the develop branch. I'll create issues for each of these:

1. We've lost temporary access to `vvv.dev` as it was part of the default `vvv-hosts` file. We'll probably need to restore `vvv-hosts` just for that.
2. After seeing how well this is working for the individual WP repos, I think it makes sense to break out phpmyadmin, memcached-admin, opcache-status, and webgrind into their individual VVV repos so that they can be selectively disabled.
3. I like all of the config names in the YAML file. It may be cool to one day support zip/tar packages as well, but I'm not sure how often that would be used.
4. An option to provide a custom YAML file. If say `vvv-custom.yml` is provided, then we defer to that instead of the version controlled copy. This will make it easier to customize while staying updated with the upstream project.
5. @cfoellmann let's start a new issue to track the inclusion of multisite configs by default. I'm a bit wary of going too crazy with the repos, but it definitely will be useful to determine what's needed.

Thanks for all the work on this @LoreleiAurora! Such a great move forward for VVV. :100:

Mte90:
There is a docs about that file? like the name of the file? what is the path? Also this works with old installation of vvv?
sorry for the many questions XD

LoreleiAurora:
@Mte90 The file is `vvv-config.yml` in the root dir.
For this to work with your existing installation your config would look something like this:
```yaml
sites:
  wordpress-default:
    repo: https://github.com/Varying-Vagrant-Vagrants/vvv-wordpress-default.git
    vm_dir: '/srv/www/wordpress-default-provisioner'

  wordpress-develop:
    repo: https://github.com/Varying-Vagrant-Vagrants/vvv-wordpress-develop.git
    vm_dir: '/srv/www/wordpress-develop-provisioner'

  // The below lines are the directory names of your site configs followed by a colon.
  preexisting-site-1: 
  preexisting-site-2:
  preexisting-site-3: 
```

Mte90:
So in the case that I have many installed with vv this is compatible?
In case it's better that vv get the support for this before to use in my case

LoreleiAurora:
I have not tested compatibility with vv, but if its not i would love to help make it compatible. 
Ping bradp

LoreleiAurora:
@jeremyfelt Rather than having a repo for each of utilities what do you think about adding another key to the yaml file and storing them in one repo. This would give a config like this:
```yaml
sites: 
  <Redacted for simplicity>

utilities:
  phpmyadmin: true
  memcached-admin: true
  opcache-status: true
  webgrind: true

Then any other utils (including third party ones) can be added to this repo and work with just one line. Kind of like a VVV plugin directory.

For example i have mongodb & [adminmongo](http://adminmongo.com/) installed in vvv, this config could live as part of the utils repo.

cfoellmann:
Shouldn't we look into provisioning the utilities via composer? I wanted to do it in the past but time...

LoreleiAurora:
@cfoellmann we can do that in the scripts, but we keep the scripts that trigger the installs in that repo allowing non composer tools to be added with 1 uniform method of configuration.

topdown:
What about adding a Yaml linter to validate these YAML files?
https://github.com/adrienverge/yamllint

Would hopefully lower support issues/questions when this PR is rolled out.
Just a thought.

LoreleiAurora:
@topdown can you please open that as a new issue as this is already in develop.

PR #982, opened 2016-08-30
Adds plugins directory sync and symlinks and auto activation

roelveldhuizen:
I added a plugins directory where each directory is symlinked in the WordPress plugins folders and all plugins are automatically activated.


jeremyfelt:
Hi @roelveldhuizen, thanks for taking the time to submit a PR. This seems like something that would be great as part of a custom installation, but I don't see including this as part of the core configuration.

PR #990, opened 2016-09-18
WP-CLI: Use composer install instead of update

khromov:
I noticed that WP-CLI wasn't updating successfully on `vagrant provision` and noticed the error `Cannot pull with rebase: You have unstaged changes.` so I dug around in the provision file and found that after pulling the latest WP-CLI version `composer update` runs.

This command does install dependencies but also updates them to the latest version, writing `composer.lock` in the process and makes it impossible to pull a new version from GitHub. 

This is easily solved by running `composer install` instead which is the proper way to install dependencies.

Ping @jeremyfelt

### DIFF
--- a/provision/provision.sh
+++ b/provision/provision.sh
@@ -582,7 +582,7 @@ wp_cli() {
     echo -e "\nUpdating wp-cli..."
     cd /srv/www/wp-cli
     git pull --rebase origin master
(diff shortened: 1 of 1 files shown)


grappler:
Please can you merge this. I can confirm it fixes the problem.

PR #995, opened 2016-09-29
Script to facilitate changes to xdebug.ini and php-custom.ini

btsimonh:
These three commits do the following:
One removes max_execution_time from php-custom.ini, and puts it in xdebug.ini.  This means that max_execution_time changes when xdebug is enabled automatically.  A value in php-custom.ini would override it.
the next one adds a script to homebin called 'copyxdebug'.  This script replicates what provision.sh does for the xdebug.ini and php-custom.ini - copies them to the correct location (and adds a line to the top of xdebug.ini indicating the location of the xdebug library.
The final one sets xdebug.remote_autostart = 0
It also moves this to the top of xdebug.ini for easy access.  When set to the VVV default '1', if xdebug is turned on, then you can't view the website unless you have a debugger at the target debug address running?  not sure of the impact of this on other using different environments.

### DIFF
--- a/config/homebin/copyxdebug
+++ b/config/homebin/copyxdebug
@@ -1,0 +1,13 @@
+#!/bin/bash
+sudo cp /srv/config/php-config/xdebug.ini /etc/php/7.0/mods-available/xdebug.ini
+sudo cp "/srv/config/php-config/php-custom.ini" "/etc/php/7.0/fpm/conf.d/php-custom.ini"
+
+  # Find the path to Xdebug and prepend it to xdebug.ini
+XDEBUG_PATH=$( find /usr/lib/php/ -name 'xdebug.so' | head -1 )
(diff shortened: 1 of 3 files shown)


LoreleiAurora:
why have you disabled xDebug remote autostart?

btsimonh:
hmm... not sure exactly (some time since i did this).  Looks like i was having issues with accessing the website just from a browser without a debugger started, and turned this off to make it work (maybe just more quickly?).  I'm using netbeans; from the comment in xdebug.ini, if the debug env sends (i think it's in the query string) something specific, then xdebug is turned on anyway. - may not be applicable to all debug environments?

jeremyfelt:
Hi @btsimonh, thanks for taking the time to open a pull request. I understand wanting to copy over configuration changes, but I don't think that this is necessary as part of the default configuration. At the moment, running `vagrant provision` will copy the latest versions of these files over. I could see creating a custom script to accomplish specific tasks like this on an individual basis. The new provisioning process that will be a part of the next VVV release should make it much easier to maintain these custom scripts and configurations.

PR #998, opened 2016-10-12
Update Vagrantfile

glensc:
set vim modeline to follow current formatting style


jeremyfelt:
Looks good, thanks @glensc!

PR #1004, opened 2016-10-20
Update to phpunit to major version 5

aaronjorbin:
PHPUnit 5.6 is the current stable release series.
It became stable on October 7, 2016.

It's also the only version officially supported by php7+ 

`PHPUnit 4.8 is supported on PHP 5.3, PHP 5.4, PHP 5.5, and PHP 5.6.`


jeremyfelt:
:+1:, thanks @aaronjorbin!

PR #1007, opened 2016-10-25
Switch to the Node.js 6.x LTS release

thebrandonallen:
On October 18th, with the release of Node.js 6.9.0, the 6.x branch was moved to LTS status. Seems like now is a good time to update our default version.


cfoellmann:
Sounds reasonable to me.

jeremyfelt:
Thanks for the PR @thebrandonallen. Pinging @ntwb as my goto source as to whether this could break the WP dev build in any way?

Issue #286, opened 2014-02-21
provision phpmyadmin via composer?


cfoellmann:
Maybe provisioning phpmyadmin via composer would save us updating every version by hand?

jeremyfelt:
This may deserve another look. After poking at #1012, which I'm going to close in a minute, there are a couple things we should answer before proceeding:

1. Will `composer create-project phpmyadmin/phpmyadmin:4.6.* database-admin --repository-url=https://www.phpmyadmin.net/packages.json --no-dev` work when there's an existing `database-admin/` directory?
2. How can this be configured to update whenever `vagrant provision` is run?

Once we have some conclusions on that approach, then the PR will need to be created against the [VVV Utilities](https://github.com/Varying-Vagrant-Vagrants/vvv-utilities/) repository, which is part of the new provisioning process.

Mte90:
I am studying what to do but seems that with composer is not possible to download a package with all the locales.

benlumia007:
is this still doable, if not, then we shouldn't need to worried bout this if we can't install phpmyadmin using composer ?

tomjn:
I'm not too worried about this now, I think we can close this out

PR #1012, opened 2016-10-29
Install phpMyAdmin via Composer

thebrandonallen:
This has been discussed a few times (#286 and #576), but for various reasons, it's never been commited.

Using Composer allows us maintain the current behavior of only installing on provision. We can control which x.y release we'd like to use, since new x.y versions can have PHP/MySQL requirement changes. But, on a subsequent provision, users can get the latest bugfix/security release without having to explicity update the provision script.

### DIFF
--- a/provision/provision.sh
+++ b/provision/provision.sh
@@ -684,10 +684,7 @@ phpmyadmin_setup() {
   if [[ ! -d /srv/www/default/database-admin ]]; then
     echo "Downloading phpMyAdmin..."
     cd /srv/www/default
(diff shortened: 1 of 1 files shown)


LoreleiAurora:
Can this be modified to install the latest stable every time provision is run?

cfoellmann:
We should go with `:*`
But the official repo won't provide translations. Just to put it out there

LoreleiAurora:
@cfoellmann just running this on my local machine with * i get the error 
```
[InvalidArgumentException]                       
  Project directory database-admin/ is not empty.
```

Wouldn't that men we would have to rm the directory on every provision?

jeremyfelt:
Thanks for the effort on this @thebrandonallen. I'm going to close it out for now as we have moved the provisioning of standard utilities to a new repository. I'd also like to answer the questions in my last comment on #286 before we proceed with a solution.

Issue #457, opened 2014-09-25
Update VM with security releases on any provision


enejb:
More info can be found here. http://www.bbc.com/news/technology-29361794

running `vagrant ssh` and then 
`env x='() { :;}; echo vulnerable' bash -c 'echo hello'` I get back 
vulnerable
hello

Which confirms the vulnerability. 
I fixed it on my virtual box by running 
`sudo apt-get update` and then `sudo apt-get --only-upgrade install bash` 
After the two commands I ran 
`env x='() { :;}; echo vulnerable' bash -c 'echo hello'` and got back
hello

jeremyfelt:
I'm going to hijack the original title for this and make it a feature request. :wind_chime:

Mte90:
We have our custom box right now so I don't think that we have to worry about it.
Yes the update of the box is still manual but is better for the user that don't want a new provision with a waiting to load everything.

tomjn:
Closing this out as anybody can run these commands to do it:

```
vagrant halt
vagrant destroy
vagrant box update
vagrant up --provision
```

PR #1014, opened 2016-10-31
Allow security updates to be installed on provision

thebrandonallen:
Also, installs security updates on a daily basis while the server is up.

Things to note:

1. Restarts are not forced, but may be required for situations like kernel upgrades.

2. Both nginx mainline and ppa:ondrej/php are also checked for updates, even though all updates aren't explicitly security updates.

See #457

### DIFF
--- a/config/apt/10periodic
+++ b/config/apt/10periodic
@@ -1,0 +1,4 @@
+APT::Periodic::Update-Package-Lists "1";
+APT::Periodic::Download-Upgradeable-Packages "1";
+APT::Periodic::AutocleanInterval "7";
+APT::Periodic::Unattended-Upgrade "1";
--- a/config/apt/50unattended-upgrades
+++ b/config/apt/50unattended-upgrades
@@ -1,0 +1,61 @@
+// Automatically upgrade packages from these (origin:archive) pairs
+Unattended-Upgrade::Allowed-Origins {
+	"${distro_id}:${distro_codename}-security";
+//	"${distro_id}:${distro_codename}-updates";
+//	"${distro_id}:${distro_codename}-proposed";
+//	"${distro_id}:${distro_codename}-backports";
+	"nginx:stable";
+	"LP-PPA-ondrej-php:trusty";
+};
+
+// List of packages to not update (regexp are supported)
+Unattended-Upgrade::Package-Blacklist {
+//	"vim";
+//	"libc6";
+//	"libc6-dev";
+//	"libc6-i686";
+};
+
+// This option allows you to control if on a unclean dpkg exit
+// unattended-upgrades will automatically run
+//   dpkg --force-confold --configure -a
+// The default is true, to ensure updates keep getting installed
+//Unattended-Upgrade::AutoFixInterruptedDpkg "false";
+
+// Split the upgrade into the smallest possible chunks so that
+// they can be interrupted with SIGUSR1. This makes the upgrade
+// a bit slower but it has the benefit that shutdown while a upgrade
+// is running is possible (with a small delay)
+//Unattended-Upgrade::MinimalSteps "true";
+
+// Install all unattended-upgrades when the machine is shuting down
+// instead of doing it in the background while the machine is running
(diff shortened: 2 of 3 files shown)


JJJ:
I really like this idea.

How does this play with `vagrant box update`? Will it end up keeping the internals of the box up-to-date enough where that won't be necessary, or will the container box version still be out-of-date?

jeremyfelt:
This is nice, thanks for the PR, @thebrandonallen.

> How does this play with vagrant box update? Will it end up keeping the internals of the box up-to-date enough where that won't be necessary, or will the container box version still be out-of-date?

I'd like this answered before we go much. Ideally, VVV would provide some sort of path for managing critical security updates, especially to packages it provides. At the same time, upgrading many core packages could cause bugs that are more painful than the security update was worth.

Two things we need to plan for are `vagrant box update` when new boxes are available from upstream and the use of `vagrant-vbguest` to keep guest additions updated. I'm not as worried about the second, but it is worth thinking about.

If we can be sure that `vagrant box update` and unattended security updates will go hand in hand, then I'm 👍  on merge.

thebrandonallen:
> How does this play with vagrant box update? Will it end up keeping the internals of the box up-to-date enough where that won't be necessary, or will the container box version still be out-of-date?

My understanding of the container boxes provided during `vagrant box update` (at least as far as Trusty is concerned), is that they are the equivalent of a nightly build of the latest Trusty point release. If one were to run `vagrant destroy && vagrant up` with a local the Trusty box version `20170310.0.0` on March 10, 2017, there should, in theory be no (or very few updated packages) that need to be pulled from the source list. However, if one were to perform `vagrant destroy && vagrant up` against that same Trusty box version `20170310.0.0` on March 13th, and the current version of box is `20170313.0.7` then on provision, any packages updated between `20170310.0.0` and `20170313.0.7` would be pulled from the source list repos.

Assuming this understanding is correct, then this PR shouldn't change much. Running `vagrant up --provision` or `vagrant provision` on an existing box would essentially the be the same as `vagrant destroy && vagrant up`, just with less time. Running an initial `vagrant up`, whether after destroy, or for the first time, always downloads the latest packages from the source list repos. Similarly, running `vagrant destroy && vagrant up` with a more recent box version should just mean one is downloading fewer packages during provision.

Again, assuming my initial understanding is correct, this PR shouldn't increase the risk of breakage any more than running `vagrant destroy && vagrant up` would at any other point. It may even allow one to catch breakage sooner.

As configured in this PR, `unattended-upgrades` is fairly conservative. The most "aggressive" part of `unattended-upgrades` is the fact that it will download all nginx mainline and ondrej/php updates, since they don't actually have security branches. The config doesn't automatically restart, so kernel upgrades won't become active until one runs `vagrant halt && vagrant up` or `vagrant reload`.

Sorry for the verbosity. I wasn't sure I understood the question, so I've tried to give as much detail about what's happening here as possible.

LoreleiAurora:
Personally, i don't like the idea of security updates while the box is up at random times. How will this play with a mobile data connection/on a train/on a plane?

thebrandonallen:
> Personally, i don't like the idea of security updates while the box is up at random times.

It's not completely random, as unattended-upgrades are run via cron.daily (so on could manage _when_ the upgrades are run), but I do understand that concern. We can also alter the config so that any security updates are only installed on shutdown, which would be on `vagrant halt` or `vagrant reload`.

As far as being on mobile, kernel upgrades are, by far, the heaviest, and we can turn those off in the unattended-upgrades config. However, if we're going to do security updates, I, personally, think kernel upgrades should be on priority.

JJJ:
Maybe we can ship with an alias to toggle auto-updates on/off, similar to XDebug?

thebrandonallen:
Does anyone have a preference on script name?

Off the top of my head something like `auto_updates_(on|off)` would work.

tomjn:
Closing this out as security updates will get installed when the box is rebuilt in VVV 3, a user need only pull down the newer VVV box

PR #1017, opened 2016-11-04
Allow vvv.dev access in feature/split-provisioners.

LoreleiAurora:
This restores access to the dashboard by explicitly adding it do the domain array in the Vagrantfile.


jeremyfelt:
:+1:

PR #1021, opened 2016-11-04
Split out utilities into a different repository

LoreleiAurora:
This adds support for provisioning the tools/resources from another repo using the config file.

```YAML
// Defining additional repositories
utility-repos:
  unofficial: https://github.com/LoreleiAurora/vvv-utilities-2.git
  unofficial2: https://github.com/LoreleiAurora/vvv-utilities-3.git

// Defining utilities to install
utilities:
  official:
  // Utilities in the official repository
    - memcached-admin
    - opcache-status
    - phpmyadmin
    - webgrind
  unofficial:
  // Utilities in the unofficial repository defined above
    - test1
    - test2
    - test3
  unofficial2:
  // Utilities in the unofficial2 repository defined above
    - test4
    - test5
    - test6
```

The utilities defined in unofficial repositories just echo there names as an example/testing


LoreleiAurora:
@jeremyfelt can you please create a vvv-utilities repository so i can split out the provisioning.

LoreleiAurora:
Is there anything else that should be moved to the other repo? (maybe mailcatcher)

LoreleiAurora:
@jeremyfelt needs-feedback needs-testing

jeremyfelt:
Thanks @LoreleiAurora. I think mailcatcher would be good as well. Anything that could be considered "optional" as part of the default web directory makes sense.

At first glance I like the layout of the new YAML options. I'm going to keep giving it some thought. :)

jeremyfelt:
I'm wondering about the term `official` here. What do you think about something like `core` or `vvv-core`?

Also, how about `utility-sources` rather than `utility-repos`. One day it may make sense to support a tar/zip or local directory here.

Other than that, :+1:!

LoreleiAurora:
Those terms sound better to me, will make the changes now.
As for tar/zip support:
- i am not entirely sure how mush this would be used
- that could be a pain considering we cant use gems not bundled with Vagrant (without creating a vagrant plugin) as we cant guarantee someone will have ruby installed and gems would be another installation step.

jeremyfelt:
> that could be a pain considering we cant use gems not bundled with Vagrant

Ahh, good point. We'll at least leave the door open, however unlikely. :smile:

PR #1027, opened 2016-11-08
Remove old WordPress Config `mu-plugins` directory.

JJJ:
The included plugin is no longer necessary, leaving the directory empty.

See 4addb21bd81828466c11baaa77f9c21ea235b700.


JJJ:
Thanks!

PR #1030, opened 2016-11-10
Add hosts for hostsupdater plugin

kkatano:


JJJ:
Hey @kkatano. I think I understand what problem you're hoping to solve, but I think there this isn't going to be the correct approach in the long term.

A few things this misses:

* The multisite repos
* The WordPress.org meta sites
* Custom websites

For this to be complete, will require the hosts updater to be more aware of what sites are enabled, which I'm pretty sure my VVV checkouts have always done a pretty good job at doing already.

Can you explain more about why you think this change is necessary? Maybe I can help test this or other solutions.

LoreleiAurora:
@kkatano @JJJ This is something ive neglected to add to the new config file, will fix this over the next 24 hours.

LoreleiAurora:
@kkatano @JJJ The VVV hosts files have moved into the individual site dirs and are available after the first provision, therefore you just have to reprovision. Im trying to think of a better way to deal with this.

kkatano:
@JJJ @LoreleiAurora 
Thanks, i see.
I tried re-provision and got fine result.

PR #1040, opened 2016-11-25
deleted local.wordpress-trunk.dev

dutchmichael:
To my knowledge, local.wordpress-truck.dev is not loaded in the current version of VVV.
So, I deleted it in the file /www/vvv-hosts


jeremyfelt:
Hi @dutchmichael, thanks for the PR. This was purposely left in `vvv-hosts` for backward compatibility as we don't actually delete the directory or the config, it just isn't setup on new provisions.

Issue #983, opened 2016-08-31
Unable to update wp-cli on provision


jjeaton:
In provision.sh, wp-cli gets [updated like this](https://github.com/Varying-Vagrant-Vagrants/VVV/blob/develop/provision/provision.sh#L591):

```
cd /srv/www/wp-cli
git pull --rebase origin master
composer update
```

This will work the first time that wp-cli is updated on provision. However, `composer update` modifies the `composer.lock` file, so now there are unstaged changes in the git checkout. The next time you provision, the git pull fails due to unstaged changes.

Would it be better to install wp-cli using the normal install method as suggested here: https://github.com/Varying-Vagrant-Vagrants/VVV/issues/82#issuecomment-19367345?

theukedge:
Can someone smarter than me help me to update wp-cli in the interim?

jjeaton:
@theukedge 

you can try this:

```
vagrant ssh
cd /srv/www/wp-cli
git checkout -- .
git pull --rebase origin master
composer update
```

theukedge:
@jjeaton you're a legend. Thank you sir

jeremyfelt:
Thanks for opening an issue @jjeaton. I think it would be nice to switch the provisioning for WP-CLI to the "official" process. For those that contribute to directly to WP-CLI, it should be relatively painless to setup a custom config. For those that don't, it will be much easier to keep updated with `wp cli update --nightly`.

jjeaton:
@jeremyfelt I've opened a PR #1057 that makes this change. Although, after seeing #990, that fixes the core issue. I agree that the standard installation will make wp-cli updates easier for most users.

Issue #82, opened 2013-06-12
Update WP-Cli on Provision


Mamaduka:
We should handle WP-Cli installation same way as Composer (851ac8cb908d7d1507bb7add5c8a659dbee309b5), so it can be updated on provision.

Currently we're cloning master branch so simply we can add `git pull origin`, if wp-cli directory already exists.

PR #1041, opened 2016-11-26
Run composer install after WP-CLI update

grappler:
If we run `composer update` then it will update the `composer.lock` which we don't want. This may also be the cause of issues in the past.

Once the `composer.lock` has been updated in the WP-CLI update then we just need to install the newest versions with `composer install`.

Related: #82, #983


grappler:
Close in favour of #990

Issue #927, opened 2016-06-23
Network connection not detected. Unable to reach google.com...


linjielig:
In china, I can't access google.com then the `vagrant up` can't complete. I have to adjust _google.com_ to _baidu.com_ or other.

jeremyfelt:
Hi @linjielig, thank you for opening a ticket. I agree, we should be able to find a service that is more likely to be accessible throughout the world.

claudiocoratolo:
i have the same problem. I fix it by commenting the part that check the google ping connection
`network_check() {
  network_detection
  if [[ ! "$ping_result" == "Connected" ]]; then
    echo -e "\nNo network connection available, skipping package installation"
	echo -e "\nPERO CONTINUO PORQUE SI TENGO CONECCION!!!!"
    **#exit 0**
  fi
}`

claudiocoratolo:
i have the same problem. I fix it by commenting the part that check the google ping connection on provision.sh
network_check() {
       network_detection
       if [[ ! "$ping_result" == "Connected" ]]; then
           echo -e "\nNo network connection available, skipping package installation"	
           **#exit 0** 
       fi
}

mikromakro:
Yeah Google is one thing, but google.analytics is required before google.com. Not very nice.

mrben522:
I was getting this error in the US on a brand new MBP.  Added 8.8.8.8 to DNS servers in network settings fixed it for me

rumur:
I have the same problem and I've increased attempts and timeout for connecting the google.com
<pre>if [[ "$(wget --tries=10 --timeout=10 --spider http://google.com 2>&1 | grep 'connected')" ]]; then</pre> 
that's helped me

jeremyfelt:
In #1048, the `wget` call has been changed to `wget --tries=3 --timeout=5 --spider --recursive --level=2 http://google.com 2>&1 | grep 'connected'`, which should help here. If not, we may need to find a more reliable domain than `google.com` to start with.

PR #1048, opened 2016-12-08
Improves Network Connection Detection

dsawardekar:
Google.com redirects to a local server like Google.co.in based on geolocation. Without recursion the wget call does not detect such redirects. That further results in major provisioning errors since the `network_check` call happens at multiple stages.

This PR follows such redirects upto 2 levels. I've tested provision over mobile networks with and without this PR. Network connection detection is much more reliable with this recursion flag enabled.


jeremyfelt:
Thanks for the PR, @dsawardekar! At first glance it looks good. I believe this will address #927 once merged.

PR #1057, opened 2016-12-15
Change wp-cli to standard installation

### LINKED ISSUE #983: Unable to update wp-cli on provision

jjeaton:
Previously, VVV was cloning the wp-cli repo and installing
composer dependencies. This switches to the standard wp-cli
installation method which will make updating wp-cli easier.

`wp cli update` will run on re-provision.

Users can run `wp cli update` at any time
or `wp cli update --nightly` to update to the latest nightly.

Fixes #983.

### DIFF
--- a/provision/provision.sh
+++ b/provision/provision.sh
@@ -559,19 +559,27 @@ services_restart() {
 
 wp_cli() {
   # WP-CLI Install
-  if [[ ! -d "/srv/www/wp-cli" ]]; then
+  local exists_wpcli
+
+  # Remove old wp-cli symlink, if it exists.
+  if [[ -L "/usr/local/bin/wp" ]]; then
+    echo "\nRemoving old wp-cli"
+    rm -f /usr/local/bin/wp
+  fi
+
+  exists_wpcli="$(which wp)"
+  if [[ "/usr/local/bin/wp" != "${exists_wpcli}" ]]; then
     echo -e "\nDownloading wp-cli, see http://wp-cli.org"
-    git clone "https://github.com/wp-cli/wp-cli.git" "/srv/www/wp-cli"
-    cd /srv/www/wp-cli
-    composer install
+    curl -sO https://raw.githubusercontent.com/wp-cli/builds/gh-pages/phar/wp-cli-nightly.phar
+    chmod +x wp-cli-nightly.phar
+    sudo mv wp-cli-nightly.phar /usr/local/bin/wp
+
+    # Install bash completions
+    curl -s https://raw.githubusercontent.com/wp-cli/wp-cli/master/utils/wp-completion.bash -o /srv/config/wp-cli/wp-completion.bash
   else
     echo -e "\nUpdating wp-cli..."
-    cd /srv/www/wp-cli
-    git pull --rebase origin master
-    composer install
+    wp --allow-root cli update --nightly --yes
   fi
-  # Link `wp` to the `/usr/local/bin` directory
-  ln -sf "/srv/www/wp-cli/bin/wp" "/usr/local/bin/wp"
 }
 
 php_codesniff() {
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,9 @@
 # Ignore a custom bash_prompt if it exists
 /config/bash_prompt
 
+# Ignore downloaded wp-cli bash completions
+/config/wp-cli/wp-completion.bash
+
 # No need to share our mysql data with each other
 /database/data/*
 
--- a/config/bash_profile
+++ b/config/bash_profile
@@ -36,7 +36,7 @@ export WP_CORE_DIR=/srv/www/wordpress-develop/src/
 eval "$(grunt --completion=bash)"
 
 # add autocomplete for wp-cli
-. /srv/www/wp-cli/utils/wp-completion.bash
+. /srv/config/wp-cli/wp-completion.bash
 
 # PHPCS path
 export PATH="$PATH:/srv/www/phpcs/scripts/"


provision/provision.sh
```diff
@@ -559,19 +559,18 @@ services_restart() {
...
-    composer install
+    curl -O https://raw.githubusercontent.com/wp-cli/builds/gh-pages/phar/wp-cli.phar
+    chmod +x wp-cli.phar
+    sudo mv wp-cli.phar /usr/local/bin/wp
   else
     echo -e "\nUpdating wp-cli..."
-    cd /srv/www/wp-cli
-    git pull --rebase origin master
-    composer install
+    wp cli update --yes
```
LoreleiAurora: Can you change this to the nightly instead.
This way if there are any changes in WP Develop/Trunk they will be patched in CLI quicker.

provision/provision.sh
```diff
@@ -559,19 +559,18 @@ services_restart() {
...
-  if [[ ! -d "/srv/www/wp-cli" ]]; then
+  local exists_wpcli
+
+  exists_wpcli="$(which wp)"
+  if [[ "/usr/local/bin/wp" != "${exists_wpcli}" ]]; then
     echo -e "\nDownloading wp-cli, see http://wp-cli.org"
-    git clone "https://github.com/wp-cli/wp-cli.git" "/srv/www/wp-cli"
-    cd /srv/www/wp-cli
-    composer install
+    curl -O https://raw.githubusercontent.com/wp-cli/builds/gh-pages/phar/wp-cli.phar
```
LoreleiAurora: As below

provision/provision.sh
jjeaton: Yep, good call.

provision/provision.sh
```diff
@@ -559,19 +559,18 @@ services_restart() {
...
-    composer install
+    curl -O https://raw.githubusercontent.com/wp-cli/builds/gh-pages/phar/wp-cli-nightly.phar
+    chmod +x wp-cli.phar
+    sudo mv wp-cli.phar /usr/local/bin/wp
   else
     echo -e "\nUpdating wp-cli..."
-    cd /srv/www/wp-cli
-    git pull --rebase origin master
-    composer install
+    wp cli update --nightly --yes
```
jeremyfelt: What happens here if provisioning is run on a setup that was previously using the `git` clone?

LoreleiAurora: It wouldnt work, but neither will most of the changes in v2

LoreleiAurora:
Bash completion will not work with this, If you are unsure how to repair this allow commits from maintainers and i can jump in and fix this.

jjeaton:
@LoreleiAurora edits should be allowed. Yeah I wasn't sure how to handle provisioning existing installs.

I'm happy to take a stab at fixing the completions as well, but if you are raring to go on it right now, feel free! 😄

provision/provision.sh
jjeaton: Would we want to add a check for the `/srv/www/wpcli` dir and attempt the earlier version of the upgrade here if it exists instead?

LoreleiAurora:
Im on mobile, so cant do that now :)

provision/provision.sh
LoreleiAurora: If we add something like:
```
if [[ -L "/usr/local/bin/wp" ]]; then
  echo "Removing old wp-cli"
  rm -rf /usr/local/bin/wp /srv/www/wp-cli
fi
```
before `if [[ "/usr/local/bin/wp" != "${exists_wpcli}" ]]; then` it should migrate people seemlessly.

THE ABOVE SNIPPET IS UNTESTED.

jjeaton:
Added installation of wpcli bash completions and updating them. Although I'm not sure whether it should be redownloaded on every provision, as that file has only had one change in 2 years?

I can look more into how we can upgrade existing users later on.

LoreleiAurora:
@jjeaton i added a comment [further up](https://github.com/Varying-Vagrant-Vagrants/VVV/pull/1057#discussion_r92657841) explaining how i would do it.

provision/provision.sh
jeremyfelt: I'd like to see if we can look for `/srv/www/wp-cli/.git` if `/usr/local/bin/wp` exists and then reinstall with the phar. We can include - remove `/srv/www/wp-cli` - as part of the upgrade instructions, but it would be nice if it accounted for this as an upgrade procedure.

LoreleiAurora: My snippet above only does rm if /usr/local/bin/wp is a symlink, does that not achieve tha same goal?

jeremyfelt: Huh, GitHub was hiding your previous comment from me before. :)

That should do the trick, though I'd leave `/srv/www/wp-cli/` alone just in case somebody was using that to contribute and had branches.

jjeaton: I'll make this change and test 👍

jjeaton:
@LoreleiAurora updated to remove just the wp symlink if it exists, so existing users get upgraded to the standard install, silenced the curl requests, and removing the updating of the bash completions.

I've squashed the commits and this is ready for review again.

Issue #1044, opened 2016-12-05
Make setting up a custom site EASY. (How to do it right now?)


nextgenthemes:
Please provide me with simple steps how to have a custom site with wordpress trunk and its own database setup.

I will keep it short there seems to be a new provisioning system and I just after trying various things I failed to set up a custom local domain. I keep it short. I read the wiki, tried to setup a new site and failed.

I also tried the vv site setup tool, it failed, also current build shows as failed on github. I think such a tool, at leasts the basics of it should actually be part if this project I think. Like a simple command to setup a new site for example.

There seems to be new provisioning system involving the yml config file with documentation in the works.

I am a technical person and usually have no problems figuring things out but I don't know whats new way whats the old way. It seems new way needs to be on top of the old way. It seems to me its a pain in the ass to just setup another site. I think it should be actually made **easy!**

Maybe I am missing something and its actually easy but ...

It seems I need to create/copy/edit 3 different files to just setup a new site? Is this right? https://github.com/Varying-Vagrant-Vagrants/VVV/wiki/Auto-site-Setup and the new provisioning yml. I do not know what to put in there. I setup `vm_dir: '/srv/www/arve-dev-provisioner' for a arve_dev: site I want to setup but it says the directory does not exist during provisioning.

Here is my proposal how to create a site.

```
  arve-dev:
    domain: arve.dev
    wp_version: nightly
```

Simple as that, database folders and stuff should be created automatically.

kopepasah:
@nextgenthemes this may be related to #1045. Can you try checking out the latest tag (currently `1.4.0`) and see if you still experience this issue?

nextgenthemes:
@kopepasah Its not about a "issue" I asked various specific things ...

LoreleiAurora:
@nextgenthemes 
- If you have a vvv-config.yml you are not on 1.4.0 you are on the develop branch.
- 1.4.0 is available on the master beanch or as a tag.
- The VV setup tool will work with 1.4.0.
- The docs in the wiki work for all versions up to and including 1.4.0.
- For as long as VVV has supported custom sites it has required 3 files. (an example of these files can be found [here](https://github.com/Varying-Vagrant-Vagrants/vvv-wordpress-default))
- You create your own provisioning scripts as alot of people have different requirements for alot of sites.

If you downgrade to 1.4.0 then you can just base your config off my example above.

[...]

Issue #1034, opened 2016-11-16
Local .dev sites no longer rendering


mkjonesuk:
Did a vvv re-build recently, new VirtualBox, Vagrant, new local folder etc.

After a few hours of suffering I still can't get the custom .dev domains I had working previously (via 'Auto site Setup') to render. 

VirtualBox: `5.1.8 r111374 (Qt5.5.1)`
Vagrant: `Vagrant 1.8.6` (1.8.7 was tried and removed as an attempt to fix)
VVV: `latest` 

To be clear, when I visit `coolsite.dev` this would normally load the custom site, setup using `www/coolsite/vvv-hosts` - but instead I'm just getting the basic `vvv.dev` welcome page.

All of the .dev domains **are** being added to my /hosts file - they also remove on reload/provision as standard.

All I can work out at this hour is that once the provision goes through the business of creating the built-in WP test sites it skips to the end of the provision rather than creating the custom .dev sites, see:

```
==> default: Cleaning the virtual machine's /etc/hosts file...
==> default: -----------------------------
==> default: Provisioning complete in 139 seconds
==> default: For further setup instructions, visit http://vvv.dev
==> default: Running provisioner: site-wordpress-default (shell)...
    default: Running: /var/folders/6n/4pcgv2ps39g5rfqcx8dsd6bm0000gn/T/vagrant-shell20161116-633-dd2z5y.sh
==> default: Updating wordpress-default...
==> default: From https://github.com/Varying-Vagrant-Vagrants/vvv-wordpress-default
==> default:  * branch            master     -> FETCH_HEAD
==> default: Already up-to-date.
==> default: Already on 'master'
==> default: Your branch is up-to-date with 'origin/master'.
==> default: Creating database 'wordpress_default' (if it's not already there)
==> default:  DB operations done.
==> default: Updating WordPress Stable...
==> default: Error: The site you have requested is not installed.
==> default: Run `wp core install`.
==> default: Adding domains to the virtual machine's /etc/hosts file...
==> default:  * Added local.wordpress.dev from /srv/www/wordpress-default-provisioner/vvv-hosts
==> default: Running provisioner: site-wordpress-develop (shell)...
    default: Running: /var/folders/6n/4pcgv2ps39g5rfqcx8dsd6bm0000gn/T/vagrant-shell20161116-633-1pbuepb.sh
==> default: Updating wordpress-develop...
[...]

Issue #1033, opened 2016-11-15
Incomplete first 'vagrant up', *-provisioner directories


ihorvorotnov:
Decided to update the stack by completely uninstalling everything and installing from scratch. Never had any issues before, but today everything went fine until the first `vagrant up`. It went as usual, showing some npm warnings etc - nothing wrong from the first sight. However, when it went up, only `vvv` and `vvv.dev` domains were added to `/etc/hosts` and in `www` directory I see couple extra directories:

- default
- phpcs
- wordpress-default
- **wordpress-default-provisioner**
- wordpress-develop
- **wordpress-develop-provisioner**
- wp-cli

Those `-provisioner` directories I see for the first time. And from what I understand, it's missing one more standard WP install - `wordpress-trunk`.

If I halt vagrant and up it again, the domains are added to `/etc/hosts` and both `wordpress-default` and `wordpress-develop` installations work. However, those `-provisioner` dirs are still there. Any ideas what went wrong?

<img width="628" alt="screen shot 2016-11-15 at 22 41 12" src="https://cloud.githubusercontent.com/assets/1066083/20323407/3fec678a-ab85-11e6-8da2-0d9b97021f01.png">

ihorvorotnov:
Re-installed again, used VVV stable zip, works fine.

LoreleiAurora:
Nothing has gone wrong, thats the start of the efforts to split provisioners into different repos, see #980

Issue #963, opened 2016-08-16
Refactor the README


jeremyfelt:
The README has gotten pretty darn long.

With the new and improved setup for documentation via varyingvagrantvagrants.org, we should be able to trim this down and link off to more detailed instructions where necessary.

jeremyfelt:
This is fixed in #1073.

PR #1060, opened 2016-12-19
clarify recommended version for new users

skorasaurus:
Hi, 
As a first-time user to VVV, I was unsure which branch to use and although I'm somewhat adept technically, it had taken some time to figure out (after reading several closed issues; especially https://github.com/Varying-Vagrant-Vagrants/VVV/issues/1034#issuecomment-266656093) that there were significant changes to the provisioning system between the develop branch and the latest stable release. 

I understand that stable is generally recommended for newer users of software, but I'd appreciate a more firmer guidance for newer users. I have modified the README to strongly recommend to use 1.4.0 instead of develop. 

Very recently, several other users (https://github.com/Varying-Vagrant-Vagrants/VVV/issues/1033 , https://github.com/Varying-Vagrant-Vagrants/VVV/issues/1034 , https://github.com/Varying-Vagrant-Vagrants/VVV/issues/1044 ) have also been confused when they first started out with the develop branch instead of the stable release. 

(That https://github.com/Varying-Vagrant-Vagrants/VVV/  loads the develop branch on loading by default isn't helpful for newer users; having it as the landing page reinforces that the develop branch is what should be first seen by any new users...)


jeremyfelt:
Hi @skorasaurus, thanks for taking the time to open a PR with your suggestions. I think you're on the right track. I want to make sure this is tackled as part of the work on #963 and #961. This will be more of a sweeping change than the smaller targeted changes proposed here. We'll get there though!

PR #1064, opened 2016-12-22
Clone repos recursively in new provisioning system.

kopepasah:
This fixes #1063.

PR #1065, opened 2016-12-22
Simplify clone and checkout of initial branch in new provisioning system.

kopepasah:
By using `--branch ${BRANCH}`, the specific branch will be cloned and checkout.


kopepasah:
CC @jeremyfelt @LoreleiAurora

jeremyfelt:
This looks good too, thanks @kopepasah. I updated the PR after #1064 was merged in so that both `--recursive` and `--branch` were handled.

PR #1071, opened 2017-01-04
Custom keys for custom provisioning in vvv-init.sh

tomjn:
The new `vvv-config.yml`/`vvv-custom.yml` allows you to specify a site or sites, with various keys. However, a `vvv-init.sh` may want to specify custom keys that are too specific to apply to all VVV sites.

This PR installs `python-pip` and the `shyaml` dependency, and adds a bash function named `customvalue` for retrieving those keys. When the `custom` section or the requested key are not defined or found, the function will return nothing

E.g. in `vvv-init.sh`

```
colour=`customvalue 'colour' 'pink'`
echo "Hello there :) This site is a ${colour} site"
```

Which when ran will print `Hello there :) This site is a pink site`, and can be changed as follows in `vvv-custom.yml`:

```
  redsite:
    custom:
      colour: red
```

Resulting in `Hello there :) This site is a red site`

Nested keys can be accessed using dot syntax, e.g.:

```
  redsite:
    custom:
      colour:
        primary: red
        secondary: blue
```

```
primarycolour=`customvalue 'colour.primary' 'pink'`
echo "Hello there :) This site is a ${primarycolour} site"
```

### DIFF
--- a/provision/provision-site.sh
+++ b/provision/provision-site.sh
@@ -15,7 +15,7 @@ noroot() {
 
 # Takes 2 values, a key to fetch a value for, and an optional default value
 # e.g. echo `customvalue 'key' 'defaultvalue'`
-customvalue() {
+get_config_key() {
   local config=/vagrant/vvv-config.yml
   if [[ -f /vagrant/vvv-custom.yml ]]; then
     config=/vagrant/vvv-custom.yml
--- a/provision/provision-site.sh
+++ b/provision/provision-site.sh
@@ -14,8 +14,8 @@ noroot() {
 }
 
 # Takes 2 values, a key to fetch a value for, and an optional default value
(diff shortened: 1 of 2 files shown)


jeremyfelt:
This is really slick, thanks @tomjn!

I'd like to revisit the name `customvalue` as I think it can be more descriptive. How about something like `get_config_key()` or `get_yaml_option`? Happy to hear any other ideas as well.

tomjn:
I like `get_config_key`, an obvious next step in functionality here would be to return values rather than a singular value, so `get_config_value` makes more sense, as it leads to the  `get_config_values` function

tomjn:
I've renamed the function to `get_config_value`

jeremyfelt:
Great stuff, thanks @tomjn!

PR #1076, opened 2017-01-04
[PR] Add the php-memcached package to default provisioning

### LINKED ISSUE #1067: Provide php-memcached in default provisioning
I'm not familiar with all of the details, but it seems that the `php7.0-memcache` package we install during provisioning doesn't really all that well. If it does work, then it at least does not work with the WordPress core unit tests and [batcache](https://wordpress.org/plugins/batcache) as the drop-in.

I think we can install `php-memcached` alongside the `php7.0-memcache` package. With this available, we can use the [PECL memcached](https://github.com/tollmanz/wordpress-pecl-memcached-object-cache) drop-in instead.

jeremyfelt:
This appears to be in better shape for PHP7 than php7.0-memcache
and provides php5.5-memcached, php5.6-memcached, php7.0-memcached,
and php7.1-memcached.

Fixes #1067.

### DIFF
--- a/provision/provision.sh
+++ b/provision/provision.sh
@@ -39,6 +39,7 @@ apt_package_check_list=(
 
   # Extra PHP modules that we find useful
   php-pear
+  php-memcached
   php7.0-imagick
   php7.0-memcache
   php7.0-bcmath


LoreleiAurora:
For the sake of continuity can we not refer to it as its alias name `php7.0-memcached`

PR #1080, opened 2017-01-04
Add hosts defined in config file to machine host file

LoreleiAurora:


jeremyfelt:
This looks good, @LoreleiAurora. Can you update it so that there are no changes to `provision.sh` now that #1071 is merged?

LoreleiAurora:
Updated :)

PR #1110, opened 2017-02-24
Adds a VVV2 logo to provisioning

tomjn:
Also adds a check to prevent triggers outputting the header multiple times

Here's the result:

screen shot 2017-02-24 at 15 09 35
screen shot 2017-02-24 at 15 14 10

The colours don't show in Powershell, but instead gracefully reverts to Monochrome. Tested on Windows and MacOS


tomjn:
I like mentioning the stability, though I would keep the version number for
completeness sake so that people know the significance. I'll reword
tomorrow/Monday

tomjn:
screen shot 2017-03-07 at 01 36 10

This is the current version. I'm not fond of adding mentions of stable because right now there is no stable VVV 2 other than the `develop` branch, and advising checking out `master` will induce a downgrade that requires killing the box and starting from scratch to recover from.

I'm also mindful that this is hardcoded and will require manual work everytime somebody wants to merge `develop` into `master` to rename it to stable

jeremyfelt:
Good call, @tomjn. I think the current version is :100:. Thanks!

PR #1136, opened 2017-03-13
[PR] Continued docs work for 2.0.0

jeremyfelt:


LoreleiAurora:
Note that you can't use `--provision-with` during initial migration or when installing PHP versions. This **WILL** cause breakage.

PR #1155, opened 2017-03-19
Run git as the regular vagrant user when provisioning

felixarntz:
The `git` commands to clone and update a repository are currently run as the root user which can be problematic. They should preferably be run by the regular vagrant user.

This applies both to site provisioning as well as retrieving custom utility repositories.

### DIFF
--- a/provision/provision-site.sh
+++ b/provision/provision-site.sh
@@ -44,12 +44,12 @@ if [[ false != "${REPO}" ]]; then
   # Clone or pull the site repository
   if [[ ! -d ${VM_DIR}/.git ]]; then
     echo -e "\nDownloading ${SITE}, see ${REPO}"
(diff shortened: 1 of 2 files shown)


jeremyfelt:
I wonder if we can check permissions on one file and only `chown` in that case. Is that too dangerous to rely on?

`chown` may also take a while on some repos.

tomjn:
I'm inclined to recreate this PR with 2 separate ones for utilities and site provisioners, but a check via `vagrant ssh` indicates the files and folders in `/srv/www` are all owned by `vagrant` and `www-data`, so the permissions issue isn't so big a problem

tomjn:
screen shot 2018-05-08 at 19 35 07

tomjn:
Closing in favour of 

https://github.com/Varying-Vagrant-Vagrants/VVV/pull/1491
https://github.com/Varying-Vagrant-Vagrants/VVV/pull/1490

PR #1156, opened 2017-03-19
Run composer as the regular vagrant user when provisioning

felixarntz:
Composer should not be run as the root user. This currently produces warnings when provisioning.


tomjn:
What can we do to switch the files already checked out by composer from the `root` user to the `vagrant` user so existing installs aren't broken?

tomjn:
Closing in favour of #1489

Issue #922, opened 2016-06-14
Install PHPCS globally with composer rather than git


jeremyfelt:
Via #608, VVV doesn't necessarily need to provide PHP_Codesniffer via git clone and can instead use Composer to install it globally.

It would be interesting to hear from anyone working on the [WordPress Coding Standards](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards) project as to whether its helpful to have the PHP_Codesniffer git repository available. /cc @westonruter, @GaryJones 😄

westonruter:
I would say it is not really necessary to have it installed by default as a Git clone. It is rare that I need to change out the PHPCS version when developing for WPCS.

tomjn:
Closing as it's already been implemented

PR #1169, opened 2017-04-02
provision phpcs + wp sniffs via composer

cfoellmann:
Relates to #922 

Introduce composer installation of phpcs and the sniffs for WP and keep it in the right place.
Removes 2 git operations per provision.

### DIFF
--- a/provision/phpcs/composer.json
+++ b/provision/phpcs/composer.json
@@ -1,0 +1,15 @@
+{
+    "name": "VVV/phpcs",
+    "type": "project",
+    "require": {
+        "oomphinc/composer-installers-extender": "^1.1",
+        "wp-coding-standards/wpcs": "*"
+    },
+    "extra": {
+        "installer-types": ["library"],
+        "installer-paths": {
+            "/srv/www/phpcs/": ["squizlabs/php_codesniffer"],
+            "/srv/www/phpcs/CodeSniffer/Standards/WordPress/": ["wp-coding-standards/wpcs"]
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


travisseitler:
(Anecdotal, but I merged this patch on my machine and it appears to also address the phpcs provisioning issue at #1179.)

thebrandonallen:
`"squizlabs/php_codesniffer": "*"` will need to be changed to `"squizlabs/php_codesniffer": "2.*"`, since the WordPress Coding Standards sniffs are not compatible with 3.0.

cfoellmann:
@thebrandonallen made the change to phpcs 2.x

provision/phpcs/composer.json
```diff
@@ -0,0 +1,16 @@
+{
+    "name": "VVV/phpcs",
+    "type": "project",
+    "require": {
+        "oomphinc/composer-installers-extender": "^1.1",
+        "squizlabs/php_codesniffer": "2.*",
```
TangRufus: I think requiring just `wp-coding-standards/wpcs` is enough, as `wpcs` requires `squizlabs/php_codesniffer`

cfoellmann: you are right. Fix for the PR is coming

cfoellmann:
squashed everything (not best practice, i know).
Works on my install. review would be nice.

tomjn:
This PR broke `PATH` access to `phpcs` and diverged old and new machines by changing the folders the executables are saved to:

old: `/srv/www/phpcs/scripts/phpcs`
new: `/srv/www/phpcs/bin/phpcs`

PR #1178, opened 2017-04-19
Correct Hyper-V Box Template

jdvachal:
The previous trusty64 box is no longer available.  Replaced with withinboredom/Trusty64


jeremyfelt:
Thanks for the PR @jdvachal. Let's go with the Hyper-V box in #1307 instead as it's based on the one we're currently using.

Issue #595, opened 2015-03-18
Provisioning is sloooow


joelworsham:
Okay, I don't know much about how Vagrant provisioning works, but is there ANY way to provision just one site install? Or is there any way to mimick this? Having to provision the whole thing after installing one site can be quite time consuming when you have 10+ sites.

Any alternative methods even?

BronsonQuick:
@joelworsham If you just want a one site install that's super fast to spin up and provision checkout https://github.com/Chassis/Chassis/ it'll probably suit you needs better from the sound of it.

cfoellmann:
@joelworsham do you mean the first `vagrant up`? Yes, that is super slow. Or just a `vagrant provision` to add an additional site?

joelworsham:
vagrant provision. I've since done some workarounds to only run a couple sites at a time to help provisioning take less time.

cfoellmann:
this shouldn't be. What extensive provisioning are doing per site?
When you run  `vagrant provision` on a fresh VVV directly after the initial `vagrant up` it should take about ~30-60 seconds since all checks for updates run. If I misunderstood you and you mean the initial provisioning of all services and packages I feel your pain.
I have some PRs up that can reduce provision time drastically but progress is slow

joelworsham:
I don't mean the initial provisioning. I'm referring to once I've got a few sites (currently about 10)

EHLOVader:
I agree it has been dreadfully slow and seems to do unnecessary installs or checks each time. For every site.

jonathan-dejong:
+1 on this. VVV and vagrant is awesome but now that I've got 20+ projects running provisioning is very slow. Luckily I of course don't have to provision every day but it sort of takes away the "quick start" of a new project.

topdown:
I also see this issue. At first I thought it was the custom site creation scripts, but I noticed all of the additional checks just running ```vagrant provision``` 

Isn't there a PR being worked on for the provision.sh? Will that address some of this performance issue?

jeremyfelt:
It is slooooow. :) I'd like to see some better handling of incremental updates rather than the blanket provisioning process. IMO, provisioning is probably best for updating the server. Making quick changes to add a site could probably go just fine without it if we prepped the right way.

devster31:
[...]

PR #1182, opened 2017-04-26
Speed up checking for `vvv-hosts` files

davidpagepsycle:
The existing recursive blob causes massive slowdown of everything (provision, ssh etc) once I've got a few projects setup, especially if they have numerous directories.

This change limits the lookups to only look at the primary directory and a `provision` subfolder rather than unlimited depth. This matches what is documented, though other code such as provision-site.sh still looks [4 levels deep](https://github.com/Varying-Vagrant-Vagrants/VVV/blob/develop/provision/provision-site.sh#L93).

On my setup doing `vagrant ssh` returns in seconds rather than potentially minutes. Should be useful in situations such as https://jjj.blog/2016/05/slow-vagrant/

Related: #595

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -41,7 +41,7 @@ end
 
 vvv_config['hosts'] += ['vvv.dev']
 
-host_paths = Dir[File.join(vagrant_dir, 'www', '*', 'vvv-hosts'), File.join(vagrant_dir, 'www', '*', 'provision', 'vvv-hosts')]
+host_paths = Dir[File.join(vagrant_dir, 'www', '*', 'vvv-hosts'), File.join(vagrant_dir, 'www', '*', '*', 'vvv-hosts'), File.join(vagrant_dir, 'www', '*', '*', '*', 'vvv-hosts'), File.join(vagrant_dir, 'www', '*', '*', '*', '*', 'vvv-hosts')]
 
 vvv_config['hosts'] += host_paths.map do |path|
   lines = File.readlines(path).map(&:chomp)
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -64,8 +64,7 @@ vvv_config['sites'].each do |site, args|
 
   vvv_config['sites'][site] = defaults.merge(args)
 
-  site_local_dir = vvv_config['sites'][site]['local_dir']
-  site_host_paths = Dir[File.join(site_local_dir, '*', 'vvv-hosts'), File.join(site_local_dir, '*', '*', 'vvv-hosts'), File.join(site_local_dir, '*', '*', '*', 'vvv-hosts'), File.join(site_local_dir, '*', '*', '*', '*', 'vvv-hosts')]
+  site_host_paths = Dir.glob(Array.new(4) {|i| vvv_config['sites'][site]['local_dir'] + '/*'*(i+1) + '/vvv-hosts'})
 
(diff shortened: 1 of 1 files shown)


LoreleiAurora:
This could break site provisioners that rely on scanning for vvv-hosts multiple levels deep.
For example [WordPress/meta-environment](https://github.com/WordPress/meta-environment/) keeps its host files 3 levels deep.

Instead i think we should just restrict this to 4 levels deep bringing the hosts file on the hosts machine in sync with the hosts file on the VM.

davidpagepsycle:
@LoreleiAurora I've rewritten the line to be more generic but restricted to 4 levels deep. Have checked locally using WordPress/meta-environment and it still picks up the vvv-hosts files but is still much faster in provisioning, vagrant ssh etc.

LoreleiAurora:
Looking at this again we need to come up with a different way to do the scan. the scan 4 levels deep in provision-site.sh is relative to a site dir where as this is relative to the www root.

tomjn:
This makes sense, I don't think more than 4 levels deep is necessary, as long as the docs are updated

davidpagepsycle:
Looking into it a bit more, the vvv-hosts code within Vagrantfile only looks within the www root hierarchy whereas the docs state that with custom paths the site local_dir path could be anywhere. With a custom local_dir the vvv-hosts file would only be used within provision-site.sh so any custom hosts would have to be duplicated in vvv-custom.yml otherwise nothing would be added to the hosts file on the hosts machine. At least that's how i'm reading it.

As the docs state that using vvv-custom.yml is the recommendation for setting hosts and a vvv-hosts file is only for backwards compatibility would a better solution be to see if hosts have been set in vvv-custom.yml and if not only then to look within that sites' local_dir to find a vvv-hosts file?

LoreleiAurora:
Personally i would go for the approach of
1. Fix hosts files to scan site dirs
1. Add a `skip_vvv_hosts: true` flag so the scan can be turned off globally and per site

This approach would prevent breaking bc.

@jeremyfelt @tomjn Thoughts?

davidpagepsycle:
How about this? Covers point 1. Priority is hosts defined in vvv-custom.yml then on each site checks within the local_dir 4 directories deep for any vvv-hosts files to add those hosts.
Looking to add point 2 next.

Vagrantfile
```diff
@@ -71,6 +64,14 @@ vvv_config['sites'].each do |site, args|
 
   vvv_config['sites'][site] = defaults.merge(args)
 
+  site_local_dir = vvv_config['sites'][site]['local_dir']
+  site_host_paths = Dir[File.join(site_local_dir, '*', 'vvv-hosts'), File.join(site_local_dir, '*', '*', 'vvv-hosts'), File.join(site_local_dir, '*', '*', '*', 'vvv-hosts'), File.join(site_local_dir, '*', '*', '*', '*', 'vvv-hosts')]
```
LoreleiAurora: You can shorten these 2 lines to
`site_host_paths = Dir.glob(Array.new(4) {|i| vvv_config['sites'][site]['local_dir'] + '/*'*(i+1) + '/vvv-hosts'})`

LoreleiAurora:
I have left a comment inline. Once this change is made this can be merged and a seperate ticket opened for `skip_vvv_hosts: true`

davidpagepsycle:
Have made that change, getting used to using Ruby so didn't realise that that was possible.

PR #1193, opened 2017-05-11
fixing typos with provision commands in migration guide

rtpHarry:


jeremyfelt:
Good catch, thanks @rtpHarry!

PR #1197, opened 2017-05-16
Add additional properties to virtualbox

inetbiz:
- No Sound
- Display Memory
- Clipboard and drag&drop
- Other hardware glips


LoreleiAurora:
Why should these be added?
Generally none of them are needed in a headless dev environment.

inetbiz:
I removed some of the options I committed previously.

Issue #1199, opened 2017-05-18
phpcs not available on PATH


victorlopezg:
## Expected Behavior
After running the initial provisioning `phpcs` should be available on PATH. 

## Current Behavior
Executing `which phpcs` returns an empty response.

## Possible Solution
The provsion script creates links to `/srv/www/phpcs/scripts/phpcs` and `/srv/www/phpcs/scripts/phpcs`, however that directory does not contain the binary files.

Instead, they're in `/srv/www/phpcs/bin/`

<img width="835" alt="screen shot 2017-05-17 at 7 54 00 pm" src="https://cloud.githubusercontent.com/assets/2754872/26183212/9e147614-3b3a-11e7-9294-db112f453848.png">

The corresponding command for linking on provision.sh (`php_codesniff()`) should be updated.

## Steps to Reproduce (for bugs)

1. Provision a new machine (`vagrant up`)
2. `vagrant ssh`
3. `which phpcs`

thebrandonallen:
The problem here is that we're still provisioning with the `master` branch of PHP_CodeSniffer, which contains version 3.0. However, we should actually be on the `2.9` branch, because WordPress Coding Standards sniffs aren't, yet, compatible with PHP_CodeSniffer 3.0 (see https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/718).

Changing the paths will fix `phpcs` not being available, but it will break the WP sniffs.

This may be a good time to make a move on #1169.

TangRufus:
Any temporary workaround for now?

victorlopezg:
I just tested the fix for #1169 and it also fixes this. Can be closed in my opinion.

@TangRufus as for a workout, manually applying the patch for #1169 and doing a `vagrant reload --provision` should work.

TangRufus:
thanks @victorlopezg 
Using the `develop` branch works

tomjn:
I've updated to the latest `develop` branch and my own `phpcs` is located at `/srv/www/phpcs/scripts/phpcs` not `/srv/www/phpcs/bin/phpcs` as it is in new machines

philipjohn:
Pulling the latest from `develop` resolves this for me too.

bjorn2404:
Looks like it's been a few months since the last comment but pulling from `develop` is not working on my end. `which phpcs` is still returning nothing.

tomjn:
@bjorn2404 does reprovisioning fix that?

[...]

PR #1200, opened 2017-05-18
Fix wrong path for phpcs and phpcbf executables.

victorlopezg:
Fix for issue #1199

### DIFF
--- a/provision/provision.sh
+++ b/provision/provision.sh
@@ -610,8 +610,8 @@ php_codesniff() {
   fi
 
   # Link `phpcbf` and `phpcs` to the `/usr/local/bin` directory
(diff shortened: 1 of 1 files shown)


tomjn:
This change would break my local copy, @LoreleiAurora it would appear that the change to PHPCS installation is not applied consistently, and now we have 2 populations of PHPCS installs. 1 with PHPCS at `/srv/www/phpcs/bin/phpcs` and one at `/srv/www/phpcs/scripts/phpcs`. The change to composer didn't migrate existing usrs

tomjn:
Something like this is might fix the issue for new users without breaking it for existing users:

```
{
    "config": {
        "bin-dir": "scripts"
    }
}
```

tomjn:
So my local copy actually requires this change, full approval to merge

tomjn:
Fixes #1199

Issue #954, opened 2016-08-11
Upgrade ubuntu to 16.04 LTS


halvardos:
Hi,

Thank you for creating VVV, it has saved me tremendous amounts of work!

Is there a reason why upgrading the base image for ubuntu to 16.04 has not been discussed?
Would it make the upgrading of other packages easier to maintain to bump the core system version?

LoreleiAurora:
I would love this to happen but when trying to do this `vagrant up` fails at creating network interfaces as sudo fails.

Sudo fails because of [this issue](https://bugs.launchpad.net/cloud-images/+bug/1561250) filed in launchpad. As the hostname is not set sudo gives the wrong exit status so Vagrant assumes the command failed even when it succeeded.

Therefore until this issue is fixed the official Ubuntu Xenial Vagrant box cannot be used.

Jeger:
aaah, that makes sense.

Thank you for the quick answer.
I started using the provisioning files from this project as a base for a docker container to get in line with the rest of our organisation but still keep my development environment the same.

And they use a more recent distro as base, so thats why I'm asking :)

If anyone else is interested i could share my docker implementation, it has some limitations atm. but generally works the same.

ivankruchkoff:
Both issues have a confirmed fix in place.

LoreleiAurora:
As the issues were changed from Fix Commited to Fix released on 2016-09-06 and was included in box release `20160907` i will start working on this at the weekend.

LoreleiAurora:
There is yet another bug to get in our way 😢
[xenial64 image (20160907.1.0) has a broken (empty) /etc/resolv.conf](https://bugs.launchpad.net/ubuntu/xenial/+source/livecd-rootfs/+bug/1621393)
The fix is committed but not yet in release.

jeremyfelt:
Chatted with @LoreleiAurora a bit about this today. Let's keep an eye on the Xenial image over the next 6 months. Trusty is doing fine with the configuration that we have, and moving to a new box will definitely be a breaking change, so let's be sure it's working well with VirtualBox, VMWare Fusion, Parallels, etc.. before upgrading.

wturrell:
It *is* possible to update your existing box manually if you want/need 16.04.

Briefly, the approach I took was:
- take a snapshot for backup
- upgrade Ubuntu using `do-release-upgrade`
[...]

PR #1218, opened 2017-06-12
Current lts

DArcMattr:
Here's my work towards upgrading the Ubuntu version from 14.04 to 16.04, as requested in issue #954.


tomjn:
Closing this for now as VVV 3 will be a prebuilt box, allowing us to test and build with newer versions of Ubuntu then push them out on continuously via a custom box

PR #1237, opened 2017-07-15
Add live_url parameter for proxying images from livesite

saulirajala:
This is an enhancement to the current provisioning. With this pull request you can define in vvv-custom.yml -file, if nginx should proxy mediafiles from site that is already in public url. This is handy feature to all those, who has dev-site apart from production. You don't need to download all the necessary images from production server to make your local dev-site to work.

Example of vvv-custom.yml definations:
```
SITE_NAME:
    live_url: https://www.SITE_NAME.com
    hosts:
        - SITE_NAME.dev
```

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -421,7 +421,8 @@ Vagrant.configure("2") do |config|
         args['branch'],
         args['vm_dir'],
         args['skip_provisioning'].to_s,
-        args['nginx_upstream']
+        args['nginx_upstream'],
+        args['live_url'].to_s
       ]
   end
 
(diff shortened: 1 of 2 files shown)


tomjn:
This is a nice feature, but does it really belong in the main provisioner? IMO this should be implemented in the custom site template, which would make for a more reliable/cleaner implementation

LoreleiAurora:
Agreed with @tomjn this should go into the custom site template.

tomjn:
@saulirajala let us know if you need help or guidance doing it for the custom site template

saulirajala:
Aah, yes. I see what you mean. I'll do the pull request for the custom site template

PR #1239, opened 2017-07-22
PHPCS composer.json needs `phpcodesniffer-standard`

thebrandonallen:
WordPress Coding Standards are, currently, installed/updated to the wrong directory. Rather than being added to `/srv/www/phpcs/CodeSniffer/Standards/WordPress` as expected, the standards are being installed in `/vargrant/provision/phpcs/vendor`. I tracked the issue down to this comment, https://github.com/oomphinc/composer-installers-extender/issues/6#issuecomment-303054875.

Basically, we just need to make sure that `oomphinc/composer-installers-extender` knows that it should be working with dependencies of the type, `phpcodesniffer-standard`.

### DIFF
--- a/provision/phpcs/composer.json
+++ b/provision/phpcs/composer.json
@@ -6,7 +6,7 @@
         "wp-coding-standards/wpcs": "*"
     },
     "extra": {
-        "installer-types": ["library"],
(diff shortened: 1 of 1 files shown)


tomjn:
This fixes #1266

PR #1241, opened 2017-08-02
Vagrant plugins that VVV recommend

prajapatisagar:
Added vagrant necessary plugin

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,11 @@ VVV requires recent versions of both Vagrant and VirtualBox.
 
 [Vagrant](https://www.vagrantup.com) is a "tool for building and distributing development environments". It works with [virtualization](https://en.wikipedia.org/wiki/X86_virtualization) software such as [VirtualBox](https://www.virtualbox.org/) to provide a virtual machine sandboxed from your local environment.
 
+There are also several Vagrant plugins that VVV recommends you to install:
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -41,6 +41,11 @@ VVV requires recent versions of both Vagrant and VirtualBox.
 
 [Vagrant](https://www.vagrantup.com) is a "tool for building and distributing development environments". It works with [virtualization](https://en.wikipedia.org/wiki/X86_virtualization) software such as [VirtualBox](https://www.virtualbox.org/) to provide a virtual machine sandboxed from your local environment.
 
+There are also several Vagrant plugins that VVV recommends you to install:
```
tomjn: These are required rather than recommended, we shouldn't imply they're optional

nylen: https://varyingvagrantvagrants.org/docs/en-US/installation/ says differently (`vagrant-hostsupdater` is required, and `vagrant-triggers` is recommended).  However there doesn't seem to be anything that actually checks whether they are present.

nylen:
I wonder if we need this at all?  Isn't explaining it clearly in the installation instructions enough?  Also, as of #1248 these instructions are much easier to find.

README.md
tomjn: eh it should be changed there, I had to clean up a lot of these kinds of things from VVV1, I'm not surprised a handful got through

tomjn:
I agree with @nylen, thanks for the PR @prajapatisagar, I hope we see more :)

PR #1275, opened 2017-09-03
Remove docs directory from repository

jeremyfelt:
The management of VVV documentation has been moved to the https://github.com/Varying-Vagrant-Vagrants/varyingvagrantvagrants.org repository.


README.md
```diff
@@ -45,7 +45,7 @@ Besides VirtualBox, provider support is also included for Parallels, Hyper-V, VM
 
 Installation instructions can be found [here](https://varyingvagrantvagrants.org/docs/en-US/installation/).
 
-Full documentation can be found in the `docs/` directory of this repository and on the [varyingvagrantvagrants.org](https://varyingvagrantvagrants.org) website.
+Full documentation can be found on the [varyingvagrantvagrants.org](https://varyingvagrantvagrants.org) website.
```
nylen: > The management of VVV documentation has been moved to the https://github.com/Varying-Vagrant-Vagrants/varyingvagrantvagrants.org repository.

How would people who don't already know this find out about it?

How about an extra sentence in the readme about it?  "Submit documentation PRs [here](https://github.com/Varying-Vagrant-Vagrants/varyingvagrantvagrants.org).

A link in the site footer to "edit this page on GitHub" would also be lovely, but that is a separate change.

jeremyfelt: I think having something in the readme makes sense. I also opened https://github.com/Varying-Vagrant-Vagrants/VVV/issues/1277 the other day to address `CONTRIBUTING.md`.

LoreleiAurora:
As we do this i think we should also add a sparse checkout of the docs repo on provision so they are still mirrored on peoples machines.

julianfoad:
Yes please remove the docs directory! As a new user looking to contribute fixes to the docs, I first found that subdirectory and later learnt it is an obsolete copy of the docs. That was confusing and timewasting, only a little bit but it was.

My understanding now of the above suggestions on what *else* to do is this. The docs have already been 'moved' (actually copied), so having the old copy of them still in place is misleading and wrong: deleting the directory is a bug fix. The other suggestions above (add notes, sparse checkout) are enhancements which should already and separately be done as direct consequences of the 'move' (copying) -- regardless if the old directory is deleted or not yet -- and so should not be blocking the deletion. Don't hesitate to delete the dir first and do those things later. Makes sense?

PR #1281, opened 2017-09-17
Add variable to provision-site.sh for NGINX's server_name

coderaaron:
This adds a variable using the `get_hosts` function in `provision-site.sh` that allows the variable `vvv-hosts` to be used in vvv-nginx.conf so the user doesn't need to manually specify the hostnames in both vvv-custom.yml and vvv-nginx.conf.


coderaaron:
example vvv-nginx.conf
```
server {
  listen 80;
  listen   [::]:80;
  listen 443 ssl;

  server_name {vvv_hosts};
  
  root {vvv_path_to_site}/public_html;

  error_log {vvv_path_to_site}/log/error.log;
  access_log {vvv_path_to_site}/log/access.log;

  set $upstream {upstream};

  include /etc/nginx/nginx-wp-common.conf;

  ssl    on;
  ssl_certificate    {vvv_path_to_site}/ssl/{vvv_site_name}.crt;
  ssl_certificate_key    {vvv_path_to_site}/ssl/{vvv_site_name}.key;

  if ($scheme = http) {
    return 301 https://$server_name$request_uri;
  }
}
```

tomjn:
I like this, @LoreleiAurora thoughts?

Issue #583, opened 2015-03-01
Change TLD for test sites


pento:
As [Google has applied for control of the .dev TLD](https://gtldresult.icann.org/application-result/applicationstatus/applicationdetails/1339), there's the potential for all sorts of weirdness to occur if their application is accepted.

Per [RFC 2608, Section 2](http://tools.ietf.org/html/rfc2606#section-2), `.test`, `.example`, `.invalid` and `.localhost` are the only TLDs guaranteed to never be allocated.

westonruter:
`.localhost` seems particularly appropriate among these options. But what about [`.local`](http://en.wikipedia.org/wiki/.local)?

zamoose:
That has too many overloaded meanings in a Bonjour environment. Unless specifically going for such a la VIP QuickStart, it may be more trouble than it's worth. 

--
Doug Stewart

markkelnar:
I'm ok changing this to avoid any collision.  But it should still work and resolve to the vagrant since we're shoving the IPs in /etc/hosts.  I like .test or .localhost.

michaelbeil:
`.test` would seem to be the better of the options.

jeremyfelt:
I'm up for `.test` as well. We can probably introduce the new TLD in the configs while continuing to support old `.dev` installations.

cfoellmann:
I would vote for `.localhost` since it would make the most sense to me.
`.local` can not be used because it would clash with directory  and other services.

We need to reach some kind of consensus here and can add the tld to the configs of the sites while keeping the current domain endings for the time being.

jeremyfelt:
Ok, feedback time.

Options:
* `.localhost`
* `.test`

I excluded `.example` and `.invalid` because they don't seem right. Decisions. :)

Please leave a comment, this is super serious stuff.

raulillana:
Short is best. I would go for dot test.

richardtape:
Could it not be a choice on provision? If not, I'd prefer `.test` over the other options

jeremyfelt:
Per @rmccue, `.local` is a valid option too. IMO, better than `.localhost`. https://twitter.com/rmccue/status/701526969008062465

johnbillion:
`.local`

morganestes:
.test

I'd like to avoid potential problems with Apple and .local.

diddledan:
[...]

PR #1289, opened 2017-09-22
Alternate Dashboard domains

tomjn:
Adds support for `vvv.local` `vvv.test` and `vvv.localhost`, and changes prompts to recommend `vvv.test`

This goes some of the way towards correcting https://github.com/Varying-Vagrant-Vagrants/VVV/issues/583 though it doesn't impact the default sites

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -40,6 +40,9 @@ if ! vvv_config['hosts'].kind_of? Hash then
 end
 
 vvv_config['hosts'] += ['vvv.dev']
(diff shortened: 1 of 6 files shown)


tomjn:
PR ready for review and merge, the new primary dashboard domain will be http://vvv.test, .dev will still work, follow up PR's will be needed to change the docs and to switch the default bundled sites

tomjn:
Added a warning in the dashboard if .dev TLD is used:

screen shot 2017-09-22 at 17 35 07

PR #1300, opened 2017-10-05
replaces the old VVV ascii logo with the newer logo

tomjn:


tomjn:
screen shot 2017-10-06 at 00 49 09

It now matches the new docs site

PR #1311, opened 2017-10-21
Customize the initial working directory when connecting via SSH

westonruter:
By far my biggest annoyance with Vagrant is connecting to the VM over SSH and being dropped into `$HOME`, not the corresponding synced working directory on my host machine.  Previously I was using [`vassh`](https://github.com/xwp/vassh) to work around this, but I was not really fully satisfied because it was very slow, being that it needed to make two SSH connections for each connection (at least when invoking `vasshin` for TTY). I realized that VVV could just add first-class support for changing the initial working directory when connecting to the VM. 

What this PR adds support for a new `/srv/config/custom/initial_cwd` config file. When this is present, the `.bash_login` script will it and `cd` to whatever is inside. Additionally, it also adds support for `/srv/config/custom/initial_cwd_once` which will do the same, but the file will get removed immediately after it is read. This allows the host machine to do something like this:

```sh
pwd | sed 's:.*/vvv/www/:/srv/:' > /path/to/vvv/config/custom/initial_cwd_once; vagrant ssh
```

And then you will be dropped into the corresponding synced directory in the VM. For example:

`/Users/jsmith/vvv/www/wordpress-develop/public_html/src/wp-content` 
=> 
`/srv/www/wordpress-develop/public_html`

Eventually this could be used to create a `vagrant-ssh` script bundled with VVV which could then do this for you.

### DIFF
--- a/config/bash_login
+++ b/config/bash_login
@@ -1,0 +1,7 @@
+# Change initial working directory if one is present.
+if [ -e "/srv/config/custom/initial_cwd_once" ]; then
+    cd "$( cat /srv/config/custom/initial_cwd_once )"
+    rm /srv/config/custom/initial_cwd_once
+elif [ -e "/srv/config/custom/initial_cwd" ]; then
+    cd "$( cat /srv/config/custom/initial_cwd )"
+fi
--- a/provision/provision.sh
+++ b/provision/provision.sh
@@ -149,7 +149,16 @@ noroot() {
 }
 
 profile_setup() {
+
+  # Make sure .bash_login is loaded in TTY and non-TTY.
+  if ! grep -q "bash_login" /home/vagrant/.bashrc; then
+    echo "if [ -e ~/.bash_login ]; then . ~/.bash_login; fi" >> /home/vagrant/.bashrc~
+    cat /home/vagrant/.bashrc >> /home/vagrant/.bashrc~
+    mv /home/vagrant/.bashrc~ /home/vagrant/.bashrc
+  fi
+
   # Copy custom dotfiles and bin file for the vagrant user from local
+  cp "/srv/config/bash_login" "/home/vagrant/.bash_login"
   cp "/srv/config/bash_profile" "/home/vagrant/.bash_profile"
   cp "/srv/config/bash_aliases" "/home/vagrant/.bash_aliases"
   cp "/srv/config/vimrc" "/home/vagrant/.vimrc"
@@ -167,6 +176,8 @@ profile_setup() {
 
   rsync -rvzh --delete "/srv/config/homebin/" "/home/vagrant/bin/"
 
+  echo " * Ensured .bash_login loading is prepended to .bashrc"
+  echo " * Copied /srv/config/bash_login                        to /home/vagrant/.bash_login"
   echo " * Copied /srv/config/bash_profile                      to /home/vagrant/.bash_profile"
   echo " * Copied /srv/config/bash_aliases                      to /home/vagrant/.bash_aliases"
   echo " * Copied /srv/config/vimrc                             to /home/vagrant/.vimrc"


westonruter:
And on top of this I present [vvv-ssh](https://gist.github.com/westonruter/ac0f964716d95cdee301880f1865ba30), a next generation of `vassh`. I'd love to include this in VVV if it makes sense.

tomjn:
hmmm is there any reason this couldn't be built as a VVV Utility? The changes in `config/bash_login` appear to be what do the critical part here, but could that not just be changed to something that appends to `.bash_profile` in the home directory? At which point it'd be a simple command to append to that file in the utilities `provision.sh`

There's also various vagrant action hooks that are called https://www.vagrantup.com/docs/plugins/action-hooks.html it should be possible with the right modifications for this to run on `vagrant ssh` with no forwarding script necessary

tomjn:
If you can get it running on `vagrant ssh` I'm happy to check and approve though

westonruter:
@tomjn Thanks for the review.

> The changes in `config/bash_login` appear to be what do the critical part here, but could that not just be changed to something that appends to `.bash_profile` in the home directory? At which point it'd be a simple command to append to that file in the utilities `provision.sh`

Those are the critical changes, yes. The reason why it cannot just be appended to `.bash_profile` is that it is not read when connecting to Vagrant in non-TTY mode (e.g. `vagrant ssh -c "wp post list"`). Additionally, the logic can't be appended to `/home/vagrant/.bashrc`, a pre-existing script in the VM because at the top it short-circuits with:

```sh
# If not running interactively, don't do anything
case $- in
    *i*) ;;
      *) return;;
esac
```

So that's why I'm _prepending_ to `.bashrc`:

```sh
if [ -e ~/.bash_login ]; then . ~/.bash_login; fi
```

When a `.bashrc` is present, the `~/.bash_login` file is not read. I figured it would be better to put the logic into a separate file for maintainability rather than putting all of it into the `.bashrc`.

I'll have to look at the hooks to see if they would work.

tomjn:
Upon further examination, this should be possible inside the main Vagrant File using Vagrant Triggers to intercept `vagrant ssh`, and look up the local dir and vm dir from the config. This would replicate the functionality in `vvv-ssh` and account for the custom config tags. The `VAGRANT_CWD` env variable should come in handy

tomjn:
Specifically:

```
puts "pwd: " + ENV['PWD'].inspect
puts "vagrant dir: " + vagrant_dir
```

With those 2 variables one can figure out which sites folder you're located in by processing the `vvv_config['sites']` array. Then once a matching site is found, the `vm_dir` property can be used to retrieve the location inside the VM to `cd` to, which is by default:

```
  defaults['vm_dir'] = "/srv/www/#{site}"
  defaults['local_dir'] = File.join(vagrant_dir, 'www', site)
```

tomjn:
```
    config.trigger.before :ssh, :stdout => true do
      vvv_config['sites'].each do |site, args|
        #puts site
        # Test this site and if it's the site we want, set the appropriate folder inside the VM to a file that can be reached at /vagrant to cd into
      end
    end
```

tomjn:
```
    config.trigger.before :ssh, :stdout => true do |trigger|
      current_location = "~"
      current_site = ""
      vvv_config['sites'].each do |site, args|
        if ENV['PWD'].start_with?(args['local_dir'])
          current_location = args['vm_dir']
          current_site = site
        end
      end
      puts "current_site: " + current_site
      puts "current_location: " + current_location
    end
```
Gives the following output:
```
~/dev/www/vvv/www/hashes git::master 
❯ vagrant ssh


  __     ___     ___     __  ____    
  \ \   / \ \   / \ \   / / |___ \   
   \ \ / / \ \ / / \ \ / /    __) |  
    \ V /   \ V /   \ V /    / __/   
     \_/     \_/     \_/    |_____|  

  Varying Vagrant Vagrants 2.0.0

  Dashboard:  http://vvv.test



==> default: Running triggers before ssh...
current_site: hashes
current_location: /srv/www/hashes
```

tomjn:
See https://github.com/Varying-Vagrant-Vagrants/VVV/pull/1314, it won't handle the subdir but it gets me from `www/gutenberg` to `/srv/www/gutenberg` when doing `vagrant ssh`, making it handle the subfolders if I'm deeper inside `www/gutenberg` shouldn't be difficult, it's just string manipulation at this point. The only requirement is vagrant triggers

tomjn:
I'm going to close this in favour of #1314 now that it handles subdirectory traversal

westonruter:
Per https://github.com/Varying-Vagrant-Vagrants/VVV/pull/1314#issuecomment-338503530 I think this should be still considered as it is faster in that the doesn't have the overhead of `vagrant ssh`.

westonruter:
There is also no dependency on Vagrant triggers with this approach.

tomjn:
I'm not a fan of bundling a script for this, especially when everybody has their own vagrant ssh shortcut with a different name, be it `vudo` or in my case `vooosh`. None of those will gain any of these abilities.

However, Vagrant triggers is considered a requirement from VVV 2 onwards, but the changes in #1314 allow you to write to a file and have VVV manage the rest, so it simplifies your own solution

westonruter:
I propose the triggers from your PR be combined with the configs in mine so that we can get the initial CWD working for both TTY and non-TTY modes: https://github.com/Varying-Vagrant-Vagrants/VVV/pull/1314#issuecomment-338521102

kopepasah:
I do see benefits to both of these features and there are key things to keep in mind that developers want:

- Ability to host sites outside of the VVV directory (introduced in VVV 2.0) and
  - easily SSH into the corresponding VVV directory (__good__)
  - easily pass commands to the corresponding VVV directory without leaving current session (__better__)

@westonruter I can see your addition and script does the latter, but does it account for custom site locations defined in `vvv-custom.yml` using `vm_dir`/`local_dir`?

westonruter:
@kopepasah good point. You're right, `vvv-ssh` doesn't account for custom site locations like the `Vagrantfile` change in #1314. So that is a deficiency in `vvv-ssh`. However, the changes proposed specifically in this PR wouldn't need anything else, right?

tomjn:
Closing this as it could be done via a Utility and wrapper script

PR #1314, opened 2017-10-22
Auto change SSH working dir

tomjn:
If I have a site named `gutenberg` and I'm in that sites folder at `www/gutenberg` then this change will make sure that when I run `vagrant ssh` I start in the `/srv/www/gutenberg` folder

note: requires vagrant triggers

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,9 @@
 # No need to share individual site configs with each other
 /config/nginx-config/sites/*.conf
 
+# Ignore the cwd config for vagrant ssh
+/config/.vvv_temp_ssh_cwd
+
 # Ignore anything in the 'custom' directory in config
 /config/custom/*
 
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -473,5 +473,16 @@ Vagrant.configure("2") do |config|
     config.trigger.before :destroy, :stdout => true do
       system({'VVV_SKIP_LOGO'=> 'true'}, "vagrant ssh -c 'vagrant_destroy'")
     end
+    config.trigger.before :ssh, :stdout => true do |trigger|
+      current_location = "~"
+      current_site = ""
+      vvv_config['sites'].each do |site, args|
+        if ENV['PWD'].start_with?(args['local_dir'])
+          current_location = args['vm_dir']
+          current_site = site
+        end
+      end
+      File.open(vagrant_dir+"/config/.vvv_temp_ssh_cwd", 'w') { |file| file.write( current_location ) }
+    end
   end
 end
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -475,11 +475,10 @@ Vagrant.configure("2") do |config|
     end
     config.trigger.before :ssh, :stdout => true do |trigger|
       current_location = "~"
-      current_site = ""
       vvv_config['sites'].each do |site, args|
         if ENV['PWD'].start_with?(args['local_dir'])
-          current_location = args['vm_dir']
-          current_site = site
+          current_location = ENV['PWD'].dup
+          current_location = current_location.sub! args['local_dir'], args['vm_dir']
         end
       end
       File.open(vagrant_dir+"/config/.vvv_temp_ssh_cwd", 'w') { |file| file.write( current_location ) }
--- a/config/bash_profile
+++ b/config/bash_profile
@@ -44,3 +44,7 @@ export PATH="$PATH:/srv/www/phpcs/scripts/"
 # nvm path
 export NVM_DIR="/srv/config/nvm"
 [ -s "$NVM_DIR/nvm.sh" ] && . "$NVM_DIR/nvm.sh" # This loads nvm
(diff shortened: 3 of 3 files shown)


tomjn:
Should now handle subfolders, e.g if you're in the `wp-content` subfolder when you `vagrant ssh` you'll be in the `wp-content` folder inside the VM too

westonruter:
@tomjn one downside of this is that it still has the overhead of `vagrant ssh`. With #1311 you can do plain `ssh` with previous configuration provided by `vagrant ssh-config`, and caching that config speeds up connections drastically. So I think I still prefer my approach.

For more on why `vagrant ssh` is slow, see https://github.com/hashicorp/vagrant/issues/4214#issuecomment-49551965

tomjn:
True, but that requires user intervention to generate the SSH config, at which point they might as well setup their own script, for which there are many. This adds the feature for the default and most used option

westonruter:
@tomjn yes, and then why not support both? In order to set up their own script, it needs that `.bashrc`modification in #1311 in order to support that option. The two PRs can just use the same name for the `.vvv_temp_ssh_cwd` file. If we stick this into `/config/custom/` then we won't need to `.gitignore` it.

westonruter:
I mean, as far as I know, in order to avoid having to avoid having to do two SSH connections in order to set the CWD for a TTY connection (see [approach in `vasshin`](https://github.com/xwp/vassh/blob/b64726ae25de7a2e41a8cabbced6f6fe6befefb5/vassh.sh#L129-L146)), the extension in #1311 to `.bashrc` is needed.

westonruter:
@tomjn Also, I see that there is a problem with the PR here in that it doesn't work in non-TTY mode. If you try passing a command to run via `vagrant ssh` it doesn't work here:

```bash
$ pwd; vagrant ssh -c "pwd"
/Users/westonruter/Projects/vvv/www/wordpress-develop
/home/vagrant
Connection to 127.0.0.1 closed.
```

Whereas compare with #1311 and my forwarding script:

```bash
$ pwd; vvv-ssh pwd
/Users/westonruter/Projects/vvv/www/wordpress-develop/public_html
/srv/www/wordpress-develop/public_html
```

Notice the working directory is set because the `bash_login` is being loaded in non-TTY mode.

So yours works great for TTY mode, but it doesn't work for non-TTY mode.

So what I propose is that we include support for the config files in #1311 but then also include the triggers from yours, so then we can have the best of both.

kopepasah:
@tomjn this is a good feature, but does it allow for passing commands to the corresponding site without leaving the session?

Also, see my [comment on Weston's PR](https://github.com/Varying-Vagrant-Vagrants/VVV/pull/1311#issuecomment-338554409).

tomjn:
@kopepasah I'm unsure what you mean? The behaviour of `vagrant ssh` is unchanged, aside from the change of directory when the session is opened

tomjn:
Also, this PR handles custom `local_dir` and `vm_dir` values. Having said that, there's little that can be done outside of the vagrant folder, as `vagrant ssh` needs to know which machine to use and needs a vagrantfile

tomjn:
@kopepasah are you referring to things like `vagrant ssh -c "command goes here"`? If so, then yes, that's standard Vagrant stuff, this PR doesn't break any of that, all it changes is the folder you end up in once you ssh in.

E.g. 

```
~/dev/www/vvv/www/gutenberg git::master 
❯ vagrant ssh -c "touch banana.txt"


  __     ___     ___     __  ____    
  \ \   / \ \   / \ \   / / |___ \   
   \ \ / / \ \ / / \ \ / /    __) |  
    \ V /   \ V /   \ V /    / __/   
     \_/     \_/     \_/    |_____|  

  Varying Vagrant Vagrants 2.0.0

Connection to 127.0.0.1 closed.
~/dev/www/vvv/www/gutenberg git::master 
❯ ls
README.md   banana.txt  log         provision   public_html wp-cli.yml
```

westonruter:
@tomjn Correction on my https://github.com/Varying-Vagrant-Vagrants/VVV/pull/1314#issuecomment-338521102:

With your patch I _am_ able to get the the CWD set when I call `vagrant ssh` with a command:

```sh
$ pwd; vagrant ssh -c "pwd"
/Users/westonruter/Projects/vvv/www/wordpress-develop/public_html/src
/srv/www/wordpress-develop/public_html/src
Connection to 127.0.0.1 closed.
```

However, the same is not working when I do `ssh` with the config exported from `vagrant ssh`:

```sh
$ pwd; ssh -F ssh-config default pwd
/Users/westonruter/Projects/vvv/www/wordpress-develop/public_html/src
/home/vagrant
```

For some reason when connecting directly via `ssh` the `.bash_profile` is not read, whereas it is read when `vagrant ssh` is called.

tomjn:
This needs changing to work with the new vagrant triggers, but tbh I think this could be built as a utility, and doesn't need to be in VVV core

PR #1318, opened 2017-10-31
Branch & 2.1 in Splash

tomjn:
Right now the splash indicates 2.0.0 for both `master` and `develop`, which isn't accurate.

Instead, lets bump `develop` to 2.1.0, indicating 2.1 is in development, and add the git branch to the splash. This PR does that:

screen shot 2017-10-31 at 23 48 03

When we release 2.1, we can bump the version in `develop` to 2.2 or 2.1.1, without needing to do the faff of updating the version number on release


tomjn:
Handles situations where VVV was downloaded as an archive
screen shot 2017-11-01 at 14 13 13

PR #1319, opened 2017-11-01
Better default config

nextgenthemes:
First of all thanks to making vvv better, I suggested some config options a while back and it turned into  https://github.com/Varying-Vagrant-Vagrants/custom-site-template.git among other things.

I think this is a better default config, and I wish I would have encounter this when I digged into VVV the first time because:

1. I think the default site thing should be abandoned because its basically the same as a custom site just with hardcoded stuff right? That can bite you hard if you try to edit the config nothing will happen! And I think it should have perfect parity with the default site setup so why maintain both when one can do it?
2. `.dev` is becoming a actual domain soon. This is why just picked .local no subdomain just straight to the point.
3. Made it the folders actually represent the domains like its often seen on actual servers.
4. Put a multisite example in there, the `custom: wp_type: subdirectory` does not make it very obvious that it will create a multisite because 'multisite' is no where mentioned in the config, that could be improved if you ask me.
5. Commented out then development env because I think not everyone needs that installed by default.

This way a new user can have a way easier start with VVV without reading much of the docs at all after installing.

### DIFF
--- a/vvv-config.yml
+++ b/vvv-config.yml
@@ -14,34 +14,37 @@
 # and options are
 sites:
   # The wordpress-default configuration provides a default installation of the
-  # latest version of WordPress.
-  wordpress-default:
-    repo: https://github.com/Varying-Vagrant-Vagrants/vvv-wordpress-default.git
+  # latest version of WordPress in www/example-site/ in aviable at 
+  # http://example-site.local.
+  example-site.local:
+    repo: https://github.com/Varying-Vagrant-Vagrants/custom-site-template.git
     hosts:
-      - local.wordpress.test
+      - example-site.local
 
-  # The wordpress-develop configuration is useful for contributing to WordPress.
-  wordpress-develop:
-    repo: https://github.com/Varying-Vagrant-Vagrants/vvv-wordpress-develop.git
-    hosts:
-      - src.wordpress-develop.test
-      - build.wordpress-develop.test
-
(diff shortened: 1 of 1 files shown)


tomjn:
> 1. I think the default site thing should be abandoned because its basically the same as a custom site just with hardcoded stuff right? That can bite you hard if you try to edit the config nothing will happen! And I think it should have perfect parity with the default site setup so why maintain both when one can do it?

I agree, but changing it isn't so simple, there's an issue for it elsewhere

> 2. `.dev` is becoming a actual domain soon. This is why just picked .local no subdomain just straight to the point.

This was already handled in a previous PR where we changed everything to `.test` by default. `.local` has problems with conflicts when using zeroconf/bonjour on a network if any device has the same name

> 3. Made it the folders actually represent the domains like its often seen on actual servers.

Sites can have multiple domains, and some of the default sites do have multiple domains

> 4. Put a multisite example in there, the `custom: wp_type: subdirectory` does not make it very obvious that it will create a multisite because 'multisite' is no where mentioned in the config, that could be improved if you ask me.

There's actually an example with explanations on the dashboard at http://vvv.test at the bottom

screen shot 2017-11-01 at 23 25 32


> 5. Commented out then development env because I think not everyone needs that installed by default.

This is important for core dev, and relied on for a lot of contributor day VVV setups. You can always use the `skip_provisioning: true` option if it's not your cup of tea

PR #1329, opened 2017-11-07
Dashboard: Increase line height

ocean90:
This PR improves the code style of style.css but also increases the line height to prevent text clashing, especially for `<code>`.

Before:
before

After:
after


jeremyfelt:
Thanks @ocean90! Going to squeeze this into the 2.1 release. :)

PR #1331, opened 2017-11-08
Generate a new certificate and private key including .test domains

jeremyfelt:
This will automatically refresh the cert used for domains by default by appending a version number to the server.crt and server.key files.

Pretty sure this cert is half broken anyway, but it at least aligns with where we're currently at. 🙈


jeremyfelt:
Going to merge this in for the 2.1.x release. Feel free to scold me later. :)

PR #1338, opened 2017-11-10
adds trusted hosts in by default

tomjn:


tomjn:
I thought you'd merged it until I saw a setup step on another project

Issue #1351, opened 2017-11-14
New site instructions at vvv.test uses incorrect .localhost TLD


ChrisHardie:
## Expected Behavior

New site example should use .test TLD.

## Current Behavior

New site example uses .localhost TLD.

PR #1352, opened 2017-11-14
Fix TLD in new site example to match recommended VVV TLD

ChrisHardie:
Fixes https://github.com/Varying-Vagrant-Vagrants/VVV/issues/1351


tomjn:
Nifty :D @LoreleiAurora want to merge?

PR #1408, opened 2018-01-21
Improve the default bash prompt.

codebymikey:
As well as ensure that the root and vagrant user prompts use the same bash profiles.
Here's the old prompt:
prompt-old
compared the new one:
prompt-new

### DIFF
--- a/config/bash_profile
+++ b/config/bash_profile
@@ -44,3 +44,43 @@ export PATH="$PATH:/srv/www/phpcs/scripts/"
 # nvm path
 export NVM_DIR="/srv/config/nvm"
 [ -s "$NVM_DIR/nvm.sh" ] && . "$NVM_DIR/nvm.sh" # This loads nvm
+
+__has_parent_dir () {
+    # Utility function so we can test for things like .git/.hg without firing up a
+    # separate process
+    test -d "$1" && return 0;
+
+    current="."
+    while [ ! "$current" -ef "$current/.." ]; do
+        if [ -d "$current/$1" ]; then
+            return 0;
+        fi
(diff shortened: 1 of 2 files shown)


codebymikey:
It just dawned on me as I was making this PR that `cp "/srv/config/bash_profile" "/home/vagrant/.bash_profile"` should probably be `noroot cp "/srv/config/bash_profile" "/home/vagrant/.bash_profile"` or something.

Otherwise the owner of the `/home/vagrant.bash_profile` is the `root` user - I didn't want to commit it without testing.

 `noroot cp "/srv/config/bash_profile" "/home/vagrant/.bash_profile"` will most likely fail on existing boxes as the vagrant user won't have permissions to update it. It'll probably have to be a `chown` command instead.

tomjn:
I'm not sure about terminal changes like this, part of me thinks they should be a utility, or a personal thing to setup as people have different needs and preferences

For example, I have my own prompt I prefer with various features not normally found at https://github.com/tomjn/bashosx

codebymikey:
I think you're right. I'll look into creating a custom `vvv-utilities` type repo for my bash changes, or use a `/provision/provision-custom.sh` to define it.

Is there a reason `/srv/config/bash_profile` is only copied to the vagrant user's account?

tomjn:
rather than also the root users account? I don't know, but if you're needing to `sudo su` inside a vagrant shell, chances are you're going far beyond VVV's scope, or doing something wrong that should really be done via a utility ( e.g. installing and configuring extra software, it's all going to disappear if you need to recreate the box, not an issue if you use a Utility )

PR #1449, opened 2018-03-19
Allow to mount custom shared directories

widoz:


tomjn:
hmmm this would need to be documented, and shared folders are an iffy thing, how does this work?

LoreleiAurora:
This is not something i am willing to support, as many shared folders &
overlapping shared folders can cause performance issues.
People who know what they are doing can do this with a `Customfile` in a
site dir and `allow_customfile: true`

widoz:
@LoreleiAurora Sorry for wasting time, didin't know how to deal with the Customfile I was thinking it was another thing.

You can reject the merge, you're totally right.

PR #1472, opened 2018-04-30
Makes the check for git vs zip in the splash a little more reliable

### LINKED ISSUE #1471: Splash Screen VCS/Zip Missing in Bug Reports
We've had a lot of people, likely Windows users, who are missing the git branch or `zip` marker in the splash when they submit logs, making it necessary to ask which version of VVV they're using.

Moving the check into Ruby would likely fix this, as I suspect the problem is trying to run bash code in a Windows context

tomjn:
Also explicitly states that it's a git based install

This might fix #1471 by removing the bash conditional, which probably fails on Windows

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -14,8 +14,13 @@ if ENV['VVV_SKIP_LOGO'] then
   show_logo = false
 end
 if show_logo then
-  branch = `if [ -f #{vagrant_dir}/.git/HEAD ]; then git --git-dir="#{vagrant_dir}/.git" --work-tree="#{vagrant_dir}" rev-parse --abbrev-ref HEAD; else echo 'novcs'; fi`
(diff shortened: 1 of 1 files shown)


tomjn:
Splash screen now looks like this on MacOS:

screen shot 2018-04-30 at 14 26 05

Testing on Windows still needs to be done to verify this fixes the issue

tomjn:
To test this, checkout this branch then run `vagrant status` and screenshot the results with your OS version, and the Command prompt/shell used

tomjn:
So I fired up my old gaming PC and tested the PR out, and success! It even fixes the odd `unexpected -f` issue:

Using Git Shell from GitHub:

vvv_gitbash

Using the git bash shell from git-scm.com

vvv_mingw32gitbash

Elevated Powershell:

vvv_elevated_powershell

PR #1487, opened 2018-05-08
initial attempt to switch to the new VVV core triggers

tomjn:
WIP, it accepts the new format, but it can't find the scripts inside the VM

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -513,31 +513,35 @@ Vagrant.configure("2") do |config|
 
   # Vagrant Triggers
   #
-  # If the vagrant-triggers plugin is installed, we can run various scripts on Vagrant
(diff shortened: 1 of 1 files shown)


tomjn:
Currently debugging a problem when running `vagrant halt`, it doesn't prevent halt from working, but it does prevent the db backup script from running:

screen shot 2018-05-08 at 20 28 28

tomjn:
Further debugging shows that this works:

```
  config.trigger.before :halt do |trigger|
    trigger.name = "VVV Pre-Halt"
    trigger.run_remote = { inline: "echo 'moomins'" }
  end
```
screen shot 2018-05-08 at 20 38 21

tomjn:
On further debugging it looks to be the `mysql` statement in the `db_backup` script

tomjn:
Vagrant halt now works, all checks out!

PR #1520, opened 2018-05-23
Sets onerror triggers

tomjn:
In theory this should prevent vagrant destroy and halt from failing when a trigger error occurs

Helps make vagrant destroy more robust

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -528,6 +528,7 @@ Vagrant.configure("2") do |config|
   config.trigger.before :reload do |trigger|
     trigger.name = "VVV Pre-Reload"
     trigger.run_remote = { inline: "/vagrant/config/homebin/vagrant_halt" }
+    trigger.on_error = :continue
   end
   config.trigger.after :reload do |trigger|
     trigger.name = "VVV Post-Reload"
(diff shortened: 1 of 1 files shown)


tomjn:
This is for when `vagrant destroy` fails, e.g. running `vagrant destroy` on a VM that's already been halted

Mte90:
This is CD USB Generator https://github.com/Varying-Vagrant-Vagrants/CD-USB-Generator/

There is a way to generate the error? So we can see if it is fixed.

Mte90:
I am not sure also how to test, I am executing the CD USB generator to see if the problem disappear but looking on your screenshot you used the generator only to have a VVV version to destroy for try this patch.

Mte90:
To test:
```
git checkout  tomjn-trigger-failures
vagrant destroy```

with the vagrant machine halted and yes is like on the screen

tomjn:
I tested it locally here, I ran into am issue with the CD generator repo
because of changes in VVV and couldn't get rid of the VM the usual way. I
fixed that repo up but this fixes the general problem of not being able to
destroy or halt if an error occurs

As for the CD generator CD stands for contributer day. It provisions VVV
and downloads installers to put on USB sticks for machine setup

PR #1524, opened 2018-05-25
Use the version file in the splash screen

tomjn:
reads in the version file and removes the trailing new line in the vagrant splash


tomjn:
That was fast!

Mte90:
Well it is a simple patch this one :-D

PR #1525, opened 2018-05-25
require the hosts updater plugin in the vagrant file

tomjn:


Mte90:
I noticed this error, I didn't saw when I checked this one:
```
Vagrant.require_plugin is deprecated and has no effect any longer.
Use `vagrant plugin` commands to manage plugins. This warning will
be removed in the next version of Vagrant.
There was an error loading a Vagrantfile. The file being loaded
and the error message are shown below. This is usually caused by
a syntax error.

Path: /home/www/VVV/Vagrantfile
Line number: 21
Message: Errno::ENOENT: No such file or directory @ rb_sysopen - version
```

tomjn:
I'm inclined to revert this as a result

PR #1530, opened 2018-05-29
Revert "require the hosts updater plugin in the vagrant file"

tomjn:
Reverts Varying-Vagrant-Vagrants/VVV#1525


tomjn:
This should fix a warning given about `require_plugins` on vagrant up

Mte90:
So probably add a warning about this plugin if not found in the vagrantfile can be an alternative solution.

PR #1534, opened 2018-05-30
Switch to vvv-base/ubuntu1404 for virtualbox

LoreleiAurora:
This has multiple advantages:
- We controll the box
- Security updates from Jenkins weekly rebuilds

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -210,7 +210,7 @@ Vagrant.configure("2") do |config|
   # This box is provided by Ubuntu vagrantcloud.com and is a nicely sized (332MB)
   # box containing the Ubuntu 14.04 Trusty 64 bit release. Once this box is downloaded
   # to your host computer, it is cached for future use under the specified box name.
-  config.vm.box = "ubuntu/trusty64"
+  config.vm.box = "vvv-base/ubuntu1404"
 
   # The Parallels Provider uses a different naming scheme.
   config.vm.provider :parallels do |v, override|


tomjn:
awesome, let me test it and I'll get back to you

Mte90:
Just a question is not better to move on 16.04?

LoreleiAurora:
@Mte90 this box is a swap out replacement, an upgrade would not be. 

I would much prefer to be testing packer with something stable than untested.

Mte90:
Ok but I think that we need to start planning an upgrade because more we wait and more will be difficult to do that.

LoreleiAurora:
Im building the box for V3 with 18.04 (Bionic)

tomjn:
Once VVV 2 is using `vvv-base`, updating Ubuntu or even swapping it out entirely becomes easier in the longterm, as well as changes for V3

benlumia007:
just out of curiosity, when building a box, would it be a good idea to use .vhd rather than .vdi so other platform like hyper-v can use.. same box, different vm software. 

just thinking outside of the box.

tomjn:
I don't believe that would be possible, my understanding is that Virtualbox could use them but it would convert them in the process, so after it happened you'd end up diverging

benlumia007:
ahhhh that's interesting, never thought about that...

benlumia007:
I actually wanted to ask, is this pull request will be added as a VVV 2.2 or will this be added as part of VVV 3.0

Mte90:
If you see is marked for 2.3 milestone but as I know the plan iso for 3 for custom builds.

benlumia007:
ahh, i didn't see that marked for 2.3 milestone,

tomjn:
I don't know if @LoreleiAurora want's this in 2.3 or a 2.4 release, or if it just needed testing to verify 3.0 stuff, but I would like to do a 2.3 release soon with MailHog and more recent quality of life improvements

LoreleiAurora:
This can go in any 2.x release, there is a different box for 3 :)

tomjn:
We just need somebody to confirm this works fine on Windows

Mte90:
I don't use windows so we have to see for others.

benlumia007:
I'll do the testing shortly.

PR #1542, opened 2018-06-07
adds the ruby-dev package

tomjn:
which is needed when building some ruby package dependencies, such as those mail catcher relies on

With this, mailcatcher builds, though I can't access it via the `http://vvv.test:1080` url


tomjn:
For now this should resolve provisioner log errors until MailHog is operational

PR #1543, opened 2018-06-08
Mailhog

tomjn:
related to https://github.com/Varying-Vagrant-Vagrants/vvv-utilities/pull/17

Attempts to install golang and setup MailHog, serving it at http://vvv.test:8025

TODO: 
 - Setup auto startup on vagrant up
 - Make the dashboard detect if mailhog is present and add a MailHog button, remove the mailcatcher button
 - test test test test

I'd advise waiting until the utilities PR has been merged to test

### DIFF
--- a/config/bash_profile
+++ b/config/bash_profile
@@ -38,6 +38,9 @@ eval "$(grunt --completion=bash)"
 # add autocomplete for wp-cli
 . /srv/config/wp-cli/wp-completion.bash
 
+# Go
+export GOPATH=$HOME/gocode
+
 # PHPCS path
 export PATH="$PATH:/srv/www/phpcs/scripts/"
 
--- a/config/bash_profile
+++ b/config/bash_profile
@@ -39,6 +39,7 @@ eval "$(grunt --completion=bash)"
 . /srv/config/wp-cli/wp-completion.bash
 
 # Go
+export PATH="$PATH:/usr/local/go/bin"
 export GOPATH=$HOME/gocode
 
 # PHPCS path
--- a/config/init/mailcatcher.conf
+++ b/config/init/mailcatcher.conf
@@ -1,8 +1,0 @@
-description "Mailcatcher"
-
-start on runlevel [2345]
-stop on runlevel [!2345]
-
-respawn
-
-exec /usr/local/rvm/bin/mailcatcher --foreground --http-ip=0.0.0.0
--- a/config/init/vvv-start.conf
+++ b/config/init/vvv-start.conf
@@ -9,5 +9,4 @@ service nginx start
 service php7.2-fpm start
 service memcached start
 service mysql start
-service mailcatcher start
 end script
--- a/config/init/vvv-start.conf
+++ b/config/init/vvv-start.conf
@@ -9,4 +9,5 @@ service nginx start
 service php7.2-fpm start
 service memcached start
 service mysql start
+service mailhog start
 end script
--- a/config/php-config/mailcatcher.ini
+++ b/config/php-config/mailcatcher.ini
@@ -1,1 +1,0 @@
-sendmail_path = "/usr/local/rvm/bin/catchmail -f admin@local.dev"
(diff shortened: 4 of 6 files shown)


Mte90:
Just for test:

* to use https://github.com/Varying-Vagrant-Vagrants/vvv-utilities/pull/17/files change in your vvv-custom.yml `utility-sources` and add `branch: mailhog`
* go to vvv folder and `git pull & git checkout mailhog`
* do a `vagrant provosion`

From what I can see https://packages.ubuntu.com/trusty/golang-go for ubuntu 14 go is available.

tomjn:
That's true and it's what I used initially, but the version is older and
mailhog fails to build when using it

tomjn:
For reference I've been following this: https://www.lullabot.com/articles/installing-mailhog-for-ubuntu-1604

Though I've deliberatley left out the IPs so that it uses the default `0.0.0.0`, setting just the web interface so it had the same port as mailcatcher caused issues with websocket connections. I did get it to catch via `mhsendmail`, and I pushed up a branch to VVV utilities that adjusts PHP

I still need to make it start on startup automatically though, something that was also broken in mailcatcher, but I've not implemented the steps to do so yet, so it should run after a `vagrant up --provision` but not a `vagrant up`

tomjn:
For reference, this PR won't fully provision anyway since when it starts mailhog it blocks provisioning continuing

Mte90:
Let me know when it's time to test :-D

dingo-d:
@tomjn Is the provisioning passing for you? I get errors when I try to provision

>dpkg: dependency problems prevent configuration of php7.2-dev:
    default:  php7.2-dev depends on php7.2-cli (>= 7.2.7-1+ubuntu14.04.1+deb.sury.org+1); however:
    default:   Package php7.2-cli is not configured yet. (--configure):
    default: dpkg: error processing package php7.2-dev (--configure):error exit status 2
    default:  dependency problems - leaving unconfigured1+deb.sury.org+1) ...
    default: dpkg: dependency problems prevent configuration of php7.2-fpm:
    default:  php7.2-fpm depends on php7.2-cli; however:
    default:   Package php7.2-cli is not configured yet.
    default: dpkg: error processing package php7.2-fpm (--configure):
    default:  dependency problems - leaving unconfigured
    default: No apport report written because the error message indicates its a followup error from a previous failure.
    default: No apport report written because MaxReports is reached already
    default: Errors were encountered while processing:
    default:  install-info
    default:  php7.2-cli
    default:  php7.2-dev
    default:  php7.2-fpm
    default: E: Sub-process /usr/bin/dpkg returned an error code (1)
    default: Main packages check and install failed, halting provision

tomjn:
PHP 7.2 issues should be unrelated to this, this PR does not touch PHP, that is an unknown error

Note that with b176a02 MailHog runs for me, there's still clean up to do:

 - mailcatcher seems to prevent mailhog from running if it too is running, we need a mechanism to uninstall it
 - right now it tries to stop the mailcatcher service and remove the service file, even if they don't exist, which is sloppy
 - Mailhog is available at `vvv.test:8025` the dashboard should be updated to reflect this

tomjn:
I believe it's testing time

Mte90:
Wow I can confirm that works!

tomjn:
Can you post a screenshot showing it catching mails? It'd be handy for the 2.3 release

Additionally, any thoughts on uninstalling mailcatcher?

Mte90:
I made a screen that show in Firefox the permission to sent notifications, but I saw later that is in Italian. I hope that is not a problem.
screenshot_20180626_154405

For mailcatcher we can do that when is set mailhog is checked the mailcatcher folder and is removed everything about it.

tomjn:
Awesome, then I think all that's left is to approve and merge, we can optimise mailcatcher cleanup later

Mte90:
I think too and the feature of browser notification is very handy, you have only to remember to press connect that is not made automatically, for the rest is very cool.

Issue #1504, opened 2018-05-14
MariaDB Provision Log issue


tomjn:
While this doesn't seem to impact the final result, this shows up in the provisioner log:

```
    default: Setting up mariadb-client-core-10.1 (10.1.33+maria-1~trusty) ...
    default: Setting up mariadb-client-10.1 (10.1.33+maria-1~trusty) ...
    default: Setting up mariadb-server-core-10.1 (10.1.33+maria-1~trusty) ...
    default: Setting up mariadb-server-10.1 (10.1.33+maria-1~trusty) ...
    default:  * Stopping MariaDB database server mysqld
    default:    ...done.
    default: 2018-05-14 16:36:36 139819030898624 [Note] /usr/sbin/mysqld (mysqld 10.1.33-MariaDB-1~trusty) starting as process 28506 ...
    default: 2018-05-14 16:36:36 139819030898624 [Note] InnoDB: Using mutexes to ref count buffer pool pages
    default: 2018-05-14 16:36:36 139819030898624 [Note] InnoDB: The InnoDB memory heap is disabled
    default: 2018-05-14 16:36:36 139819030898624 [Note] InnoDB: Mutexes and rw_locks use GCC atomic builtins
    default: 2018-05-14 16:36:36 139819030898624 [Note] InnoDB: GCC builtin __atomic_thread_fence() is used for memory barrier
    default: 2018-05-14 16:36:36 139819030898624 [Note] InnoDB: Compressed tables use zlib 1.2.8
    default: 2018-05-14 16:36:36 139819030898624 [Note] InnoDB: Using Linux native AIO
    default: 2018-05-14 16:36:36 139819030898624 [Note] InnoDB: Using SSE crc32 instructions
    default: 2018-05-14 16:36:36 139819030898624 [Note] InnoDB: Initializing buffer pool, size = 256.0M
    default: 2018-05-14 16:36:36 139819030898624 [Note] InnoDB: Completed initialization of buffer pool
    default: 2018-05-14 16:36:36 139819030898624 [Note] InnoDB: Highest supported file format is Barracuda.
    default: 2018-05-14 16:36:36 139819030898624 [Note] InnoDB: 128 rollback segment(s) are active.
    default: 2018-05-14 16:36:36 139819030898624 [Note] InnoDB: Waiting for purge to start
[...]

PR #1553, opened 2018-06-16
Simplify Package Installation

tomjn:
Attempts to simplify the package installing by letting apt-get check if they need updating when it goes to install them

The hope is that this is a faster way to do things


tomjn:
closes https://github.com/Varying-Vagrant-Vagrants/VVV/issues/1504

tomjn:
This should be working now and operational, just needs somebody else to confirm it's not just me

benlumia007:
i actually tested this 30 minutes ago as well and it works without issues.

PR #1559, opened 2018-06-20
Tomjn splash info

### LINKED ISSUE #1550: Adjust Splash with system Info
At the bottom of our new issue template we have:

```
## Your Environment
<!--- Include as many relevant details about the environment you experienced the bug in -->
* VVV version: <!-- don't forget the point releases -->
* VVV Git Branch: <!-- master, develop, zip download? -->
* Vagrant version:
* VM Provider name: <!-- This is likely VirtualBox, but VVV supports Parallels and VMWare Fusion as well. -->
* VM Provider version:
* Operating System and version:
```

But wouldn't it be great if we could tell them to copy paste the colourful splash they get in `vagrant status`? This way we can be sure we get all the needed info in a common format, and save ourselves asking questions, and the user filling things out. We can include a git commit hash to know exactly which VVV they're using, and platform info. Even a screenshot would do the job

tomjn:
Closes #1550, needs testing on Windows, tested on MacOS

screen shot 2018-06-20 at 20 44 45

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -41,13 +41,56 @@ STARS
 \033[1;38;5;196m#{red}__ #{green}__ #{blue}__ __ 
 #{red}\\ V#{green}\\ V#{blue}\\ V / #{red}Varying #{green}Vagrant #{blue}Vagrants
 #{red} \\_/#{green}\\_/#{blue}\\_/  #{purple}v#{version}#{creset}-#{branch_c}#{git_or_zip}#{branch}
- 
+ HEREDOC
+
+  platform = '' + Vagrant::Util::Platform.platform + ' '
+  if Vagrant::Util::Platform.windows? then
+    if Vagrant::Util::Platform.wsl? then
+      platform = platform + 'wsl '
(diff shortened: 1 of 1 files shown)


Mte90:
I can test on linux but in any case is very cool!

tomjn:
So it works in Windows but if `VBoxManage.exe` isn't in PATH then the virtualbox version doesn't show

vvv-splash-info

tomjn:
vvv-powershell

it seems to think it's running in mingw32 for some reason

tomjn:
Similar results in Powershell ISE, although PS ISE doesn't handle the colour codes
vvv-powershell-ise

I suspect this has something to do with how mingw32 git installed itself as cmd.exe shows the same results albeit with colours

tomjn:
here's git bash and mingw64

vvv-windows-shells

I think for windows we'll need a modified splash colour scheme for better compat. As for virtualbox version, there's only so much we can do at the moment, I'm inclined to merge and circle back for an iterative approach

PR #1565, opened 2018-06-27
Abort on no network

tomjn:
- use https://google.com
 - increase timeout
 - stops provisioning on no network
 - tells the user next steps
 - warns the user not to provision on contributor day
 - puts `ifconfig` output inline for debugging purposes

screen shot 2018-06-27 at 17 00 05


tomjn:
Tested by turning wifi off and provisioning

Mte90:
I am not sure if ping google is the best option, maybe ping wordpress.org instead? so we are neutral.

benlumia007:
I'm in the process of testing this now.

tomjn:
I'm considering switching from https://google.com to `8.8.8.8`, as it's also a google property, but it's a raw IP so DNS lookups won't be a problem. In an ideal scenario it might actually be a bit faster

benlumia007:
looks good.

Issue #1567, opened 2018-06-28
Network connection not detected. Unable to reach google.com on brand new VM


oscarssanchez:
Hi, 

Could you please take a look at this problem? i have installed VVV previously with no major problems whatsoever on another PC's, however, I'm running into a lot of trouble with a brand new laptop i recently bought. 

## Expected Behavior
My internet conection is working and VVV tells me there is no network connection.

## Current Behavior

The following message appears when trying to either `vagrant up --provision` or` vagrant up`:

```
default: Network connection not detected. Unable to reach google.com...
    default:  
    default: #################################################################
    default:  
    default: Problem:
    default: Provisioning needs a network connection but none was found.
    default: VVV tried to ping google.com, and got no response.
    default:  
    default: Make sure you have a working internet connection, that you 
    default: restarted after installing VirtualBox and Vagrant, and that they
    default: aren't blocked by a firewall. If you can load https://google.com
    default: in your browser, then VVV should be able to connect.
    default:  
    default: Additionally, if you're at a contributor day event, be kind,
    default: provisioning involves downloading things, a full provision may 
    default: ruin the wifi for everybody else :(
    default:  
    default: Network ifconfig output:
    default:  
    default: eth0      Link encap:Ethernet  HWaddr 08:00:27:d0:46:cc  
    default:           inet addr:10.0.2.15  Bcast:10.0.2.255  Mask:255.255.255.0
    default:           inet6 addr: fe80::a00:27ff:fed0:46cc/64 Scope:Link
    default:           UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
    default:           RX packets:1734 errors:0 dropped:0 overruns:0 frame:0
    default:           TX packets:1444 errors:0 dropped:0 overruns:0 carrier:0
    default:           collisions:0 txqueuelen:1000 
    default:           RX bytes:205424 (205.4 KB)  TX bytes:169876 (169.8 KB)
    default: 
    default: eth1      Link encap:Ethernet  HWaddr 08:00:27:05:fa:db  
    default:           inet addr:192.168.50.4  Bcast:192.168.50.255  Mask:255.255.255.0
    default:           UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
    default:           RX packets:2 errors:0 dropped:0 overruns:0 frame:0
    default:           TX packets:9 errors:0 dropped:0 overruns:0 carrier:0
    default:           collisions:0 txqueuelen:1000 
[...]

PR #1570, opened 2018-06-29
Added git lfs support

obilodeau:
`git lfs` is very useful in web design environments where a lot of large `.psd`, `.jpg`, etc. files are thrown around. It makes git a lot faster with files it has no chance of optimizing due to their high compression.

It is not a complex or long install but the key current pgp key I used will expire in 2019 so be aware of that.

Need #1567 fixed before it can work.


tomjn:
@obilodeau This looks good, needs testing though, is there a repo with git LFS that we can use for reference?

obilodeau:
> This looks good, needs testing though, is there a repo with git LFS that we can use for reference?

I tested it but I agree that this needs to be tested by someone else.

I added a site in `vvv-custom.yml` with the following definition:
```yaml
  mysite:
    repo: https://github.com/Apress/repo-with-large-file-storage.git
    hosts:
      - git-lfs.test
```

and inspected the files manually using `vagrant ssh` and git lfs did its job.

tomjn:
Apologies for the delay in testing, this slipped my mind

tomjn:
@obilodeau I gave this a go and it checked out the repo but the large file zip is just a text file placeholder:

```
version https://git-lfs.github.com/spec/v1
oid sha256:6e1b890dafa9956b1bdff43b3fd46aaf273085a1d2041bde7efce6cf5eb0262b
size 207809687
```

It doesn't look like this was succesful, or at least not after a `vagrant halt; vagrant up --provision`

tomjn:
This works! The only caveat is that if a git repo has already been created, it won't retroactively fix LFS files

PR #1571, opened 2018-06-29
Switches network detection back to google.com

tomjn:
also expands the help to mention VPNs, props to @abrightclearweb


tomjn:
Closes #1567

tomjn:
@oscarssanchez does this PR resolve the issues at your end?

PR #1576, opened 2018-07-02
adds a slightly nicer default prompt that adds colours

tomjn:
Updated the bash profile script to provide a more pleasant default prompt:

screen shot 2018-07-02 at 19 49 19

Users can still override this by putting a `.bash_prompt` file in their homefolder


tomjn:
Note that this only adds colours, the prompt is otherwise the same

PR #1582, opened 2018-07-10
skips running a site provisioner if the site has been skipped in the config file

tomjn:
If you set a site to skip provisioning, its provisioner still runs, but exits immediatley.

Instead, this PR adds the same check to the Vagrant file, speeding things up a little bit, and reducing provisioner output

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -532,17 +532,19 @@ Vagrant.configure("2") do |config|
   end
 
   vvv_config['sites'].each do |site, args|
-    config.vm.provision "site-#{site}",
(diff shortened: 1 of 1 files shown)


Mte90:
How this parameter is used?

tomjn:
It's `skip_provisioning: true` in a site config,

e.g.

```yaml
  gutenberg:
    repo: https://github.com/Varying-Vagrant-Vagrants/custom-site-template
    skip_provisioning: true
    hosts:
      - gutenberg.test
```

PR #1584, opened 2018-07-10
fix mailhog on startup

tomjn:
phpenmod mailhog not mailcatcher, and simplify the mailcatcher service cleanup


desrosj:
`mailhog.ini` is now showing in the list of parsed `.ini` files for me!

Issue #1579, opened 2018-07-05
Errors using npm in VVV box on Windows host machine


mnelson4:
This is an issue that's general to using Windows, a virtual machine with shared folders, and `npm`, but I thought I'd report it here in case another VVV user encounters the problem.

## Expected Behavior

I would expect running `npm install` to work on projects setup for it.

## Current Behavior

After running `vagrant ssh` and cd'ing to a project setup for `npm`, I'd run `npm install` and it would fail, with output similar to this:

```
vagrant@vvv:/srv/www/ee71-develop/public_html/build/wp-content/plugins/event-espresso-core$ npm install --no-bin-links
npm WARN rollback Rolling back globals@11.7.0 failed (this is probably harmless): ETXTBSY: text file is busy, unlink '/srv/www/ee71-develop/public_html/build/wp-content/plugins/event-espresso-core/node_modules/@babel/traverse/node_modules/globals/package.json.2503924390'
npm WARN rollback Rolling back @emotion/hash@0.6.2 failed (this is probably harmless): ETXTBSY: text file is busy, unlink '/srv/www/ee71-develop/public_html/build/wp-content/plugins/event-espresso-core/node_modules/@emotion/hash/package.json.1897366673'
npm WARN rollback Rolling back @emotion/is-prop-valid@0.6.1 failed (this is probably harmless): ETXTBSY: text file is busy, unlink '/srv/www/ee71-develop/public_html/build/wp-content/plugins/event-espresso-core/node_modules/@emotion/is-prop-valid/package.json.203011432'
npm WARN rollback Rolling back @emotion/stylis@0.6.5 failed (this is probably harmless): ETXTBSY: text file is busy, unlink '/srv/www/ee71-develop/public_html/build/wp-content/plugins/event-espresso-core/node_modules/@emotion/stylis/package.json.990352348'
npm WARN rollback Rolling back @emotion/memoize@0.6.1 failed (this is probably harmless): ETXTBSY: text file is busy, unlink '/srv/www/ee71-develop/public_html/build/wp-content/plugins/event-espresso-core/node_modules/@emotion/memoize/package.json.3666918240'
npm WARN rollback Rolling back to-fast-properties@2.0.0 failed (this is probably harmless): ETXTBSY: text file is busy, unlink '/srv/www/ee71-develop/public_html/build/wp-content/plugins/event-espresso-core/node_modules/@babel/types/node_modules/to-fast-properties/package.json.1294182330'
[...]

PR #1593, opened 2018-07-23
allow symlinks

mnelson4:
These are the changes I made to my local vagrant file as described on https://github.com/Varying-Vagrant-Vagrants/VVV/issues/1579.

Basically, this just allows you to create symlinks inside shared folders. This is handy, as described on the related ticket, so users on Windows host machines can replace their node_modules folders inside shared folders with a symlink, because npm, when ran from the guest OS, doesn't work well in folders shared with a windows host machine.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -226,6 +226,7 @@ Vagrant.configure("2") do |config|
     v.customize ["modifyvm", :id, "--rtcuseutc", "on"]
     v.customize ["modifyvm", :id, "--audio", "none"]
     v.customize ["modifyvm", :id, "--paravirtprovider", "kvm"]
+    v.customize ["setextradata", :id, "VBoxInternal2/SharedFoldersEnableSymlinksCreate//srv/www", "1"]
 
     # Set the box name in VirtualBox to match the working directory.
     vvv_pwd = Dir.pwd
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -227,6 +227,7 @@ Vagrant.configure("2") do |config|
     v.customize ["modifyvm", :id, "--audio", "none"]
     v.customize ["modifyvm", :id, "--paravirtprovider", "kvm"]
     v.customize ["setextradata", :id, "VBoxInternal2/SharedFoldersEnableSymlinksCreate//srv/www", "1"]
+    v.customize ["setextradata", :id, "VBoxInternal2/SharedFoldersEnableSymlinksCreate//srv/config", "1"]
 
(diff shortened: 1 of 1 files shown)


mnelson4:
An important note though: the fix I described on https://github.com/Varying-Vagrant-Vagrants/VVV/issues/1579 isn't fixing `npm install` when used on `srv/www/wordpress-develop/public_html`. I'm not sure why

tomjn:
This may also be necessary on the `/vagrant` folder

tomjn:
Also, this will require the shell to have admin rights on windows to work

Mte90:
I cannot test it because I don't use windows, if linux is enough I can do that.

mnelson4:
@tomjn I'm pretty inexperienced at this sort of stuff. Why would this require admin privileges in the windows shell? And why do you think the entire vagrant folder needs to allow symlinks? (Out of curiosity, it's of course an easy add, just not sure what effect to expect).

@Mte90 thanks, I don't know, but I bet your testing on linux would be helpful to make sure nothing gets broken for linux host OSs

tomjn:
> I'm pretty inexperienced at this sort of stuff. Why would this require admin privileges in the windows shell? And why do you think the entire vagrant folder needs to allow symlinks? (Out of curiosity, it's of course an easy add, just not sure what effect to expect).

I read it elsewhere, the reason for it I do not know, but you should be running in a prompt with elevated priveledges anyway for hosts updater to work. Most things like this need those kinds of priveledges on Windows anyway

Mte90:
To test what I have to do? execute `npm install` on the folders of the pr?

mnelson4:
@Mte90 I think you should mostly verify that provisioning the VM still works ok (ie, from your host machine, try running a `vagrant up` on the folder of your clone of this repository). I think if there were a problem in this pull request it would appear during provisioning (eg you'd probably receive a message saying the newly added lines cause an error).
(These changes shouldn't affect `npm install` for you, when ran from inside your guest machine as these changes don't add anything for you. So I don't think testing that is necessary.)

rleeson:
@LoreleiAurora Administrators are the only group with the privilege to create symbolic links by default on Windows 7+. The privileges required to created symbolic links in Windows 10 are outlined [here](https://docs.microsoft.com/en-us/windows/security/threat-protection/security-policy-settings/create-symbolic-links), and the Group Policy object for this exists (at least) back to Windows 7.

mnelson4:
🙇‍♂️

Issue #1617, opened 2018-08-24
Unable to use vagrant halt


tannerchung:
## Expected Behavior

`vagrant halt` should shut down vagrant

## Current Behavior

```
$ vagrant halt
__ __ __ __
\ V\ V\ V / Varying Vagrant Vagrants
 \_/\_/\_/  v2.2.1-git::master

Docs:       https://varyingvagrantvagrants.org/
Contribute: https://github.com/varying-vagrant-vagrants/vvv
Dashboard:  http://vvv.test

==> default: Running triggers before halt ...
==> default: Running trigger: VVV Pre-Halt...
    default: Running: inline script
    default: /tmp/vagrant-shell: line 1: /vagrant/config/homebin/vagrant_halt: No such file or directory
==> default: Trigger run failed
The SSH command responded with a non-zero exit status. Vagrant
assumes that this means the command failed. The output for this command
should be in the log above. Please read the output to determine what
went wrong.
```
## Steps to Reproduce (for bugs)

1. `vagrant halt`

## Context

Trying to halt the server so I can delete it and restart it.

## Your Environment

* VVV version: git
* VVV Git Branch: master
* Vagrant version: 2.1.2
* VM Provider name: Hyper-V
* VM Provider version:
* Operating System and version: Windows 10

tomjn:
Did you try it with the develop branch? Always try it with the develop
branch, there's a very high chance this issue has already been fixed

tannerchung:
Okay I'll try it

tannerchung:
@tomjn I'm getting this now :\

```
$ vagrant up
__ __ __ __
\ V\ V\ V / Varying Vagrant Vagrants
 \_/\_/\_/  v2.2.2-develop-git::develop

Platform:   mingw32 HyperV-Admin HasWinAdminPriv
Vagrant:    2.1.2
VirtualBox:

Docs:       https://varyingvagrantvagrants.org/
Contribute: https://github.com/varying-vagrant-vagrants/vvv
Dashboard:  http://vvv.test

Bringing machine 'default' up with 'hyperv' provider...
==> default: Verifying Hyper-V is enabled...
==> default: Verifying Hyper-V is accessible...
==> default: Machine already provisioned. Run `vagrant provision` or use the `--provision`
==> default: flag to force provisioning. Provisioners marked to run always will still run.
==> default: Running provisioner: shell...
An error occurred in the underlying SSH library that Vagrant uses.
[...]

Issue #1616, opened 2018-08-24
vagrant up fails for Windows 10 Hyper-V


tannerchung:
## Expected Behavior
`vagrant up` should fully provision with vvv

## Current Behavior
I did a `vagrant up` and on the first run with Windows 10 and Hyper-V and it exited. In the logs a lot of packages were unable to be installed

## Steps to Reproduce (for bugs)
1. Install vagrant
2. Download vvv from git
3. vagrant up
4.

```
$ vagrant provision
__ __ __ __
\ V\ V\ V / Varying Vagrant Vagrants
 \_/\_/\_/  v2.2.1-git::master

Docs:       https://varyingvagrantvagrants.org/
Contribute: https://github.com/varying-vagrant-vagrants/vvv
Dashboard:  http://vvv.test

==> default: Running provisioner: fix-no-tty (shell)...
    default: Running: inline script
==> default: Running provisioner: default (shell)...
    default: Running: C:/Users/tanner/AppData/Local/Temp/vagrant-shell20180823-17024-10lrjvn.sh
    default: Network connection detected...
    default: Bash profile setup and directories.
    default: cp:
    default: cannot stat ‘/srv/config/bash_profile’
    default: : No such file or directory
    default: cp:
    default: cannot stat ‘/srv/config/bash_aliases’
    default: : No such file or directory
    default: cp:
    default: cannot stat ‘/srv/config/vimrc’
    default: : No such file or directory
    default: cp:
    default: cannot stat ‘/srv/config/subversion-servers’
    default: : No such file or directory
    default: cp:
    default: cannot stat ‘/srv/config/subversion-config’
    default: : No such file or directory
    default:  * Copied /srv/config/bash_profile                      to /home/vagrant/.bash_profile
    default:  * Copied /srv/config/bash_aliases                      to /home/vagrant/.bash_aliases
    default:  * Copied /srv/config/vimrc                             to /home/vagrant/.vimrc
    default:  * Copied /srv/config/subversion-servers                to /home/vagrant/.subversion/servers
    default:  * Copied /srv/config/subversion-config                 to /home/vagrant/.subversion/config
    default: Network connection detected...
    default:
    default: Main packages check and install.
    default: Adding ppa:git-core/ppa repository
    default:  * php7.2-fpm                                          7.2.9-1+ubuntu14.04.1+deb.sury.org+1
[...]

Issue #1615, opened 2018-08-20
Deprecate Providers


tomjn:
At the moment we have provider support that's dependent on historic legacy implementations. This leads to a poor experience when users try to use those and it breaks because nobody has tested it in a long time.

These are also paid software providers, and so imo it's unreasonable to expect a free open project to support without a point person who has that software committing to maintaining and testing, something we do not have at the moment.

So, I propose we show a deprecation warning pushing users to virtualbox and warning that they proceed at their own risk.

We should support:

 - **HyperV** - should probably be kept as those with HyperV enabled for docker use are forced to use this provider
 - **VirtualBox** - should remain the recommended provider

We should deprecate:

 - **Parallels** - this probably works at the moment
 - **vmware_fusion** and **vmware_workstation** - these providers are an unknown, #1614 suggests they are broken. There are no resources or domain knowledge available to debug or diagnose issues with this provider that I'm aware of
 
I would not remove the existing code, and accept fixes if they arrive, but I don't believe we should encourage, advertise or actively support them

Mte90:
I agree on support open source solution and for others, without someone that can test/use it is quite difficult to keep them.

benlumia007:
if we can get hyperv to work, assuming that @LoreleiAurora is working on it. I am happy to test it on Windows 10 Hyper-V.

tomjn:
@benlumia007 I thought it did work?

benlumia007:
@tomjn 
Apparently it did work, now i've been trying to figure it out. right now. i can't mount the shared folders, and  it's giving me a username and password to mount i think. 
[mout.txt](https://github.com/Varying-Vagrant-Vagrants/VVV/files/2304131/mout.txt)

tomjn:
hmmm if it won't mount shared folders then it doesn't work

tomjn:
Also:
```
Vagrant requires administrator access to create SMB shares and
may request access to complete setup of configured shares.
```

Can you run this with admin privileges?

LoreleiAurora:
Parallels will not be removed any time soon as I continue to use it :)

benlumia007:
@tomjn 
i actually did run git bash (git-scm.com) as administrator, but, it still asks me for the username and password no matter what.

benlumia007:
hmm i have parallel for Mac too, but i think its an older version tho. but i can't find that license key sigh.

benlumia007:
I may try parallel for mac soon too.

[...]

PR #1619, opened 2018-08-25
Fixes to HyperV

tomjn:
Removes the mounting options, fixes the ubuntu user unknown problem, adds a performance parameter to improve provisioning speeds, comments out the SMB shares part, switches to the bento projects ubuntu 14 box, and fixes a mounting issue related to inconsistent use of trailing slashes in folder names


tomjn:
Related to #1617 #1616 and #1615

tomjn:
This seems to fix some of the problems encountered but a provision under hyperv here throws up read and write permission problems with the mounted folders

tomjn:
@LoreleiAurora I could do with a second opinion on how to wrangle the permissions, as well as valid mount options

benlumia007:
@tomjn and @LoreleiAurora 
Does this need more testing. I apparently ditched Windows and went on using Ubuntu and MacBook, but I do have a Surface Pro 4 as of right now to test hyper-v

tomjn:
@benlumia007 so right now this doesn't provision because the shared folders get mounted but the packages can't write to it, so package installation fails on the first provision. It's further than `master` or `develop` get, and it's fixable, but I'm unsure which user or group to mount it as ( or why most non virtualbox ubuntu boxes lack an `ubuntu` user to begin with )

PR #1625, opened 2018-08-30
Reduce nvm and Go footprint

Aankhen:
Currently, Go is installed twice: through `apt` (where we get v1.2.1, which is too old for MailHog), and then through a tarball (where we get v1.10.3). This PR removes the `golang-go` package from the list and updates the tarball to v1.11. I could not find a way to avoid getting the source and tests, unfortunately—when I tried excluding them in the `tar` command, compiling MailHog failed.

In addition, nvm is currently installed by cloning the repository. This is the normal method, but it pulls in a host of unnecessary files, including test files whose names end in dots, which are a nuisance to deal with on Windows. This PR switches the `script` install method so that only the three files necessary are downloaded.

### DIFF
--- a/provision/provision.sh
+++ b/provision/provision.sh
@@ -331,23 +331,30 @@ package_install() {
   return 0
 }
 
+# taken from <https://gist.github.com/lukechilds/a83e1d7127b78fef38c2914c4ececc3c>
+latest_github_release() {
+    local LATEST_RELEASE=$(curl --silent "https://api.github.com/repos/$1/releases/latest") # Get latest release from GitHub api
+    local GITHUB_RELEASE_REGEXP="\"tag_name\": \"([^\"]+)\""
+
+    if [[ $LATEST_RELEASE =~ $GITHUB_RELEASE_REGEXP ]]; then
+        echo "${BASH_REMATCH[1]}"
+        return 0
+    fi
+
+    return 1
+}
+
 tools_install() {
   # Disable xdebug before any composer provisioning.
   sh /vagrant/config/homebin/xdebug_off
 
+  local LATEST_NVM=$(latest_github_release "creationix/nvm")
+
   # nvm
-  if [[ ! -d "/srv/config/nvm" ]]; then
-    echo -e "\nDownloading nvm, see https://github.com/creationix/nvm"
-    git clone "https://github.com/creationix/nvm.git" "/srv/config/nvm"
-    cd /srv/config/nvm
-    git checkout `git describe --abbrev=0 --tags --match "v[0-9]*" origin`
-  else
-    echo -e "\nUpdating nvm..."
-    cd /srv/config/nvm
-    git fetch origin
-    git checkout `git describe --abbrev=0 --tags --match "v[0-9]*" origin` -q
-  fi
-  # Activate nvm
+  mkdir -p "/srv/config/nvm" &&
+      curl -so- https://raw.githubusercontent.com/creationix/nvm/$LATEST_NVM/install.sh |
+          METHOD=script NVM_DIR=/srv/config/nvm bash
+
   source /srv/config/nvm/nvm.sh
 
   # npm
--- a/provision/provision.sh
+++ b/provision/provision.sh
@@ -93,10 +93,6 @@ apt_package_install_list=(
   # nodejs for use by grunt
   g++
   nodejs
-
-  # MailHog requirement
-  golang-go
-
 )
 
 ### FUNCTIONS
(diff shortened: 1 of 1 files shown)


Mte90:
I like the method to check the latest version from github. 
I think that we can use also for other stuff, what do you think @tomjn?
I have to test it but seems good for me :-)

Aankhen:
I wish I could take credit for that! I wanted to convert it to use a conditional test instead of piping the text through `grep` and `sed`, but my bash-fu is too weak. If you think it would be useful generally, I’ll give that another shot.

Aankhen:
I got the bash regex working.

provision/provision.sh
```diff
@@ -526,10 +529,8 @@ phpfpm_setup() {
 
 go_setup() {
   if [[ ! -e /usr/local/bin/mailhog ]]; then
-    echo " * Installing GoLang 1.10.3"
-    curl -sO https://dl.google.com/go/go1.10.3.linux-amd64.tar.gz
-    tar -xvf go1.10.3.linux-amd64.tar.gz
-    rm go1.10.3.linux-amd64.tar.gz
+    echo " * Installing GoLang 1.11"
+    curl -so- https://dl.google.com/go/go1.11.linux-amd64.tar.gz | tar zxvf -
```
Mte90: wondering if we can do the same like for nvm also for go. In that way use github to get always the latest version so we don't have to bump that.

Aankhen: I think it should be possible, but there’s one thing about Vagrant I haven’t quite got my head around yet: can the architecture vary? If not, I can always download `go${LATEST_RELEASE}.linux-amd64.tar.gz`.

Mte90: the architecture is defined on vagrantfile and I don't think that will change :-)

Aankhen: Okay, great. Then I’ll update this so it automatically downloads the latest version.

Aankhen: Does `curl --silent "https://api.github.com/repos/golang/go/releases/latest"` give you a 404? I can get the contributors, but not the releases, it seems.

tomjn: As long as it matches the architecture of the VM OS it should be fine, but I'm unaware of any non-`amd64` scenarios, and it would require VVV running on a non-intel/AMD processor

tomjn:
I like this change, it needs testing but once that's done and the commented suggestions are resolved it should be good to merge for `2.3`

provision/provision.sh
Aankhen: Gotcha, thanks.

According to GitHub support, the reason why the release list won’t work for Go is that there are no official releases, just tags. I can’t get the latest tag either, though, so I’m waiting to hear from them.

Mte90: ok so I will waiting before an official test of this pr for this change :-)

Aankhen:
So the long and short of it is that `golang/go` has no ‘official’ releases and the tags list both is paginated and starts with a bunch of weekly tags from around 2010–2012. Getting the latest official version would require *at least* two HTTP requests to see where the weekly tags stopped, unless we hardcoded it, and that’s if we had some sort of logic to ignore the weekly tags.

This is way beyond my skills, so what I’ve tried instead is getting the latest version from [the downloads page](https://golang.org/dl/) using regexes. This is brittle—it parses HTML with regexes, and it assumes the first match is the most recent version—but I think it’s robust enough for our needs.

Thoughts?

tomjn:
I'm happy for now to stick to a known good version

tomjn:
@Aankhen I would undo https://github.com/Varying-Vagrant-Vagrants/VVV/pull/1625/commits/f7f96763793ff51da1c489d02cd16a093224c8f0

tomjn:
So this PR looks to be bugged at the moment, if you `sudo apt-get remove golang-go` or start a new VM, we get the following:

```
vagrant@vvv:~$ which go
/usr/local/go/bin/go
```

But trying to use `go` fails:

```
vagrant@vvv:~$ go version
-bash: /usr/bin/go: No such file or directory
```

Also, removing `golang-go` shouldn't be a manual step

tomjn:
Despite this mailhog still runs somehow, and a `vagrant halt;vagrant up --provision` sorts the problem

Aankhen:
> So this PR looks to be bugged at the moment, if you sudo apt-get remove golang-go or start a new VM, we get the following:

Huh, that definitely should not be happening! `golang-go` should not be installed at all; it should be using the one from the tarball. Do you still see this on on a fresh VM?

tomjn:
I removed it myself, existing installs will have that package installed
already from previous versions of VVV

tomjn:
Keep in mind that we can't just tell people to use a fresh VM as that
involves destroying database data then restoring it from backups, a lot of
people will just choose not to update

Aankhen:
Right, I understand the case where there’s an existing installation; as you said, though, this shouldn’t affect those installations unless they manually uninstall it. Thank you for massaging the PR into a usable form. :)

tomjn:
Thanks for spotting the duplication :) I think removal of unused packages would be nice, but considering the extension to provisions, and unforeseen consequences like the one I mentioned, it might be more trouble than it's worth prior to VVV 3 which makes it all a moot point

tomjn:
On this org, it's mostly just taking the provisioning of packages etc and doing it in CI to generate a prebuilt box

Aankhen:
Ah, sounds very cool.

PR #1630, opened 2018-09-02
Contributor day mode WIP

tomjn:
added a general section to the vvv config and used it to add a parameter that disables provisioning if set to anything other than all

```yaml
general:
  provision: none
```

or

```yaml
general:
  provision: all
```

Unsure if this is the best way to do the parameters, or if they should be separate, but I envisage the `general` section also being used to control if DB backups happen

Intending to add a `contributor_day` flag too that will change the defaults to `none`

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -116,6 +116,14 @@ if ! vvv_config['sites'].kind_of? Hash then
   vvv_config['sites'] = Hash.new
 end
 
+if ! vvv_config['general'].kind_of? Hash then
+  vvv_config['general'] = Hash.new
+end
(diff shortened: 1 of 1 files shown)


Mte90:
I am wondering if a contributor day require more features but anyway we can start from disabling the provision.

tomjn:
I'm thinking we need a dedicated contributor mode option, so we can change dashboard messaging etc

tomjn:
I no longer like this approach, I think this is better done as part of moving the vagrant file towards being modular and breaking it out into separate files

PR #1635, opened 2018-09-20
Site Provisioner Improvements

tomjn:
if no nginx config is found for a site, mention it in the provisioner log, also reduce the search depth from 4 to 3


tomjn:
@Mte90 all done

PR #1636, opened 2018-09-23
changes the owner of the log folder from ubuntu to vagrant

tomjn:
This should improve hyperV and parallels compatibility, as well as compat with Ubuntu boxes that don't have an ubuntu user


tomjn:
I've tested by reprovisioning an existing VM after clearing out the log folder, and destroying and recreating it, each time the log folder was writable

PR #1655, opened 2018-10-19
Fix #1643

### LINKED ISSUE #1643: Nginx config templating
There's a few issues and PR's such as #1632 and https://github.com/Varying-Vagrant-Vagrants/custom-site-template/pull/5, that add things to the nginx config via sed commands, it would make more sense to try and do a basic templating engine, so that the provisioner file only wrangles more complex things

Suggestions?

Mte90:
## Summary:
#1643

### DIFF
--- a/config/nginx-config/nginx.conf
+++ b/config/nginx-config/nginx.conf
@@ -122,5 +122,8 @@ http {
     # This directory is mapped to config/nginx-config/sites in the varying vagrant
     # vagrants repository. Additional conf files for sites should be put there.
     include /etc/nginx/custom-sites/*.conf;
+    
+    # This directory is used for custom nginx files to use for services downloadable by VVV.
+    include /etc/nginx/custom-utilities/*.conf;
 
 }
--- a/provision/provision.sh
+++ b/provision/provision.sh
@@ -497,6 +497,10 @@ nginx_setup() {
     mkdir "/etc/nginx/custom-sites/"
   fi
   rsync -rvzh --delete "/srv/config/nginx-config/sites/" "/etc/nginx/custom-sites/"
+  
+  if [[ ! -d "/etc/nginx/custom-utilities" ]]; then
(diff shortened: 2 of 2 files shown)


Mte90:
Right now this is the first part to be used by utilities so they not conflicts with sites.
https://github.com/Varying-Vagrant-Vagrants/custom-site-template/pull/5/ now I am working to add a support to inject settings for nginx inside a file.

Mte90:
Ok seems that nginx works, with this utilities but also custom sites can inject rules globally to all the websites.

config/nginx-config/nginx-wp-common.conf
```diff
@@ -79,3 +79,5 @@ location ~ \.php$ {
     # And get to serving the file!
     fastcgi_index  index.php;
 }
+
+include /etc/nginx/custom-rules/*.conf;
```
tomjn: My understanding is that this is included by all site nginx configs. I don't think that this is needed as a site template can already do this via `vvv-nginx.conf`

tomjn:
I don't see how this fixes that issue. It provides a mechanism for utilities to add rules yes, but it doesn't simplify site templates modifying the nginx config, and I'm unsure what `/etc/nginx/custom-rules/` is for, especially since every rule would be in effect on every WP site

tomjn:
I'd suggest streamlining it to the utilities rules change, then using the tideways PR as a means of testing it

config/nginx-config/nginx-wp-common.conf
Mte90: I implemented this one for https://github.com/Varying-Vagrant-Vagrants/custom-site-template/pull/5
But with that we cannot load on specific site, but as you said for this case is not the best solution.
I was thinking if someone want a service that can inject their own rules on a website, with this can be possible. For this idea I don't have an example now but probably is better to remove it.

Mte90:
Changes done, now is only for custom-utilities

PR #1656, opened 2018-10-22
add /g flag for {vvv_site_name} var replacement

runofthemill:
## Summary:

It is common to have more than one server name defined via a `server_name` directive; currently, trying to do this in `vvv-nginx.conf` will cause provisioning to fail, as the `/g` flag is not used, so only the first `{vvv_site_name}` gets replaced on a line.

Since the `vvv_provision_site_nginx()` runs after `vvv-init.sh`, it is not possible to fix this in `vvv-init.sh`, as the changes get overridden.

## Checks

### DIFF
--- a/provision/provision-site.sh
+++ b/provision/provision-site.sh
@@ -61,7 +61,7 @@ function vvv_provision_site_nginx() {
   DIR="$(dirname "$SITE_NGINX_FILE")"
   sed "s#{vvv_path_to_folder}#$DIR#" "$SITE_NGINX_FILE" > "/etc/nginx/custom-sites/${DEST_NGINX_FILE}"
   sed -i "s#{vvv_path_to_site}#$VM_DIR#" "/etc/nginx/custom-sites/${DEST_NGINX_FILE}"
-  sed -i "s#{vvv_site_name}#$SITE#" "/etc/nginx/custom-sites/${DEST_NGINX_FILE}"
+  sed -i "s#{vvv_site_name}#$SITE#g" "/etc/nginx/custom-sites/${DEST_NGINX_FILE}"
   sed -i "s#{vvv_hosts}#$VVV_HOSTS#" "/etc/nginx/custom-sites/${DEST_NGINX_FILE}"
   sed -i "s#{upstream}#$NGINX_UPSTREAM#" "/etc/nginx/custom-sites/${DEST_NGINX_FILE}"
 


tomjn:
hmmm this bug is phrased in Nginx terms not VVV terms.

Can this be translated as:

> When a user defines a site with more than 1 host in `vvv-custom.yml` and provisions with the custom site template, the additional hosts fail to provision? My understanding is that the parameter you're modifying is not a valid `server_name`, but the internal VVV name of the site, e.g. the folder that gets created, not a host. Would it not have made sense for `vvv_hosts` to be the modified value?

Additionally, I've not encountered the issue you're seeing, can I see an issue raised with instructions on how to reproduce?

runofthemill:
Sorry for any lack of clarity!

Re: your translation - it's not for multiple hosts in `vvv-custom.yml`, but using variable replacement as mentioned [here](https://varyingvagrantvagrants.org/docs/en-US/site-templates/nginx-configuration/) in a situation where there are multiple server names, as in the following nginx example from [here](https://nginx.org/en/docs/http/server_names.html):

```
server {
    listen       80;
    server_name  example.org  www.example.org;
    ...
}
```

It probably would make sense for `vvv_hosts` to be the modified value, but most of the examples in the VVV docs seem to use `vvv_site_name` (as seen [here](https://varyingvagrantvagrants.org/docs/en-US/site-templates/setup-script/) (though that's the fallback) and [here](https://varyingvagrantvagrants.org/docs/en-US/site-templates/nginx-configuration/)). I do see a function available, `get_primary_host()`, that might make sense, but `vvv_hosts` could presumably return an array, so I'm guessing that's why it isn't used in the examples.

Given the following `vvv-nginx.conf`:

```
server {
    listen 80;
    listen [::]:80;
    server_name {vvv_site_name}.test;
    return 301 https://$host$request_uri;
}

server {
    listen       443 ssl;
    server_name  .{vvv_site_name}.test ~^{vvv_site_name}.test\.\d+\.\d+\.\d+\.\d+\.xip\.io$;
    root         {vvv_path_to_site}/public_html;

    error_log    {vvv_path_to_site}/log/error.log;
    access_log   {vvv_path_to_site}/log/access.log;

    set          $upstream {upstream};

    include      /etc/nginx/nginx-wp-common.conf;
}
```

Without the `/g` flag, provisioning will throw an error upon `service nginx restart`, and the resulting config in `/etc/nginx/custom-sites/vvv-auto-example-provision-0122eb6b8c3f46d14d3136d22f748442.conf` will look like the following:

```
server {
    listen 80;
    listen [::]:80;
    server_name example.test;
    return 301 https://$host$request_uri;
}

server {
    listen       443 ssl;
    server_name  .example.test ~^{vvv_site_name}.test\.\d+\.\d+\.\d+\.\d+\.xip\.io$;
    root         /srv/www/example/public_html;

    error_log    /srv/www/example/log/error.log;
    access_log   /srv/www/example/log/access.log;

    set          $upstream php;

    include      /etc/nginx/nginx-wp-common.conf;
}
```

And since nginx requires the characters "{" and "}" to be quoted inside a regular expression, the variable will not be replaced if `~^{vvv_site_name}.test\.\d+\.\d+\.\d+\.\d+\.xip\.io$` is changed to `"~^{vvv_site_name}.test\.\d+\.\d+\.\d+\.\d+\.xip\.io$"`

Let me know if you'd still like me to open an issue with steps to reproduce!

Cheers

Mte90:
Well `g` in regex is for global to replace everywhere in the document and not only on the first occurrence.
Seems that sed replace at first occurrence in the linen so enforce it for me is not a problem.

Anyway seems strange because we have this replacement from a lot and no one complained about it :-/

tomjn:
I don't understand how to get to this situation, can you give instructions
on how to reproduce? The only way I can see to generate this scenario with
the given config would involve invalid YAML which would prevent
provisioning getting that far. The site name is the main identifier of a
site that determines its name it's folder etc, I do not see how you could
arrive at this problem

This scenario should not be possible. I'm going to have to require an
example config that would generate this kind of situation

tomjn:
To clarify further:

```yaml
my-site:
  hosts:
    - my-site.test
```

Here, `my-site` is the site name. It's not intended to be the `server_name` value, that's what `hosts` is to be used for.


In order to get the scenario you're thinking of in the config you proposed, you would need something like the following:

```yaml
"site1.test site2":
  hosts:
    - site1.test
    - site2.test
```

Which I would expect to fail to provision at some point, but eitherway it looks incredibly hacky and is not the VVV way of doing things.

The custom site template gets around this by using a `{{DOMAINS_HERE}}` value in a template file that is used to create the actual nginx config. This way the `hosts` section is used

The docs need to be simple and understandable though, and putting a `sed` command infront of someone is intimidating, so a simpler example is used where the sites name is used to generate a host name for the site.

I would suggest one of the following:

 - Switch to the custom site templates provisioners
 - experiment with using `vvv_hosts` instead of `vvv_site_name`

Otherwise I cannot see a way for a valid `vvv-custom.yml` to generate this bug without doing very hacky things. In which case, the solution is to do things properly

tomjn:
Well tests conclude that `vvv_hosts` is indeed suitable for the task and the custom site templates can be simplified

tomjn:
I've adjusted the docs to use `vvv_hosts`.

runofthemill:
Thanks for taking a look; I think there's still a slight misunderstanding, which is probably my fault. When referencing "server name" above, I was referring only to the nginx `server_name` directive.

[Here's](https://github.com/runofthemill/custom-site-template) a fork of the custom template, where the only thing I changed is `vvv-nginx.conf.tmpl`

`vvv-custom.yml`:

```
    nginxtest:
        repo: git@github.com:runofthemill/custom-site-template.git
        hosts:
        - nginxtest.test
        - www.nginxtest.test
```

Provisioning output:
```
➜ vagrant provision --provision-with site-nginxtest
__ __ __ __
\ V\ V\ V / Varying Vagrant Vagrants
 \_/\_/\_/  v2.3.0-git::master

Platform:   darwin13.4.0 shell:/usr/local/bin/zsh
Vagrant:    2.2.0
VirtualBox: 5.2.20r125813

Dashboard:  http://vvv.test

==> default: [vagrant-hostsupdater] Checking for host entries
==> default: [vagrant-hostsupdater]   found entry for: 192.168.50.4 vvv
==> default: [vagrant-hostsupdater]   found entry for: 192.168.50.4 vvv.dev
==> default: [vagrant-hostsupdater]   found entry for: 192.168.50.4 vvv.test
==> default: [vagrant-hostsupdater]   found entry for: 192.168.50.4 vvv.local
==> default: [vagrant-hostsupdater]   found entry for: 192.168.50.4 vvv.localhost
==> default: [vagrant-hostsupdater]   found entry for: 192.168.50.4 wirecutter.test
==> default: [vagrant-hostsupdater]   found entry for: 192.168.50.4 nginxtest.test
==> default: [vagrant-hostsupdater]   found entry for: 192.168.50.4 www.nginxtest.test
==> default: Running provisioner: site-nginxtest (shell)...
    default: Running: /var/folders/hv/47jr_gls14v_vm8_qs88tgg80000gn/T/vagrant-shell20181023-77263-1erjsoj.sh
    default: Downloading nginxtest, see git@github.com:runofthemill/custom-site-template.git
    default: Creating database 'nginxtest' (if it's not already there)
    default:  DB operations done.
    default: Downloading WordPress...
    default: Creating directory '/srv/www/nginxtest/public_html/'.
    default: Downloading WordPress 4.9.8 (en_US)...
    default: Using cached file '/home/vagrant/.wp-cli/cache/core/wordpress-4.9.8-en_US.tar.gz'...
    default: Success: WordPress downloaded.
    default: Configuring WordPress Stable...
    default: Installing WordPress Stable...
    default: Adding domains to the virtual machine's /etc/hosts file...
    default:  * Added nginxtest.test from /vagrant/vvv-custom.yml
    default:  * Added www.nginxtest.test from /vagrant/vvv-custom.yml
    default:  * Restarting nginx nginx
    default: nginx: [emerg] directive "server_name" is not terminated by ";" in /etc/nginx/custom-sites/vvv-auto-nginxtest-provision-e75999866f253fb22c3d3cb1e7b81693.conf:10
    default: nginx: configuration file /etc/nginx/nginx.conf test failed
went wrong.
```

Resulting nginx config:
```
vagrant@vvv:~$ cat /etc/nginx/custom-sites/vvv-auto-nginxtest-provision-e75999866f253fb22c3d3cb1e7b81693.conf
server {
    listen 80;
    listen [::]:80;
    server_name nginxtest.test www.nginxtest.test;
    return 301 https://$host$request_uri;
}

server {
    listen       443 ssl;
    server_name  .nginxtest.test www.nginxtest.test ~^{vvv_hosts}\.\d+\.\d+\.\d+\.\d+\.xip\.io$;
    root         /srv/www/nginxtest/public_html;

    error_log    /srv/www/nginxtest/log/error.log;
    access_log   /srv/www/nginxtest/log/access.log;

    set          $upstream php;

    include      /etc/nginx/nginx-wp-common.conf;
}
```

And because `vvv-init.sh` is run before `vvv_provision_site_nginx`, it is not possible to resolve this issue using a custom site provisioner.
https://github.com/Varying-Vagrant-Vagrants/VVV/blob/f4576d9d11c600969be36ccd7b81ebc5154ae15d/provision/provision-site.sh#L98-L132

tomjn:
hmm where did the `~^{vvv_hosts}\.\d+\.\d+\.\d+\.\d+\.xip\.io$;`  come from? Note that there have been changes to both code referenced in the `develop` branch

runofthemill:
That's the customization this PR is trying to support; sorry if that wasn't clear!

tomjn:
Are you sure this isn't a bug in your `vvv-nginx.conf`?

https://github.com/runofthemill/custom-site-template/blob/master/provision/vvv-nginx.conf.tmpl

Notice how in the first `server` block you used just `{vvv_hosts}` and it result in the correct output of:

```nginx
server_name nginxtest.test www.nginxtest.test;
```

But the second directive on line 10 that triggered the error is:

```nginx
server_name  .{vvv_hosts} ~^{vvv_hosts}\.\d+\.\d+\.\d+\.\d+\.xip\.io$;
```

Which generates:

```nginx
server_name  .nginxtest.test www.nginxtest.test ~^{vvv_hosts}\.\d+\.\d+\.\d+\.\d+\.xip\.io$;
```

Aside from the stray `.` at the start I would expect that the `$` needs escaping. I do see the second `{vvv_hosts}` though, but I'm guessing you want `xip.io` support so you can do subdomain multisites?

tomjn:
Would it not make more sense to do something like this?

```nginx
server_name ~^(.*)\.\d+\./d+\.\d+\.\d+\.xip\.io$;
```

My understanding is that even if it replaced the second `vvv_hosts` in that line, it would be incompatible with multiple hosts defined

tomjn:
Also `vvv_hosts` should be available inside `vvv-init.sh`, as is the YAML file via the helper functions, you could just implement the replacement yourself, and insert multiple `xip.io` entries, one for each domain

runofthemill:
Ah, those are intended; the `.` is a special wildcare name, and regex names require `^` and `$` anchors.

From https://nginx.org/en/docs/http/server_names.html :

## Wildcard names
A wildcard name may contain an asterisk only on the name’s start or end, and only on a dot border. The names `“www.*.example.org”` and `“w*.example.org”` are invalid. However, these names can be specified using regular expressions, for example, `“~^www\..+\.example\.org$”` and `“~^w.*\.example\.org$”`. An asterisk can match several name parts. The name `“*.example.org”` matches not only `www.example.org` but `www.sub.example.org` as well.

A special wildcard name in the form `“.example.org”` can be used to match both the exact name `“example.org”` and the wildcard name `“*.example.org”`.

## Regular expressions names
The regular expressions used by nginx are compatible with those used by the Perl programming language (PCRE). To use a regular expression, the server name must start with the tilde character:

```server_name  ~^www\d+\.example\.net$;```

otherwise it will be treated as an exact name, or if the expression contains an asterisk, as a wildcard name (and most likely as an invalid one). Do not forget to set “^” and “$” anchors. They are not required syntactically, but logically. Also note that domain name dots should be escaped with a backslash. A regular expression containing the characters “{” and “}” should be quoted:

```server_name  "~^(?<name>\w\d{1,3}+)\.example\.net$";```

otherwise nginx will fail to start and display the error message:

```directive "server_name" is not terminated by ";" in ...```

PR #1657, opened 2018-10-24
vvv-hosts for utilities

Mte90:
## Summary:

Add the support of scanning inside utilities to check for domains

## Checks

 - [x] I've tested this PR with Vagrant **v2.2** and VirtualBox **vXX** on **Linux**

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -159,6 +159,15 @@ vvv_config['sites'].each do |site, args|
   vvv_config['sites'][site].delete('hosts')
 end
 
+# Adding utilities folders
+utilities_parent_folder = File.join(vagrant_dir, 'www/default')
+utilities_hosts_paths = Dir.glob(Array.new(4) {|i| utilities_parent_folder + '/*'*(i+1) + '/vvv-hosts'})
+  
+vvv_config['hosts'] += utilities_hosts_paths.map do |path|
+    lines = File.readlines(path).map(&:chomp)
+    lines.grep(/\A[^#]/)
+end.flatten
+
 if ! vvv_config['utility-sources'].kind_of? Hash then
   vvv_config['utility-sources'] = Hash.new
 else


tomjn:
I don't think a scan is necessary, especially since this will add hosts for utilities that aren't being used.

Since we already know which utilities are active, we should just loop through them and assume the location of `vvv_hosts` which will be faster, the same way that we load the `provision.sh`

Mte90:
Well if we are talking of performance this doesn't change a lot, check only if the tool inside that folder have the files but this will apply also to the tools that are not enabled and this is not good indeed :-/

The problem of using the utilities is that they can have a different folder name compared to the utilities name. Right now the folder for the tideways is xhgui but the tool is another one because xhgui is the UI.
With this implementation we ignore this problem.

tomjn:
> The problem of using the utilities is that they can have a different folder name compared to the utilities name. Right now the folder for the tideways is xhgui but the tool is another one because xhgui is the UI.
> With this implementation we ignore this problem.

I don't follow? If we load `php70` in the `core` utility then that means `core/php70/provision.sh` gets executed, so why not skip the scan and then load `core/php70/vvv_hosts` if it exists?

A scan would load all utilities hosts files, which could be problematic, especially if some utilities are alternatives to each other, and implement the same thing with shared hosts. That and I can see people misusing it by nesting folders.

I'm also keen for things to be as fast as they can given that we have to consider time to provision, I'd like to be able to parallelise things in the future, and I'm also keen to deprecate `vvv_hosts` scanning in favour of the config file and having the file in a specific place, ideally `provision/vvv_hosts` and load it the same way `vvv-nginx.conf` and `vvv-init.sh` are loaded

Mte90:
I misunderstood because I have a file `provision.sh` in the folder so I added the file on the utilities folder not on the global one, so right now the implementation check inside `/var/www/default/[tool]` for the file.

tomjn:
I'm keen for everything to have a proper place, and since this is something we're adding I'd prefer VVV to check in that proper place rather than search for the file.

So for a utility, a folder should contain everything, so for `tideways` I would expect a `tideways/provision.sh` and a `tideways/vvv-hosts`. One could even adjust the VVV utility provisioner so that `vvv-nginx.conf` is loaded into the appropriate place automatically.

The important part being that there is only 1 correct place and answer, and only what's enabled gets loaded. No scans, no multiple correct places, strict simple and clear

tomjn:
Speaking of which, it would be good to adjust things so that `vvv-init.sh` can be used instead of `provision.sh` so that the mental model of site templates can be reused and things make more sense

Mte90:
So you want that VVV itself to some standard stuff like move this files of every utilities:

- vvv-hosts to /VVV/provision (with merge of an existing one)
- vvv-nginx.conf to /etc/nginx/custom-utilities (inside virtual machine)

Also:

- execute vvv-init.sh instead provision.sh for utilities (inside the virtual machine)
- remove the code of scanning and use the path of the global vvv-hosts 

Just to understand what I have to do :-)

tobiolusa:
How CAN I STIOP GETTING MESSAGE FROM THIS MAIL , PLEASE ?

dingo-d:
@tobiolusa Just go to the repo and unwatch it...

tobiolusa:
Thank you very much

tomjn:
I'm not sure what you mean by global hosts file

tomjn:
The vagrant file builds an array of hosts to add to the host machine, it's
the vagrant files job to figure out what the hosts to add are. The
provisioners inside the VM would not be modifying any hosts files

In this case there would be 3 methods

- the hosts in the site config section of VVV custom
- the VVV hosts files inside site templates
- the VVV hosts files inside utilities

I get the feeling you have something else in mind or another option, but
there isn't. These all get read in by the vagrant file and added to a ruby
array, only those things that are active get their hosts added

Mte90:
Rejected because there is a better version

PR #1659, opened 2018-10-24
Utility hosts

tomjn:
read in vvv-hosts files inside enabled utilities so that they can add hosts, WIP

## Summary:

This allows utilities to define a `vvv-hosts` file that gets included

## Checks

 - [ ] I've tested this PR with Vagrant **v2.2.0** and VirtualBox **v5.2.18** on **MacOS**

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -536,6 +536,21 @@ Vagrant.configure("2") do |config|
       end
   end
 
+  vvv_config['utilities'].each do |name, utilities|
+    if ! utilities.kind_of? Array then
+      utilities = Hash.new
+    end
+    utilities.each do |utility|
+      # load the vvv-hosts file
+      utility_hosts_file = File.join(vagrant_dir, "provision/resources/#{name}/#{utility}/vvv-hosts")
+      if File.exist? utility_hosts_file then
+        File.readlines(utility_hosts_file).map do |line|
+          vvv_config['hosts'] += [ line ]
+        end
+      end
+    end
+  end
+
   vvv_config['sites'].each do |site, args|
     if args['skip_provisioning'] === false then
       config.vm.provision "site-#{site}",


tomjn:
Related to #1657 the only part I'm unsure of is that this code runs after the vagrant hosts updater plugin aliases are set, so I'm unsure if I should move that check to the end or if I should instead redo it, @LoreleiAurora ?

tomjn:
After talking with @LoreleiAurora it seems this won't work as the utilities aren't cloned until after the hosts updater is set. Otherwise on the first provision the hosts won't work. The same is true of site templates that use `vvv-hosts`, hence why we use `vvv-custom.yml` instead.

To fix this we'll need to fork the hosts updater plugin

/cc @Mte90

Mte90:
Fork the plugin and include in VVV can speed up the installation process with a less step and be sure that all the users will have that package installed so seems not a bad idea after all.

Mte90:
I made a fork that execute the update of the hosts after the provision instead before https://github.com/Mte90/vagrant-hostsupdater/blob/19e5d086ef6d28d2b10df435db84da6af37d70a4/lib/vagrant-hostsupdater/plugin.rb#L24
I think that we need to update the vagrantfile anyway with a trigger that is executed at end of provision https://www.vagrantup.com/docs/triggers/configuration.html to update the list of hosts.
I am wondering what is the order anyway before the vagrantfile and later the plugin?
I didn't done the pull request yet because of this thoughts.

tomjn:
hmm if you can get it into hostsupdater, at some point merging it into VVV would be good as we wouldn't need to bundle the gem on contributor day and could remove a step

Mte90:
Maybe with https://stackoverflow.com/a/28801317/1902215 we can avoid to integrate but execute automatically the installation with the download of our gem.
In that way the plugin can coexist in his repo etc and avoid more code inside VVV.

Also with that code we can force to install the plugin, also all the plugin that we need and avoid more confusions with the users and remove a step from the guide.

tomjn:
We can do this via vagrant itself if we bump the minimum version number with project specific plugins.

However, I'm still mindful that on contributor days we'd need to download that plugin, and that sounds painful. One WCEU we were waiting 30 minutes to download the hostsupdater plugin even though it was only 10kb due to wifi congestion

Including it in VVV itself seems the most robust path forward

tomjn:
I'm going to close this, before this can be viable we need a mechanism for updating the hosts file in vagrant **after** provisioning has occurred, otherwise this won't work on the first provision.

if that can be resovled, happy for this to be reopened

Issue #324, opened 2014-04-11
Getting error net::ERR_INCOMPLETE_CHUNKED_ENCODING - Is this to do with NGINX?


henscu:
Using VVV (see setup below), when I try to use certain options in my Dynamik Website Builder WordPress theme below (specifically Dynamik Design and Dynamik Custom pages), I get the following error in Chrome and Safari. The symptom is that my browser just locks up. I can use the BACK button to go to the previous page and continue:

**net::ERR_INCOMPLETE_CHUNKED_ENCODING**

In Firefox more of these pages are rendered correctly but the error still occurs, just a bit later...

Is this to do with an NGINX setting that I can modify? I found a reference to NGINX Chunking at:

http://wiki.nginx.org/HttpChunkinModule

My setup is:

- VVV on Macbook Pro running Mavericks
- Mark Jaquith's WP-SKELETON https://github.com/markjaquith/WordPress-Skeleton
- Wordpress Theme: CobaltApps Dynamik Website Builder (http://cobaltapps.com/) running with Studiopress Genesis (http://www.studiopress.com/)

alexsancho:
if anone is having this issue, another setting to tweak on nginx.conf "sendfile off;" this solved the problem for me

fulicat:
@alexsancho  thank u,

cfoellmann:
we have `sendfile off` in the current version (1.2.x). Is there anything else we need to do?
If not, CLOSE

henscu:
I'm using Desktop Server for my development at the moment, so fine to close it if others say it's fixed.

I will update and test with latest VVV again to check it's resolved. If not, will get back to you.

elminson:
I fixed using this post 
https://stackoverflow.com/questions/29789268/neterr-incomplete-chunked-encoding-nginx

Adding to nginx config and restart nginx 
```
server {
    ...

    location / {
        ...
        proxy_buffers 8 1024k;  
        proxy_buffer_size 1024k;
    }
}
```

tomjn:
@elminson can you do a PR against the `develop` branch with that change? It'll need doing to https://github.com/Varying-Vagrant-Vagrants/VVV/blob/develop/config/nginx-config/nginx-wp-common.conf otherwise those changes will be removed on your next provision

elminson:
@tomjn Sure!

elminson:
@tomjn Done! you can review it!

tomjn:
[...]

PR #1704, opened 2019-01-10
Fixing Getting error net::ERR_INCOMPLETE_CHUNKED_ENCODING

elminson:
## Summary:
Fixing Getting error net::ERR_INCOMPLETE_CHUNKED_ENCODING #324

## Checks

 - [X] This PR is for the `develop` branch not the `master` branch
 - [X] This PR is complete and ready for review

### DIFF
--- a/config/nginx-config/nginx-wp-common.conf
+++ b/config/nginx-config/nginx-wp-common.conf
@@ -9,6 +9,9 @@
 # See local-nginx-example.conf-sample for a full example
 location / {
     index index.php index.html;
+    # Fixing issues/324 (Getting error net::ERR_INCOMPLETE_CHUNKED_ENCODING)
+    proxy_buffers 8 1024k;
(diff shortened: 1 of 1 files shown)


tomjn:
Thanks for the changes! I think we need to do some testing on other peoples machines to make sure it doesn't introduce any unwanted behaviour, otherwise it looks straight forward

elminson:
> Thanks for the changes! I think we need to do some testing on other peoples machines to make sure it doesn't introduce any unwanted behaviour, otherwise it looks straight forward

How we get peoples test it ? how I can test it ? cc: @tomjn

tomjn:
@elminson for future reference if you want to test, you can create a new VVV instance, but switch to the branch with your changes before provisioning. Then if all works as expected, report back or leave a PR review

Also, thanks @Mte90 :)

PR #1709, opened 2019-01-13
Site provisioner optimisations and TLS keys

tomjn:
## Summary:

This attempts to optimise site provisioning by doing several things:

 - Adds the `{vvv_tls_cert}` and `{vvv_tls_key}` so that site provisioners don't have to always copy paste the same chunk of code in their provisioners
 - Optimise the `vvv-hosts` part so that it doesn't look for them if their are hosts in the config, and so that it then checks common places first
 - Massively bumps up the number of warnings and error messages that could be shown for cases when things go wrong that weren't handled previously
 - Moves hosts file parsing into its own function
 - Simplified skip provision by returning early so that half the file can be moved over 1 indent for easier reading

## Checks

 - [x] I've tested this PR with Vagrant **v2.2** and VirtualBox **v6** on **MacOS Mojave**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,11 +6,24 @@ permalink: /docs/en-US/changelog/
 
 ## 2.5.0 ( January 2019 )
 
+2.5 Brings a major bug fix, and some performance improvements to provisioning
+
 ### Enhancements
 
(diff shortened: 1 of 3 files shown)


tomjn:
Once this is done I think a release for v2.5.1 is due

PR #1726, opened 2019-01-31
Update WP_CORE_DIR and WP_TESTS_DIR after changes in vvv-config.yml

keesiemeijer:
## Summary:
This PR updates the `WP_CORE_DIR` and `WP_TESTS_DIR` variables after the changes made in vvv-config.yml (commit 67c4945ef23f5dfd544cb67f6bdf2e7df90fdf15)

The directory for the development installation changed from `/wordpress-develop` to `/wordpress-trunk`

## Checks

 - [X] I've tested this PR with Vagrant **2.2.3** and VirtualBox **5.2.20** on **OSX**
 - [X] This PR is for the `develop` branch not the `master` branch
 - [X] This PR is complete and ready for review

### DIFF
--- a/config/bash_profile
+++ b/config/bash_profile
@@ -39,9 +39,9 @@ fi
 
 # Set the WP_TESTS_DIR path directory so that we can use phpunit inside
 # plugins almost immediately.
-export WP_TESTS_DIR=/srv/www/wordpress-develop/public_html/tests/phpunit/
+export WP_TESTS_DIR=/srv/www/wordpress-trunk/public_html/tests/phpunit/
 # Set the WP_CORE_DIR path so phpunit tests are run against WP trunk
-export WP_CORE_DIR=/srv/www/wordpress-develop/public_html/src/
+export WP_CORE_DIR=/srv/www/wordpress-trunk/public_html/src/
 
 # add autocomplete for grunt
 eval "$(grunt --completion=bash)"
--- a/config/bash_profile
+++ b/config/bash_profile
@@ -39,9 +39,18 @@ fi
 
 # Set the WP_TESTS_DIR path directory so that we can use phpunit inside
 # plugins almost immediately.
-export WP_TESTS_DIR=/srv/www/wordpress-trunk/public_html/tests/phpunit/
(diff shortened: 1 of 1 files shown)


tomjn:
hmmm I didn't realise this was a thing, older sites won't have the new trunk site, newer sites may have skipped provisioning for it, and there might be more than one trunk, this is all suboptimal.

But, since it should work for new users, lets get this merged

config/bash_profile
```diff
@@ -39,9 +39,9 @@ fi
 
 # Set the WP_TESTS_DIR path directory so that we can use phpunit inside
 # plugins almost immediately.
-export WP_TESTS_DIR=/srv/www/wordpress-develop/public_html/tests/phpunit/
+export WP_TESTS_DIR=/srv/www/wordpress-trunk/public_html/tests/phpunit/
```
tomjn: we should check if this folder exists, then fall back to `wordpress-develop` if that exists, the same with `WP_CORE_DIR`

keesiemeijer: That's a good idea. I will update the PR

keesiemeijer: The old directory is also used here.
https://github.com/Varying-Vagrant-Vagrants/VVV/blob/f95ea2082e44621e8abaf0b402f5b94ec5dff269/config/homebin/makepot#L4

Should I also update that (with a conditional for the new directory)? I'm not sure if it's called anywhere.

This is also using the old directory.
https://github.com/Varying-Vagrant-Vagrants/dashboard/blob/21858cfb44785bb0800a076e997f845986f63cd3/php/blocks/main/sites.php#L23

But that, I think, should be handled separately.

tomjn: I think it should, but lets put that in another PR. As for the dashboard repo, I think that's fine to keep as is, it's mostly adding a description for much older VVV users that didn't use `description` tags

keesiemeijer: Ok, thanks

keesiemeijer: I've updated the PR to check the directories before setting the variables.

tomjn: Thanks :)

PR #1735, opened 2019-02-18
Fix #1654, auto install vagrant plugin

### LINKED ISSUE #1654: Support Vagrant 2.2.0 for autodownload plugins
## Expected Behavior
Vagrant auto download the hosts plugin so is a step already done

## Current Behavior
Now is manually

## Possible Solution
https://www.hashicorp.com/blog/vagrant-2-2-0

Vagrant 2.2.0 support a new parameter where is possible to define the plugin and he will try to download and install automatically.

Mte90:
## Summary:

https://github.com/Varying-Vagrant-Vagrants/VVV/issues/1654

## Checks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ title: Changelog
 permalink: /docs/en-US/changelog/
 ---
 
+## 2.6.0 ( TBD )
+
+* Auto download plugin for vagrant, supported vagrant 2.2.0+
+* Autoset the locale inside the virtual machine to avoid errors in the console
+
 ## 2.5.1 ( 14th January 2019 )
 
 2.5 Brings a major bug fix, and some performance improvements to provisioning
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -271,8 +271,13 @@ Vagrant.configure("2") do |config|
   end
   
   # Auto Download Vagrant plugins, supported from Vagrant 2.2.0
-  if File.file?(File.join(vagrant_dir, 'vagrant-hostsupdater.gem')) then
-    config.vagrant.plugins = {"vagrant-hostsupdater" => {"entry_point" => File.join(vagrant_dir, 'vagrant-hostsupdater.gem')}}
+  if !Vagrant.has_plugin?("vagrant-hostsupdater") then
+      if File.file?(File.join(vagrant_dir, 'vagrant-hostsupdater.gem')) then
+        system("vagrant plugin install " + File.join(vagrant_dir, 'vagrant-hostsupdater.gem'))
+        File.delete(File.join(vagrant_dir, 'vagrant-hostsupdater.gem'))
+        puts "#{yellow}VVV has completed installing local plugins. Please run the requested command again.#{creset}"
+        exit
+      end
   end
 
   # SSH Agent Forwarding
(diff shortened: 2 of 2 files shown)


tomjn:
I don't know about adding vbguest, but have we tested how this would work with the contributor day installs? We spent a good hour trying to download a 10kb gem file to peoples machines over the wifi at WCEU last year so it needs to be 100% offline

Mte90:
If the packages are already installed they are not downloaded from vagrant.
If during contributor days people will follow the instructions this problem doesn't happen.
We can remove the vbguest that is not mandatory compared to the hosts in case.

Another option looking at the docs is https://www.vagrantup.com/docs/vagrantfile/vagrant_settings.html#entry_point that let to configure the path where is the plugin file so is not downloaded.

tomjn:
hmm does that expect a gem file or a folder with ruby files? The docs on entry point are a little sparse

Mte90:
screenshot_20190219_154012
With this new code is using a local gem file that I put in the VVV folder.
I think that is interesting for us so we can do our version of the hostsupdater plugin that execute the hook early.

tomjn:
Where did the gem file get placed?

Mte90:
I put that file in the same folder of the `VagrantFile` but it is possible to configure the path.
We can do also that the CD-USB generator put that file in the VVV folder and avoid a step during contributor days.

Mte90:
I am doing few tests and seems that there are problems when `vagrant up` is done in  a subfolder after the automatic installer.
I am doing more analysis on that.

tomjn:
@Mte90 you can use `vagrant_dir = File.expand_path(File.dirname(__FILE__))`, look at how the config file is being loaded

Mte90:
I switched to another version because there are problems.
First problem seems that using the native vagrant system the package is installed also if it is not the root folder but on starting again there is an error about not possible to load the plugin installed. Also this will install the package as local for that vagrant instance and not global.

So my solution is:

* if the package is not installed
  * if local file available install it globally
  * else install from internet

We can do another package and call it in another way, also because seems that the original is like abandoned and upload in rubygems as second step.

Mte90:
I am already using `vagrant_dir` but seems a bug in vagrant because the path printed is right and also absolute.

tomjn:
@Mte90 is that why you use a `system` call instead of `config.vagrant.plugins = {"vagrant-hostsupdater" => {"entry_point" => File.join(vagrant_dir, 'vagrant-hostsupdater.gem' )}}`?

Mte90:
I removed because there is a bug in vagrant.
The plugin is installed rightly locally only for the project but at the next run there is an error that say that cannot find the folder.
If you try you can see it, instead if installed globallythe problem doesn't happen.

tomjn:
but then don't we have to run the vagrant file a second time so the plugin is loaded?

tomjn:
And is there a vagrant issue we can refer to for the bug?

Mte90:
Using this method or the native one is always required to restart vagrant. 
The command is closed and alert for it.
About the vagrant bug I didn't had time to do more tests, if someone can check it too in a way to verify if it is me or a vagrant bug.

tomjn:
Tbh Id rather not do this then if it requires the vagrant file be run again
for hosts updater

Mte90:
It is the standard behaviour of `vagrant up` to be executed again if we use the native system.
Anyway there is a big alert to execute it again, the only other option is to do a shell script that before install the plugin and later execute vagrant but I don't think that is the best solution.

tomjn:
This makes hostupdater a requirement ( which most already consider it to be ), I've no qualms with that, the only other thing is that it also makes Vagrant 2.2 a minimum requirement

PR #1747, opened 2019-03-14
Disable the trunk develop site by default

tomjn:
Skip the develop and the meta environment using skip_provisioning instead of commenting it out

## Summary:

Provisioning for new users the first time should be as fast as possible, but enabling these templates should be easy

## Checks

### DIFF
--- a/vvv-config.yml
+++ b/vvv-config.yml
@@ -21,12 +21,14 @@ sites:
 
   # latest version of WordPress, can be used for client work and testing
   wordpress-one:
+    skip_provisioning: false
     description: "A standard WP install, useful for building plugins, testing things, etc"
     repo: https://github.com/Varying-Vagrant-Vagrants/custom-site-template.git
     hosts:
       - one.wordpress.test
 
   wordpress-two:
+    skip_provisioning: false
     description: "A standard WP install, useful for building plugins, testing things, etc"
     repo: https://github.com/Varying-Vagrant-Vagrants/custom-site-template.git
     hosts:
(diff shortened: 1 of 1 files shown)


tomjn:
Releated to #1746

afercia:
Not arguing but it would be great if this kind of changes (see also #1757) were communicated to the WordPress #docs team so that they can update the related documentation on the WordPress handbook.

I'm just back from WordCamp Torino, where we've spent some time to understand why the new VVV install new contributors were setting up didn't work as expected.

Basically, no one knew about these changes. 

Luckily, @eri-trabiccolo pointed out in the right direction. However, we've wasted some time which is undesirable during a Contributor Day where time should be dedicated to more productive activities.

I'll take care of pinging the WordPress #docs team so hopefully https://make.wordpress.org/core/handbook/tutorials/installing-a-local-server/installing-vvv/ gets updated.

For the future, I'd like to kindly suggest some better coordination 🙂Thank you!

Mte90:
We need better coordination for sure, the problem that we have right now is that we cannot open a channel in wp.org slack channel to discuss about the new features and easily keep everyone updated.
I will ping on docs team anyway :-)

tomjn:
@afercia we aren't an official .org project, for a long time we were unaware such instructions existed.

Having said that, for contributor days you shouldn't be using those instructions, or the official VVV, instead, you should be running the contributor day script to generate a pre-built VVV. That way:

 - You get a proper `instructions.html`
 - You get Trunk
 - You get a prebuilt box so no provisioning is needed
 - You get a pre-downloaded VVV folder, installers, and copy of hosts updater, so no internet is needed to get started
 - You get the WP Meta environment, so that people can make contributions to Wordcamps, .org, etc

I'd actually built a copy of these and uploaded them for WC London to use on the Friday.

I've updated the linked docs page to make it clearer where to find the CD USB drive generator, as well as to direct people to our own docs

afercia:
Thanks @tomjn @Mte90.

> instead, you should be running the contributor day script 
Yeah, this doesn't need to be communicated to me personally :) It needs to be communicated to the #docs team so all future contributor day organizers are aware (hopefully). Thanks!

PR #1754, opened 2019-04-01
Try to fix WIndows VBox version ???

### LINKED ISSUE #1753: virtualbox_version()
virtualbox_version()

In the Vagrantfile, the virtualbox_version does not display in windows, it will output unknown.  we should fix this. 

```
def virtualbox_version()
    vboxmanage = Vagrant::Util::Which.which("VBoxManage") || Vagrant::Util::Which.which("VBoxManage.exe")
    if vboxmanage != nil
        s = Vagrant::Util::Subprocess.execute(vboxmanage, '--version')
        return s.stdout.strip!
    else
        return 'unknown'
    end
end
```

tomjn:
adds a virtualbox_path function to the vagrant file to help locate virtualbox on Windows

## Summary:

This tries to close #1753 by replacing the naive check for VBoxManage with the one Vagrant itself uses so that it works better on Windows. Currently it only works in Windows if `VBoxManage` is in `PATH` which isn't true for most users

## Checks

 - [x] I've tested this PR with Vagrant **v2.2.3** and VirtualBox **v6** on **MacOS**
 

Still needs testing on Windows to confirm it does what it says, but I can confirm nothing is changed under MacOS

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ permalink: /docs/en-US/changelog/
  * Added a `vagrant_provision` and `vagrant_provision_custom` script to the homebin folder that run post-provision
  * Improved the messaging to tell the user at the end of a `vagrant up` or `vagrant provision` that it was succesful
 
+### Bug Fixes
+
+ * Improved detection of VirtualBox path to avoid `???` version numbers in the VVV splash
+
 ## 2.5.1 ( 14th January 2019 )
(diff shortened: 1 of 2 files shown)


tomjn:
Thanks :)

benlumia007:
if i find any other issues regarding this for some odd reason, I'll let you know :D but so far no issues.

PR #1756, opened 2019-04-01
Post-up/Provision splash

### LINKED ISSUE #1752: Make it clearer that provisioning has finished
When a user provisions VVV, it should say at the very end that provisioning has finished, but it isn't so clear at the moment

Ideally it shows something nice and colourful

tomjn:
## Summary:

closes #1752

Sometimes users aren't sure if things finished succesfully, and the end of a provision just looks like more output that abruptly stopped. We tried adding a short message but I don't believe it's enough, so I put much more in:

Screenshot 2019-04-01 at 11 54 35
Screenshot 2019-04-01 at 11 55 14

## Checks

 - [x] I've tested this PR with Vagrant **v2.2.3** and VirtualBox **v6** on **MacOS**

I did encounter an issue where `vagrant_provision` couldn't be ran by vagrant with a permission denied error, I'm unsure why though as I've only modified the contents of the file in an existing instance, and `vagrant_up` worked fine


Mte90:
Fixed the conflicts :-)
This open the door for a VVV mascotte 😂

Issue #1797, opened 2019-05-13
PHPCS WordpressVIPMinimum coding standard not working


nitishkaila:
The **WordpressVIPMinimum** coding standard not working using **vagrant ssh** command. when I enter the **phpcs -i** it should be display all coding standard including **WordpressVIPMinimum**. but it was not working.

![phpcs](https://user-images.githubusercontent.com/33976417/57600895-a4a03980-7578-11e9-974b-33dd56a208d0.png)

## Expected Behaviour
The WordpressVIPMinimum should be work as per coding standard using vagrant ssh command.

## Possible Solution
As per the [VIP coding standard](https://github.com/Automattic/VIP-Coding-Standards), it will work with WPCS 1.*. Please see the [minimal requirements](https://github.com/Automattic/VIP-Coding-Standards#minimal-requirements) section.

I have noticed that VVV get latest WPCS package during installation. To fixed this issue, please update the **VVV/provision/phpcs/composer.json** composer package file like following.

From:

` "wp-coding-standards/wpcs": "*",`

To:

` "wp-coding-standards/wpcs": "1.*",`
 
## Steps to Reproduce (for bugs)
Enter following command to reproduce the issue.
1. vagrant ssh
2. cd /srv/www/wordpress-two/public_html/
3. phpcs --standard=WordPressVIPMinimum wp-config.php

tomjn:
I'm going to leave as is, when VIPCS updates to be compatible with WPCS then the issue will be fixed. Otherwise the temporary removal of VIPCS is more likely. We get a lot of people installing VVV on contributor day and for doing work with WP Core, so downgrading WPCS just for those working with VIP is a very heavy price to pay and would cause more problems than it solves

If you need VIPCS in the meantime, you can use it with `composer` to create a self contained local PHPCS install

nitishkaila:
@tomjn 

If we are doing this changes then both the standard working correctly. So any one can use WordPress standard or VIP standard. That is main purpose of my PR. I understand that we need to update downgrading WPCS, but that changes work for both standard.

Also is there any way where I can update vagrant composer packages locally?

PR #1798, opened 2019-05-13
Change composer package

nitishkaila:
[For this issue:](https://github.com/Varying-Vagrant-Vagrants/VVV/issues/1797)

Change composer package for WPCS to work with VIP coding standard.


tomjn:
I'm prioritising WPCS over VIPCS as more people use that, I'm expecting a version of VIPCS compatible with WPCS to be released in the near future

PR #1805, opened 2019-05-14
add an option to vvv-custom.yml to turn off db shares

tomjn:
## Summary:

This shouldn't be necessary, but for the moment this lets us turn off the DB shared folder, the raw DB data will be stored inside the VM. This does mean the DB gets destroyed on every box update though.

It also means that toggling this switch and reprovisioning will have strange behaviour

For example:

 - if you provision, then turn it off, you'll get a new set of database files and a clean slate, none of the existing databases will be present, but they'll still be on the host in the database/data folder, they just wont be used
 - If you then switched it off, it'd go back to using the database/data folder, but there would still be a database in the VM that's silenced and innaccessible

## Checks

 - [x] I've tested this PR with Vagrant **v2.2.4** and VirtualBox **v6.0.6** on **MacOS**
 
This still needs testing

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -216,6 +216,10 @@ if ! vvv_config['vm_config'].kind_of? Hash then
   vvv_config['vm_config'] = Hash.new
 end
 
+if ! vvv_config['general'].kind_of? Hash then
+  vvv_config['general'] = Hash.new
+end
+
 defaults = Hash.new
 defaults['memory'] = 2048
 defaults['cores'] = 1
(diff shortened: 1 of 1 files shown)


tomjn:
VVV config general section would look like this:

```yaml
# General VVV options
general:
  # Backup the databases to the database/backups subfolder on halt/suspend/destroy, set to false to disable
  db_backup: true
  # Import the databases if they're missing from backups
  db_restore: true
  db_share_type: false
```

PR #1809, opened 2019-05-15
Copy all the provision output in a log file

Mte90:
In this way the log is more simple to read and share.


tomjn:
I'd rather redirect these somewhere than silence them. Otherwise debugging issues would become extremely difficult, especially considering some of the issues being actively debugged are relying on the output of those commands to test things. We should never silence things but instead redirect their output to log files

Mte90:
What do you think of a file in /vagrant/log/vvv/log-[date].log?

tomjn:
@Mte90 I'd like a new mounted folder `/var/log/provisioners/` then have a new file for each provision. We can clean up any files older than a week. The file names should follow the Vagrant provisioner names, e.g. `provisioner-default.log` or `provisioner-utility-core-tls-ca.log`, with a suitable timestamp/date.

Mte90:
Ok I will try to work on that in the next days.

Mte90:
This is working, creates a new file in a folder.

tomjn:
@Mte90 I made some changes:

 - it now logs to a folder so that each provisions logs are all co-located
 - I put a `provisioned_at` file in `/vagrant` so that they all have the same timestamp, otherwise they'd be stagged as each provisioner runs one after another with a greater timestamp ( also handy for the dashboard to remind users when they last reprovisioned )
 - I added the log snippet to the other core provisioners so we get utilities/sites logged too

Needs testing though

Mte90:
I force pushed because my version is better and working on creating the folder and later the file without issues

PR #1823, opened 2019-05-24
fix composer permission

Mte90:
Require that the folder in `/home/vagrant/.composer/cache` has the wrong permission so is enough to remove it and this patch fix everything.


tomjn:
Also keep in mind that the `noroot` here was removed as it resolved an issue during testing, and that `/srv/provisioner` is a shared folder so its contents will always have the same owner

Issue #1855, opened 2019-06-12
TideWays breaking xdebug in PHPStorm


RexAK:
## Expected Behavior

Download newest VVV, add custom site via vvv-custom.yml copy, vagrant up, then start new project in PHPStorm, turn xdebug on, and set a breakpoint in PHPStorm. Expected that when I load a WP page for that custom site, debugger hits breakpoint and awaits user control
## Current Behavior

Instead, debugger in PHPStorm pauses at unknown point and gives an error about path mapping for "tideways-header.php". If you press step it hits the phantom breakpoint a few times with the same error, and then the page loads, never stopping at the selected breakpoint. Unable to add path mapping for this file
## Possible Solution

At first, I tried removing "mongodb" and "tideways" from the vvv-custom.yml and re-provisioned. Issue persists. Looking at phpinfo, I see the "auto_prepend_file" param still points to the tideways-header.php file. I track this down and comment out the line in /etc/php/7.2/fpm/conf.d/20-xhgui.ini, restart php, make sure path mappings are configured correctly in PHPStorm (I had already done this, just double checking), and the breakpoints work. Should the /etc/php/7.2/fpm/conf.d/20-xhgui.ini auto_prepend_file be removed if I remove mongo and tideways from my custom yml?

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Context

 Just trying to debug. Can work around this, but would like to know if there is a a way to work this into a custom provisioner

## Your Environment

* VVV version: 3.1.0
* VVV Git Branch: download
* Vagrant version:2.2.4
* VM Provider name: VirtualBox
* VM Provider version:v5.2.2r119230
* Operating System and version: macOS 10.12.6

tomjn:
the tideways header is somewhat self contained, is it not possible to ignore or skip over it? @Mte90 any advice?

Mte90:
It is autoloaded by php but that ini file should be removed if you are provisioning without them. 
I will check the code.

Mte90:
This pr enable a parameter for xDebug to let work with auto_prepend_file option.
The second step is understand if without the tideways provider there will be still issues (the code is simple so there shouldn't be errors https://github.com/Varying-Vagrant-Vagrants/vvv-utilities/blob/master/tideways/tideways-header.php).

tomjn:
@RexAK can you confirm that #1856 resolves the issue?

RexAK:
[...]

PR #1856, opened 2019-06-13
fix for auto_prepend_file

Mte90:
xDebug require an option enabled to work without issues with auto_prepend_file parameter.

ref: https://xdebug.org/docs/all_settings

Fix: #1855


tomjn:
Have we confirmed this resolves the problem?

PR #1882, opened 2019-06-27
Enables password SSH auth, fixes #1867

### LINKED ISSUE #1867: MySQL with SSH tunnel results to fatal error
In my case the MySQL connection information provided [here](https://github.com/Varying-Vagrant-Vagrants/VVV/wiki/Connecting-to-MySQL-MariaDB#ssh-tunnel) are no longer relevant.

I'm using HeidiSQL, which fails to connect to my VMs mariadb with an error message:

```
PLink exited unexpected. Command line was:
"c:\Program Files\PuTTY\plink.exe" -ssh vagrant@vvv.test -pw "vagrant" -N -L 3307:127.0.0.1:3306
```

When running this command from the command line, I get

```
"c:\Program Files\PuTTY\plink.exe" -ssh vagrant@vvv.test -pw "vagrant" -N -L 3307:127.0.0.1:3306
Using username "vagrant".
FATAL ERROR: Disconnected: No supported authentication methods available (server sent: publickey)
```

Note that the exact same configuration was working properly with my previous VM.

tomjn:
## Summary:

This attempts to fix #1867

To test, run:

```sh
git checkout origin fix/dbssh
vagrant provision
```

Then attempt to connect to the database from an SQL client on the host

## Checks

 - [x] I've tested this PR with Vagrant **v2.2.4** and VirtualBox **v6.0.8** on **MacOS**

### DIFF
--- a/provision/provision.sh
+++ b/provision/provision.sh
@@ -285,7 +285,7 @@ profile_setup() {
   cp -f /srv/config/ssh_known_hosts /etc/ssh/ssh_known_hosts
   echo " * Enabling SSH Password Authentication for database applications"
   echo "PasswordAuthentication yes" >> /etc/ssh/sshd_config
-  systemctl restart ssh
+  systemctl reload ssh
 }
 
 not_installed() {
--- a/provision/provision.sh
+++ b/provision/provision.sh
@@ -282,7 +282,9 @@ profile_setup() {
   fi
 
   echo " * Copying /srv/config/ssh_known_hosts to /etc/ssh/ssh_known_hosts"
- cp -f /srv/config/ssh_known_hosts /etc/ssh/ssh_known_hosts
+  cp -f /srv/config/ssh_known_hosts /etc/ssh/ssh_known_hosts
+  echo " * Enabling SSH Password Authentication for database applications"
+  echo "PasswordAuthentication yes" >> /etc/ssh/sshd_config
 }
 
 not_installed() {
--- a/provision/provision.sh
+++ b/provision/provision.sh
@@ -283,8 +283,9 @@ profile_setup() {
 
   echo " * Copying /srv/config/ssh_known_hosts to /etc/ssh/ssh_known_hosts"
   cp -f /srv/config/ssh_known_hosts /etc/ssh/ssh_known_hosts
-  echo " * Enabling SSH Password Authentication for database applications"
-  echo "PasswordAuthentication yes" >> /etc/ssh/sshd_config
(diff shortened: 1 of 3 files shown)


tomjn:
Of note, it might require a `vagrant halt` and a `vagrant up`, I'm not sure if the change takes immediate effect or not

tomjn:
@Mte90 does the change take effect immediatley or do we need to restart the SSH agent?

Mte90:
Require a restart of the sshd daemon so better a up and halt.

tomjn:
@Mte90 then we should probably do that in the script so that it works on the first provision

tomjn:
I've made the change, but I'm concerned it might sever any active connections, halting provisioning in the process

Mte90:
well during provision is better to no do anything with the machine to avoid issues so I think that is a best practice that in this case confirm that case 😁

provision/provision.sh
```diff
@@ -282,7 +282,10 @@ profile_setup() {
   fi
 
   echo " * Copying /srv/config/ssh_known_hosts to /etc/ssh/ssh_known_hosts"
- cp -f /srv/config/ssh_known_hosts /etc/ssh/ssh_known_hosts
+  cp -f /srv/config/ssh_known_hosts /etc/ssh/ssh_known_hosts
+  echo " * Enabling SSH Password Authentication for database applications"
+  echo "PasswordAuthentication yes" >> /etc/ssh/sshd_config
```
tomjn: oh dear, I realise now this file is going to get very large with the same line repeated over and over again

tomjn:
So this works for me, but the appending to the config needs changing so it only happens if it's needed

provision/provision.sh
Mte90: maybe with `cat /etc/ssh/sshd_config | grep 'PasswordAuthentication yes'``we can do a check if it is already on it

Mte90:
```
if [[ ! ((cat /etc/ssh/sshd_config | grep 'PasswordAuthentication yes')); then
    echo " * Enabling SSH Password Authentication for database applications"
    echo "PasswordAuthentication yes" >> /etc/ssh/sshd_config
    systemctl restart ssh
  fi
```
should fix the behaviour

tomjn:
Solved by copying `sshd_config` into place

Mte90:
sounds better

kmgalanakis:
Just confirming the fix. Thanks ❤️

PR #1883, opened 2019-06-27
fix #1880 - downgrade nodejs 11

### LINKED ISSUE #1880: Reinstall Node if v11 is installed
WP Core needs Node 10 LTS, and we install that now, so all good for new users!

But for existing users who installed VVV when we used Node v11, that's still a problem. We need to check for Node v11, and if it's found, uninstall it, update apt, then reinstall Node

Mte90:
requires a bit of testing :-)


tomjn:
Needs a changelog update but looks like it'll do the trick

tomjn:
Tested and working :)

Issue #1886, opened 2019-06-27
Incorrect permissions in /vagrant


paulschreiber:
## Expected Behavior
File `/vagrant/vvv-custom.yml` is readable by www user.

## Current Behavior
File `/vagrant/vvv-custom.yml` is readable only by root.

## Steps to Reproduce (for bugs)
```
vagrant@vvv:/vagrant$ ls -l
total 12
lrwxrwxrwx 1 root root   17 Jun 27 22:34 certificates -> /srv/certificates
-rw-r--r-- 1 root root   16 Jun 27 22:34 provisioned_at
-rw-r--r-- 1 root root    5 Jun 27 22:34 version
-rw------- 1 root root 1483 Jun 27 22:34 vvv-custom.yml
````

## Your Environment

* VVV version: git
* VVV Git Branch: develop
* Vagrant version: 2.2.5
* VM Provider name: VirtualBox
* VM Provider version: 6.0.8
* Operating System and version: macOS 10.14.5

benlumia007:
those permissions are correct, when you are at /vagrant, these files don't get edited by the vagrant user, its provisioned beforehand. 

by default, the /vagrant is disabled in the vagrantfile and its re-created with root, along with the files above. u shouldn't be editing this inside of vm, rather than the host.

paulschreiber:
This prevents the dashboard (https://vvv.test) from working correctly:
![image](https://user-images.githubusercontent.com/86784/60307802-17d8f180-9914-11e9-84ee-1240b0d6a750.png)

benlumia007:
So, I just checked mine and have the same version as 3.1.0, It seems like I don't have that problem. 
![Screenshot from 2019-06-27 17-13-10](https://user-images.githubusercontent.com/11844123/60308682-e43f9c80-98fe-11e9-9819-e041c6ce8a4b.png)

looks fine to me.

benlumia007:
maybe try a reload --provision and see if that fixes it. it could just be a bad provisioning

tomjn:
Nope, it needs a `chown` to be done in a provisioner

benlumia007:
`chown -R vagrant:vagrant /vagrant` should do it

benlumia007:
@Mte90 
I can do this one if you up for it.

Mte90:
ok, is not assigned to someone so you can do it and later we will test it

benlumia007:
I'll work on this today

[...]

PR #1891, opened 2019-06-29
change ownership from root to vagrant for /vagrant

benlumia007:
## Checks

 - [x] I've tested this PR with Vagrant **2.2.4** and VirtualBox **6.0.8** on **LInux**
 
 

The ownership should be now be `vagrant`
`drwxr-xr-x   2 vagrant vagrant  4096 Jun 29 22:15 vagrant`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ This is primarily a reliability update. Note that updating to v3.1 requires a `v
  - Node v11 is now auto-downgraded to Node v10
  - Fixed Database SSH access from the host by enabling password authentication in `/etc/ssh/sshd_config`
  - Added code to remove NVM
+ - Change Permission folder `/vagrant` from root to vagrant
 
 ## 3.0.0 ( 17 May 2019 )
 
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -524,6 +524,9 @@ cp -f /home/vagrant/vvv-custom.yml /vagrant
 # symlink the certificates folder for older site templates compat
 ln -s /srv/certificates /vagrant/certificates
 sudo sed -i '/tty/!s/mesg n/tty -s \\&\\& mesg n/' /root/.profile
+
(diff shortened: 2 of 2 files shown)


paulschreiber:
`sudo chown -R vagrant:vagrant /vagrant` is a partial fix. you also need chmod a+r vvv-custom.yml (or similar.

benlumia007:
I don't think it matters, the user edits the vvv-custom.yml file from their own computer and it gets copy to the vm home directory and copies again to the /vagrant folder then it only gets read and passs so that the dashboard can get the information from.

paulschreiber:
Again, the webserver needs to be able to read the vvv-custom.yml file to render https://vvv.test. See #1886.

tomjn:
@benlumia007 the config file gets read by the provisioners and the VVV dashboard, so it needs to be readable

benlumia007:
even if it gets read, it wouldn't matter much, you aren't really editing the file when you vagrant up after you finish editing the file so that it creates the site, this is more about the root and /vagrant folder that is contained files inside. even if it was under root or vagrant, is still readable, isn't. if it wasn't readable, then everyone would have problems loading sites and the dashboard. 

The last comment i posted, i never said that the file doesn't get read, i pretty sure i said, it gets read so that the dashbarod can get the information or anything that gets used by shyaml.

benlumia007:
the /vagrant folder only contains the vvv-custom.yml, provisioning_at and version,  it does not contain any sites files/folders for that matter, the www is own by vagrant:www-data,  the /vagrant can be own by root or vagrant because its readable depends on which user is been used to create the /vagrant folder.

benlumia007:
The web server has nothing to do with the vvv-custom.yml file,  the web server is just allows you to run web service. the vvv-custom.yml is just a file that gets used by `shyaml` to gain access to the vvv-custom.yml and process it and generates a site or feature that is contained in the file.

PR #1892, opened 2019-07-02
Reset color after printing splash (to avoid weird coloring after)

msaggiorato:
## Summary:

Added a simple color reset to avoid branch_c bleeding over to the some other output.


tomjn:
Thanks @msaggiorato!!

PR #1896, opened 2019-07-04
fix #1417 - Github Token in vvv-config

### LINKED ISSUE #1417: Put GitHub Token in vvv-custom.yml
There's an option for using a GitHub token in:

https://github.com/Varying-Vagrant-Vagrants/VVV/blob/develop/provision/provision.sh#L339

It's undocumented, but it should be in `vvv-custom.yml`:

 - pull the token from the config
 - add a docs page under references

Mte90:


tomjn:
hmm actually no, it needs documenting and changelog'ing but no need to check if the file exists

Mte90:
Added the changelog and for documenting I am working on that

Mte90:
And docs https://github.com/Varying-Vagrant-Vagrants/varyingvagrantvagrants.org/pull/130

PR #1902, opened 2019-07-09
remove trailing whitespace from .editorconfig

paulschreiber:
## Summary:

remove trailing whitespace from .editorconfig


Mte90:
Good catch :-D

PR #1904, opened 2019-07-16
Update vvv-config to support the new parameters on custom-template

Mte90:
Add the new parameters from https://github.com/Varying-Vagrant-Vagrants/custom-site-template/pull/16 and https://github.com/Varying-Vagrant-Vagrants/custom-site-template/pull/14


Mte90:
ready to be approved :-)

PR #1915, opened 2019-08-01
Add support for vagrant-disksize plugin.

msaggiorato:
## Summary:

This adds support (and checks for the availabilty) for the [vagrant-disksize](https://github.com/sprotheroe/vagrant-disksize) plugin.

This will allow to bypass issues related to disk space issues when `db_share_type` is set to false, or if somebody just wants more space.

## Checks

 - [x] I've tested this PR with Vagrant **v2.2.4** and VirtualBox **v6.0.8** on **macOS**

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -279,6 +279,10 @@ if show_logo then
     platform = platform + 'vagrant-vbguest '
   end
 
+  if Vagrant.has_plugin?('vagrant-disksize') then
+    platform = platform + 'vagrant-disksize '
+  end
+
   if Vagrant::Util::Platform.fs_case_sensitive? then
     platform = platform + 'CaseSensitiveFS '
   end
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -230,6 +230,7 @@ end
 defaults = Hash.new
 defaults['memory'] = 2048
 defaults['cores'] = 1
+defaults['disksize'] = '10GB'
 # This should rarely be overridden, so it's not included in the default vvv-config.yml file.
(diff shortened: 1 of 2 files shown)


Mte90:
seems a good idea!
I think that to approve we need an integration also in that part https://github.com/Varying-Vagrant-Vagrants/VVV/blob/f2282b1ef26679b1b9c35bbd84468309c5511206/Vagrantfile#L278 so the user see what are the plugin enabled and supported.
We will update the docs after the merge for the website :-D

msaggiorato:
Done @Mte90

tomjn:
I can see users raising issues that they set this and it didn't change their disk size. If we're going to do this it needs to be crystal clear, so it should:

1. make it crystal clear from the option name that the plugin is needed ( assume in the future the YAML file will be rewritten by machines and the comment will dissapear )
2. complain out loud if the user sets it but the plugin isn't present

msaggiorato:
Ok, makes sense. I'll deal with this when i get a chance and update you guys. Do you have any suggestion for the option name @tomjn ? Maybe start a new setting group for plugins ?

tomjn:
How about:

```yaml
vagrant-plugins:
  disk-size: 10GB
```

I'd leave the option out of `vvv-config.yml` and instead document it on the website

msaggiorato:
Hey  @tomjn and @Mte90  I've commited the changes you requested, I've left the option name as disksize (without the hyphen) to maintain consistency with the plugin name.

Let me know if this works for you.

tomjn:
did a rebase and merge to avoid having to faff around to update a markdown file

Issue #1887, opened 2019-06-27
grunt not installed correctly after box creation/provisioning


paulschreiber:
grunt is not installed correctly after box creation/provisioning. Possibly the same as/related to #1885.

## Expected Behavior
grunt is installed

## Current Behavior
```
Command 'grunt' not found, but can be installed with:

apt install node-grunt-cli

Please: command not found
```

## Steps to Reproduce (for bugs)
1. vagrant up --provision
2. vagrant ssh

## Workaround
```
sudo npm install -g grunt-cli grunt
```

## Your Environment
* VVV version: git
* VVV Git Branch: develop
* Vagrant version: 2.2.5
* VM Provider name: VirtualBox
* VM Provider version: 6.0.8
* Operating System and version: macOS 10.14.5

Mte90:
this should fix the issue https://github.com/Varying-Vagrant-Vagrants/VVV/pull/1889

paulschreiber:
This does not appear to be fixed in 3.1.0.

Mte90:
did you updated to develop branch? did a `vagrant destroy`?

paulschreiber:
Yes, I updated to 3.1.0 (`git co 3.1.0`), and then ran `vagrant destroy and vagrant up --provision`.

Mte90:
I cannot replicate the issue anymore, what kind of error you get?

paulschreiber:
See `current behaviour` in the description above. Upon login (`vagrant ssh`) that message is printed.

Was `3.1.0` tagged incorrectly? Did this commit not make it in?

tomjn:
I have no idea what the current state of this is, and googling on how to properly install things has been super unhelpful. There are some unanswered questions:

 - Why does it work on a fresh provision for me and @Mte90 but not for others
 - The current behaviour in the first note of this issue has no context, it isn't clear what command generated it, be it using `grunt` in a terminal, opening a new SSH session, or trying to use `npm`
 - Should we use `sudo` or not to install grunt? Googling gives many, many answers that all contradict eachother
 - `npm install -g grunt` won't work as the vagrant user because the node modules etc folders are all owned by the system, which suggests that `sudo` is required for it to work, but previous attempts to resolve this have only lead to the problem temporarily being fixed, be it in a singular folder, or for that individual SSH session, not a true fix. I don't know why
 - When we made some changes and asked if the problem was fixed, we got no response and were unable to reproduce the problem, so the issue was closed

[...]

PR #1916, opened 2019-08-02
Grunt changes

tomjn:
Remove noroot for grunt install, not just the update, and make logging a bit better.

I also don't trust the check to see if grunt is present, it needs double checking that it's ok

## Checks

 - [ ] This PR is for the `develop` branch not the `master` branch

### DIFF
--- a/provision/provision.sh
+++ b/provision/provision.sh
@@ -553,9 +553,9 @@ tools_install() {
   }
   chown -R vagrant:vagrant /usr/lib/node_modules/
   if command -v grunt >/dev/null 2>&1; then
-    install_grunt
-  else
     update_grunt
+  else
+    install_grunt
   fi
 
   # Graphviz
--- a/provision/provision.sh
+++ b/provision/provision.sh
@@ -438,6 +438,7 @@ latest_github_release() {
 }
 
 tools_install() {
+  echo " * Running tools_install"
   # Disable xdebug before any composer provisioning.
   sh /srv/config/homebin/xdebug_off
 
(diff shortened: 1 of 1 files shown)


tomjn:
related to https://github.com/Varying-Vagrant-Vagrants/VVV/issues/1887

tomjn:
@paulschreiber can you confirm if the conditional that checks for grunt is working correctly? I suspect in your case, it'll be running the update not the install function

paulschreiber:
```
git fetch --all
git co modify/grunt
vagrant destroy
vagrant box up --provision
vagrant ssh
```

Same problem:
```
Command 'grunt' not found, but can be installed with:

apt install node-grunt-cli

Please: command not found
```

tomjn:
@paulschreiber do you have the provisioner log for that? It's not clear if the conditional I mentioned failed, or if it tried correctly to install `grunt` but failed, or if neither happened

paulschreiber:
Here's the log:
https://gist.github.com/paulschreiber/82ce3cef4da4439fcbf93024b43043e3

tomjn:
Here we are:

```
    default: Generating autoload files
    default:  * Updating Grunt CLI
    default:  * Updated Grunt CLI
    default: Adding graphviz symlink for Webgrind...
```

tomjn:
@paulschreiber if you find this in `provision/provision.sh`:

```bash
  if command -v grunt >/dev/null 2>&1; then
    install_grunt
  else
    update_grunt
  fi
```

and swap them around to this

```bash
  if command -v grunt >/dev/null 2>&1; then
    update_grunt
  else
    install_grunt
  fi
```

Then reprovision, does it install grunt? Or does it still try to update a grunt that isn't installed?

tomjn:
Further testing here suggests it is indeed back to front, I've changed it around, a `git pull` and a `vagrant provision` should be all that's needed to test

paulschreiber:
Confirmed this is now fixed. Did a `git pull`, `vagrant destroy`, `vagrant up --provision` and `vagrant ssh`.

PR #1918, opened 2019-08-06
Changed db_share_type checks and defaults

tomjn:
Changed how the db_share_type was checked, and set its default value if undefined, to off

We know that off always works, so lets switch to that by default and avoid the support hassle for end users at the same time. Decisions not options! They can always opt into it in future

Additionally, the splash screen checked if the value was true, or anything else. The actual conditional checked if it was false, or anything else. Now both check if it's true or anything else for consistency

## Checks

@TODO: Update the changelog before merge as this might cause issues for users who use the shared folder and run `develop`. We should also update this PR with a message telling people the value is undefined and what VVV is going to do by default and how to change it


Mte90:
I saw late the message about the changelog -.-'

PR #1923, opened 2019-08-09
YAML validation, fix #1745

### LINKED ISSUE #1745: Add yamllint support
## Expected Behavior
When your vvv-custom.yml is broken and you start the provision but you get an error that block to start it

## Current Behavior
RIght now if is broken is executed anyway and provisioning can be broken and is not easy to spot because the log is huge.

## Possible Solution
Add yamllint or another solution more light that check if it is good.

Mte90:
This code add an exception in case the yaml file is wrong with an alert.
Also a check if sistes is defined or not.

I am looking for other stuff to check :-)

PS: seems that my editor removed all the useless whitespaces

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -139,10 +139,20 @@ end
 
 vvv_config_file = File.join(vagrant_dir, 'vvv-custom.yml')
 
-vvv_config = YAML.load_file(vvv_config_file)
+begin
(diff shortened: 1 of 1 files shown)


Mte90:
Screenshot_20190809_162103
This is the screen and the approval is for you :-D

PR #1948, opened 2019-09-30
ensure my.cnf is readable by vagrant user

paulschreiber:
## Summary:

Prior to this, `my.cnf` is only readable by root. Running `mysql dbname` results in a permissions error.

## Checks


provision/provision.sh
```diff
@@ -711,6 +711,7 @@ mysql_setup() {
     # Copy mysql configuration from local
     cp "/srv/config/mysql-config/my.cnf" "/etc/mysql/my.cnf"
     cp "/srv/config/mysql-config/root-my.cnf" "/home/vagrant/.my.cnf"
+    sudo chmod 0644 "/home/vagrant/.my.cnf"
```
tomjn: Is the `sudo` needed? This file should run as the `root` user already and be copied/symlinked into place further down the file

tomjn:
🎉

PR #1954, opened 2019-10-07
Enable mailhog on all PHP instances

### LINKED ISSUE #1953: Mailhog config not copied to custom PHP upstreams
## Expected Behavior
When setting nginx_upstream in vvv config, MailHog ini file is not created for the selected PHP version.

## Possible Solution
Add the MailHog configuration to /srv/php/<nginx_upstream php version>/fpm/conf.d/20-mailhog.ini
Would passing nginx_upstream php version as an argument to provision.sh and using the version as a variable in the script be a solution? 

1. Add `nginx_upstream: php73` to any site in vvv-custom.yml
2. Reprovision
3. Try sending an email with wp_mail
4. Email is not caught by MailHog

av3nger:
## Summary:

Fixes #1953 

## Checks

 - [x] I've tested this PR with Vagrant **v3.1.1** and VirtualBox **v6.0.12** on **macOS 10.15**

### DIFF
--- a/provision/provision.sh
+++ b/provision/provision.sh
@@ -764,7 +764,7 @@ services_restart() {
   phpdismod xdebug
 
   # Enable PHP mailcatcher sendmail settings by default
-  phpenmod mailhog
+  phpenmod -s fpm mailhog
 
   # Restart all php-fpm versions
   find /etc/init.d/ -name "php*-fpm" -exec bash -c 'sudo service "$(basename "$0")" restart' {} \;
--- a/provision/provision.sh
+++ b/provision/provision.sh
@@ -763,7 +763,7 @@ services_restart() {
   # Disable PHP Xdebug module by default
   phpdismod xdebug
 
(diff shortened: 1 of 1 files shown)


provision/provision.sh
```diff
@@ -764,7 +764,7 @@ services_restart() {
   phpdismod xdebug
 
   # Enable PHP mailcatcher sendmail settings by default
```
Mte90: can you replace also mailcatcher here with mailhog?

Mte90:
I think that is missing also https://github.com/Varying-Vagrant-Vagrants/VVV/issues/1953#issuecomment-538741001
Enable to all the php version is no very useful, mailhog is not in php and there is only one version as default.

av3nger:
@Mte90 , but won't it be there once you provision the php73 via the utilities/core provision scripts? At least the above worked on my test install.

provision/provision.sh
av3nger: done

Issue #1922, opened 2019-08-08
v3.2


tomjn:
I'd like to invetigate making the following changes for 3.2:

 - **Improve log files**, right now they're `.log` files which is a pain in the arse on MacOS, lets make them `.txt` files so they open in a nicer app by default. Also, right now they're all separate files, but it'd be great to have an extra file that just has the entire log
 - **Move `vvv-config.yml` and `vvv-custom.yml`**, I would like to move these into the `config` folder so they're back in a mapped folder. It would be good to de-emphasise `vvv-config.yml` too
 - **Dashboard update notification** (#1333), we have an issue for this already, but a JS snippet that pokes the GH releases API and checks the value of `/vagrant/version` against the newest, and if they don't match, put a banner in the dashboard
 - ~**Reload Nginx in the core site provisioner**, this way site templates don't need to, and we can make future optimisations~
 - **Put a standardised log message with colour at the start of each provisioner**, our provisioner output is a bit blank now :( lets reintroduce colours and make it easier to read

I think this should make a nice solid 3.2 release, focusing on polish and quality of life, what does everybody think? @Mte90 @benlumia007 @LoreleiAurora

Mte90:
* I don't see very good to do a big file with all the log that can be difficult to read, maybe we can integrate something that publish automatically on pastebin instead as various files.
* I think that we can move `vvv-config.yml` to config itself so people is not interested on altering it. I am not sure about `vvv-custom.yml`

For the other points let's do it!

My other proposals:

* Integrate in the dashboard the enabling of the tideways provisioner and also of xdebug. So people will be not required to access to ssh (this open also the floor to the other script that we offer).
* Move on the documentation starting with finish https://github.com/Varying-Vagrant-Vagrants/varyingvagrantvagrants.org/pull/126
* Integrate yamllint https://github.com/Varying-Vagrant-Vagrants/VVV/issues/1745
* Webpack output to custom log file https://github.com/Varying-Vagrant-Vagrants/VVV/issues/1868
* Integrate the WP-CLI provisioner https://github.com/Varying-Vagrant-Vagrants/VVV/issues/1732

tomjn:
> I don't see very good to do a big file with all the log that can be difficult to read, maybe we can integrate something that publish automatically on pastebin instead as various files.

It makes it easier for users to send us the entire thing or put it in a gist, especially for Windows users. In particular I'd really like to see the splash screen info in there

[...]

PR #1955, opened 2019-10-09
Move config file

msaggiorato:
## Summary:

Partially address https://github.com/Varying-Vagrant-Vagrants/VVV/issues/1922

Moves config file to config folder, and perform the migration of the old version to the new one. Also changed comments and puts to reflect the new location.

## Checks


msaggiorato:
Not sure if this labeler issue is a problem on the PR, please let me know. I've provisioned with these changes and everything worked (and vvv-custom was moved accordingly).

Vagrantfile
```diff
@@ -553,16 +557,13 @@ sudo chown -R vagrant:vagrant /vagrant
...
-rm -f /vagrant/vvv-custom.yml
 
 touch /vagrant/provisioned_at
 echo `date "+%Y%m%d-%H%M%S"` > /vagrant/provisioned_at
 
 # copy over version and config files
 cp -f /home/vagrant/version /vagrant
-cp -f /home/vagrant/vvv-custom.yml /vagrant
 
-sudo chmod 0644 /vagrant/vvv-custom.yml
```
Mte90: Wondering if we put that to avoid some conflicts or other issues on provisions on some systems. what do you think @tomjn?

tomjn:
That fixed an issue Paul Schreiber was having with the file being unreadable

Mte90:
So is better to leave that chmod to avoid any issue

Vagrantfile
Mte90: @msaggiorato better to give back that command :-) and is ready to be tested

msaggiorato:
But now that the file is on the config folder, i'd say it should be readable with no problem. If the file is not readable, then likely the whole provisioner wont work as the config folder is quite critical to it.

Am I wrong?

Before this change, the vvv-custom and vvv-config files were provisioned as files, not synced folders.

Correct me if I'm wrong

tomjn:
> Before this change, the vvv-custom and vvv-config files were provisioned as files, not synced folders.

Those files still have to be copied into place in `/vagrant`, as a record of what was provisioned. That way even if `config/config.yml` changes we know that if they're different a reprovision is needed.

As for why that file was unreadable and needed the `chmod`, ¯\_(ツ)_/¯, I never encountered the issue myself, but someone did, and for them it was necessary. Without it, the dashboard crashed.

.gitignore
```diff
@@ -11,7 +11,7 @@ Icon\r\r
 
 # This is a file intended for hooking in a custom Vagrant configuration on up
 /Customfile
-/vvv-custom.yml
```
tomjn: I think we should still ignore this file

Vagrantfile
```diff
@@ -125,9 +125,16 @@ end
 
 # Load the config file before the second section of the splash screen
 
-if File.file?(File.join(vagrant_dir, 'vvv-custom.yml')) == false then
-  puts "#{yellow}Copying #{red}vvv-config.yml#{yellow} to #{green}vvv-custom.yml#{yellow}\nIMPORTANT NOTE: Make all modifications to #{green}vvv-custom.yml#{yellow} in future so that they are not lost when VVV updates.#{creset}\n\n"
-  FileUtils.cp( File.join(vagrant_dir, 'vvv-config.yml'), File.join(vagrant_dir, 'vvv-custom.yml') )
+vvv_config_file = File.join(vagrant_dir, 'config/config.yml')
```
tomjn: We also need to copy `vvv-custom.yml` to `config/config.yml` if it exists for backwards compatibility

Mte90: uh I missed that

tomjn:
We also need to update the changelog, aside from the bullet point, there'll need to be a small paragraph explaining the change

Vagrantfile
msaggiorato: But it is copying `vvv-custom.yml` to `config/config.yml` for backwards compatibilty.

What am I missing?

msaggiorato:
> > Before this change, the vvv-custom and vvv-config files were provisioned as files, not synced folders.
> 
> Those files still have to be copied into place in `/vagrant`, as a record of what was provisioned. That way even if `config/config.yml` changes we know that if they're different a reprovision is needed.
> 
> As for why that file was unreadable and needed the `chmod`, ¯_(ツ)_/¯, I never encountered the issue myself, but someone did, and for them it was necessary. Without it, the dashboard crashed.

If we need to do this, should we always be reading `/vagrant/config.yml` then?

Vagrantfile
tomjn: hmm I've misunderstood, it needs to copy `config/default-config.yml` to `config/config.yml` too


```
if config/config.yml doesn't exist
    if vvv-custom.yml exists
        copy that to config/config.yml
    else
        copy config/default-config.yml to config/config.yml
```

It should also mention the filenames explicitly so we don't confuse any users

msaggiorato: This is what's it's currently doing with my changes

```
if vvv-custom.yml exists
   copy that to config/config.yml

if config/config.yml doesn't exist
    copy config/default-config.yml to config/config.yml
```

Pretty much the same.

tomjn: I've committed an update that refactors that bit

msaggiorato: Awesome, thanks!

tomjn:
> If we need to do this, should we always be reading /vagrant/config.yml then?

That depends on the use case, but for the provisioners purposes it should always be the original. Eventually it'll be Ansible reading it in anyway.

I think all we need to do now is update the changelog. I think we should hold off a bit on merging it in though so we don't catch users on the `develop` branch offguard, ideally we do this shortly before a 3.2 release so we can update the documentation at the same time

msaggiorato:
Awesome, should I update the changelog? Or you have it?

tomjn:
if you can wrangle the changelog? :)

msaggiorato:
Should we do the changelog update separately? Seems like there's a bunch of things that changed also outside of this PR since it was updated.

msaggiorato:
> Also, now that the original config file is in a synced folder again, we can remove the vagrant provisioner that copies those files to `/vagrant` and just do it bash, see #1958

Pushed a change that address #1958. Could there be any problems with provision-pre.sh? What do people usually do in there?

tomjn:
Not many people use `provision-pre.sh`, considering at some point there won't be a `provision.sh` anyway I'm fine with that.

Also we should update the changelog in this PR, a rebase on to `develop` to bring in the newer changes should resolve the conflict, afterall that file is only being moved/renamed

msaggiorato:
Rebased

tomjn:
We're a WIP docs update PR, we'll need one for the dashboard too

PR #1966, opened 2019-10-11
Add coloring to the site provisioner

msaggiorato:
This partially addresses https://github.com/Varying-Vagrant-Vagrants/VVV/issues/1922, and adds messages when starting and finishing provisioning.

There was already a message when provisioning should be skipped, but i wasn't able to test that, since whenever the skip_provisioning option is not false, the provisioner is not run at all.

In any case, I don't foresee any problems coming out of this.

### DIFF
--- a/provision/provision-site.sh
+++ b/provision/provision-site.sh
@@ -101,10 +101,17 @@ function vvv_provision_hosts_file() {
   done
 }
 
+GREEN="\033[38;5;2m"
+BLUE="\033[38;5;4m"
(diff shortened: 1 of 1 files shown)


Mte90:
I like it, we removed previously the output colored because with composer and other commands there was false positive as "red".
We can add colors also in other part of the various provisioners indeed

tomjn:
nice, I was thinking of adding a standard header with some colour to announce the provisioner has started, but adding colour to all the other messages we add is nice too

PR #1970, opened 2019-10-14
attempt to pipe all site provisioner output to the log files

tomjn:
## Summary:

This attempts to pipe all site template provisioning to the log file, bypassing the terminal, needs testing and polish

## Checks

### DIFF
--- a/provision/provision-site.sh
+++ b/provision/provision-site.sh
@@ -147,13 +147,21 @@ if [[ ! -d ${VM_DIR} ]]; then
   echo "Error: The ${VM_DIR} folder does not exist, there is nothing to provision for the '${SITE}' site!"
 fi
 
+function vvv_run_site_template_script() {
+  echo "Found ${2} at ${1}/${1}"
+  echo "Script output will be logged to: ${logfile}"
+  ( cd "${DIR}" && source $1  2>&1 | tee -a $logfile )
+  echo "${2} has completed with return code: ${?}"
+}
+
 # Look for site setup scripts
+echo "Searching for site template provisioner, vvv-init.sh"
 if [[ -f "${VM_DIR}/.vvv/vvv-init.sh" ]]; then
-  ( cd "${VM_DIR}/.vvv" && source vvv-init.sh )
+  vvv_run_site_template_script "vvv-init.sh" "${VM_DIR}/.vvv"
 elif [[ -f "${VM_DIR}/provision/vvv-init.sh" ]]; then
-  ( cd "${VM_DIR}/provision" && source vvv-init.sh )
+  vvv_run_site_template_script "vvv-init.sh" "${VM_DIR/provision}"
 elif [[ -f "${VM_DIR}/vvv-init.sh" ]]; then
-  ( cd "${VM_DIR}" && source vvv-init.sh )
+  vvv_run_site_template_script "vvv-init.sh" "${VM_DIR}"
 else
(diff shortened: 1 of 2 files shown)


tomjn:
This is functional, the final part is that it needs to output in red if the provisioner script failed. The return code doesn't seem to work for that, so further investigation is needed into what to do next

provision/provision-site.sh
```diff
@@ -111,7 +111,7 @@ if [[ true == $SKIP_PROVISIONING ]]; then
   return
 fi
 
-echo -e "${BLUE}Running provisioner for site ${SITE}${CRESET}"
+echo -e "${GREEN}Running provisioner for site ${SITE}${CRESET}"
```
msaggiorato: Shouldn't we standarize colors like this?

Green: Success
Red: Error
Yellow: Warning
Blue: Info

That's the reason I used blue.

Mte90:
Looking at https://medium.com/@dirk.avery/the-bash-trap-trap-ce6083f36700 we can add at the top on this provisioner:

```
set -e
trap 'catch $? $LINENO' EXIT

catch() {
  echo "catching!"
  if [ "$1" != "0" ]; then
    # error handling goes here
    echo "Error $1 occurred on $2"
  fi
}
```
We can evaluate `set -ex` in case.

tomjn:
I changed the blue to green because it was almost impossible to read dark
blue on black for me

Mte90:
Right now to understand the terminal background color there isn't a standard way but there are workaround (https://github.com/rocky/bash-term-background/blob/master/term-background.sh).
I think that for the shell color output we can do another pr to address better those values.

msaggiorato:
Maybe we can change to cyan?

tomjn:
I'm not sure we can rely on that, lets just use green red and yellow for now

Mte90:
About the strict execution there is a new pr so better focus on the pipe of the log

Mte90:
I tested right now and there are no issues:
```
==> default: Running provisioner: site-woocommerce (shell)...
    default: Running: /tmp/vagrant-shell20191023-22496-1u3tb3g.sh
    default: Running provisioner for site woocommerce
    default: 
    default: Updating woocommerce in /srv/www/woocommerce...
    default: Searching for site template provisioner, vvv-init.sh
    default: Found vvv-init.sh at /srv/www/woocommerce/provision/vvv-init.sh
    default: Script output will be logged to: /var/log/provisioners/20191023-123027/provisioner-site-woocommerce.log
```

Mte90:
I am getting the error printed like: `default: Deprecation warning: require-dev.mikey179/vfsStream is invalid, it should not contain uppercase characters. Please use mikey179/vfsstream instead. Make sure you fix this as Composer 2.0 will error.` but the majority of stuff is in the log itself.

tomjn:
I’d like to know what the cause of that is before merging. It might be
nothing but better safe than sorry, sure provisioning is at the core of
things so I don’t want to be taking chances

Mte90:
That error is something based on the develop provisioner but it's there since months because of a dependence of phpunit that is abandoned.
I written here as example that is tracking the errors also the "fake positive" and this is good.

tomjn:
Is there a ticket or issue for PHPUnit?

Mte90:
There is of this library https://github.com/bovigo/vfsStream/issues/173 seems that one of the dependence of composer in develop has this but is a non-issue.
Doesn't involve this pr anyway.

PR #1975, opened 2019-10-18
Adjusted default permissions for vmware using umask

### LINKED ISSUE #1974: vmware Permissions Problems
## Expected Behavior
Core, Theme, Plugin updates and uploads all work as expected.

## Current Behavior
No updates or uploads work without providing FTP/SSH creds.

1. Change provider to vmware_desktop
2. Provision
3. Permissions are borked

As far as I can tell, this is caused by not being able to pass `dir_mode` and `file_mode` into vmware. The only thing I've seen that could be a possible solution is to use `umask` but I'm not sure it's possible to replicate the permissions used in the virtualbox environment. So far my attempts at implementing `umask` have been unsuccessful.

robneu:
## Summary:

This uses `umask` to attempt to replicate the default permissions set in VirtualBox as closely as possible. Some are identical, a couple have slightly different file permissions because of how `umask` works, but it doesn't seem to be causing any issues on my end so far.

This should hopefully resolve #1974 

## Checks

 - [x] I've tested this PR with Vagrant 2.2.5 and VMware Workstation 15.5.0 build-14665864 on Ubuntu 18.04.3 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ permalink: /docs/en-US/changelog/
  - `my.cnf` is now readable by the vagrant user
  - Fixes to newline substitution in the splash screen and some rearrangement
  - MySQL binary logging is now disabled
+ - Synced folder permission fixes for VMWare
 
 ## 3.1.1 ( 2019 August 6th )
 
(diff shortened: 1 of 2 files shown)


tomjn:
This looks decent, can we get the change log updated?

PR #1978, opened 2019-10-24
fix #1930

### LINKED ISSUE #1930: Document vagrant disk size in changelog
Related to #1915 we need to document this in the changelog for 3.2

Mte90:
fix #1930 
In the changelog is already documented but the vvv-config wasn't updated.

### DIFF
--- a/vvv-config.yml
+++ b/vvv-config.yml
@@ -138,4 +138,4 @@ general:
 
 # Settings for the vagrant plugins supported by VVV
 vagrant-plugins:
-  disksize: 10GB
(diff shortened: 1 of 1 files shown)


vvv-config.yml
```diff
@@ -135,3 +135,7 @@ general:
   db_share_type: false
   # GitHub token to use from composer
   #github_token: xxxxxx
+
+# Settings for the vagrant plugins supported by VVV
+vagrant-plugins:
+  disksize: 10GB
```
tomjn:
```suggestion
  disksize: 10GB # requires the disk size vagrant plugin
```

PR #1981, opened 2019-10-24
removed /var/log/ for Microsoft Hyper-V, no need

benlumia007:
## Summary:

We have fixed where the /var/log was still available when using Hyper-V, so we removed it, and works now with no issues with MariaDB. 
## Checks

 - [x] I've tested this PR with Vagrant **2.2.6** and HyperV on **Windows 10 Pro**


benlumia007:
I was gonna do the mount_options but, i don't see any difference between them so leaving it as it is for now.

PR #1996, opened 2019-11-05
Switch mariadb repo

Mte90:
as suggested at https://downloads.mariadb.org/mariadb/repositories/#distro=Ubuntu&distro_release=bionic--ubuntu_bionic&mirror=herrbischoff&version=10.3


Mte90:
Tested and everything works for me.

Issue #1531, opened 2018-05-29
grunt watch doesn't detect changes when ran from guest machine


mnelson4:
## Expected Behavior
Since WordPress' new build process was introduced (see https://make.wordpress.org/core/2018/05/16/preparing-wordpress-for-a-javascript-future-part-1-build-step-and-folder-reorganization/) we should be running `grunt watch` which should copy the changes made from the src directory into the build one.

## Current Behavior

However on a windows host machine with with VVV, when you make a change to a file (from your host machine or inside the VM) that's in a shared directory, grunt doesn't notice the change. 

## Possible Solution

This is probably because of https://www.virtualbox.org/ticket/10660 (so it's a Virtual box issue). BUt I'm reporting here in case someone else has this issue and is looking for a fix, and it's possible we might find some workaround for VVV. 

## Steps to Reproduce (for bugs)

1. Do git clone git://develop.git.wordpress.org/ from inside VVV's www shared folder
2. from inside there, run `npm install && grunt build`. Then `grunt watch`. Wait until it says it's "Watching..."
3. Edit a file from inside src
4. Notice that the grunt task doesn't acknowledge that change, nor update the corresponding file in build

## Context

I've been using the same WP develop checkout for both developing a few plugins, and for the occasional core contribution. So for developing plugins I will probably put my `wp-content` folder outside of the normal WP src folder, and use `define( 'WP_CONTENT_DIR', dirname( dirname(__FILE__) ) . '/content' ); define( 'WP_CONTENT_URL', 'http://dev.local/content' );`) so that I don't need to wait for grunt watch to copy all my changes from src to build everytime. But I'm currently a bit stuck for WP core contributions.

## Your Environment

* VVV version: 2.1.0
* VVV Git Branch: master (commit 710f08527730bbeea5b263cd6e6ff1614cb224ea)
* Vagrant version: 2.0.2
* VM Provider name: Virtual Box
* VM Provider version: 5.1.32
* Operating System and version: Windows 7

mnelson4:
created a duplicate of this issue inside WP trac, https://core.trac.wordpress.org/ticket/44271, in case the issue can be solved from that end (or in case someone else experiences it and is searching for it from there)

tomjn:
Eh the Trac ticket will get closed, this is a general vagrant issue. You can fix this at the cost of reduced performance, but it will never be as good as running it on the host

mnelson4:
Here's a hacky workaround: 

[...]

PR #2005, opened 2019-11-18
Vagrant VVV commands

### LINKED ISSUE #1967: Add vagrant commands for the various scripts in homebin
The idea is to do a command like `vagrant vvv xdebug_on` for the scripts on https://github.com/Varying-Vagrant-Vagrants/VVV/tree/develop/config/homebin
Except the `vagrant_*` scripts that are used internally.

Mte90:
Added 2 new commands and fix #1967 and #1531.

```
vagrant command wordpress-develop "ls -l"
```
To see the list of files inside that folder  but can be used for other stuff (grunt watch).

Instead `vagrant script xdebug_on` will execute that script.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -3,15 +3,6 @@
 Vagrant.require_version ">= 2.2.4"
 require 'yaml'
 require 'fileutils'
-require 'optparse'
-require 'ostruct'
-
-options = OpenStruct.new
-OptionParser.new do |opt|
-  opt.on('-w', '--website site-folder', 'Site name in config.yml') { |o| options.website = o }
-  opt.on('-c', '--command your-command', 'Command to execute') { |o| options.command = o }
-  opt.on('-s', '--script script', 'Scripts from config/homebin/') { |o| options.script = o }
-end.parse!
 
 def virtualbox_path()
     @vboxmanage_path = nil
@@ -834,25 +825,6 @@ Vagrant.configure("2") do |config|
     config.hostsupdater.aliases = vvv_config['hosts']
     config.hostsupdater.remove_on_suspend = true
   end
-  if options.script
-    splash = <<-HEREDOC
-      #{yellow}Executing #{red}#{options.script}#{creset}
-    HEREDOC
-    puts splash
-
-    config.vm.provision "shell", keep_color: true,run: 'always', inline: "/srv/config/homebin/#{options.script}"
-    exit()
-  end
-
-  if options.website && options.command
-    splash = <<-HEREDOC
-      #{yellow}Executing command #{red}#{options.command} #{yellow}in #{red}#{options.website}#{creset}
-    HEREDOC
-    puts splash
-
-    config.vm.provision "shell", keep_color: true, inline: "/srv/www/#{options.website}/#{options.command}"
-    exit()
-  end
 
   # Vagrant Triggers
   #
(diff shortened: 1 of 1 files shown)


Vagrantfile
```diff
@@ -875,3 +875,40 @@ Vagrant.configure("2") do |config|
...
+
+class VVVCommand < Vagrant.plugin(2, :command)
+  def self.synopsis
+    "Execute custom command inside site folder"
+  end
+
+  def execute
+    with_target_vms(nil, single_target: true) do |vm|
+      @env.ui.output "#{$yellow}Executing in #{$red}#{ARGV[1]}#{$yellow}: #{$red}#{ARGV[2]}#{$creset}\n"
+      vm.action(:ssh_run, ssh_run_command: "cd /srv/www/#{ARGV[1]}/public_html; #{ARGV[2]}")
```
msaggiorato: Is there a case for people wanting to run a command above public_html?

Mte90: this is the example of the wordpress-develop template, where you need to execute `grunt watch` on the vm instead of the host for issues as the ticket say.
I think that can be helpful also for plugin development with a switch to the folder in case there are symlinks required inside the VM.

msaggiorato:
I believe we can merge this, right? Just to close some things before the end of the year? Or does this need more testing?

tomjn:
@msaggiorato this one's a bit more complicated, I don't see us merging or closing this in the near future

Mte90:
Can we discuss on slack about why we cannot merge it? I think that is a major feature with very few code and also with very little maintenance required.
I am already using it and I think that can simplify VVV a lot

Vagrantfile
tomjn: This won't work with sites that use the `vm_dir` variable

Vagrantfile
```diff
@@ -875,3 +875,40 @@ Vagrant.configure("2") do |config|
...
+end
+
+class VVVCommand < Vagrant.plugin(2, :command)
+  def self.synopsis
+    "Execute custom command inside site folder"
+  end
+
+  def execute
+    with_target_vms(nil, single_target: true) do |vm|
+      @env.ui.output "#{$yellow}Executing in #{$red}#{ARGV[1]}#{$yellow}: #{$red}#{ARGV[2]}#{$creset}\n"
```
tomjn: What does th `@` do?

Vagrantfile
```diff
@@ -875,3 +875,40 @@ Vagrant.configure("2") do |config|
     trigger.on_error = :continue
   end
 end
+
+class VVVScripts < Vagrant.plugin(2, :command)
```
tomjn: Other than the trigger files what would this be used for? The db backup and restore scripts should be in path, as should xdebug_on etc so `vagrant ssh -c "xdebug_on"` should be enough

Vagrantfile
Mte90: Without `@` doesn't work because we need to access this variable inside the class as I can remember. I am not Ruby's expert but is used also in other part of the vagrantfile.

Vagrantfile
Mte90: This can execute any script uploaded in the homebin, also custom ones.

I think that is a more simple alias compared to that command, but in case we can remove this code.

Mte90:
This code doesn't conflict with vagrant commands because is using the Vagrant api.

Using `vagrant list-commands` print in fact this 2 new commands:
Screenshot_20191220_160523

We can discuss about the script command if it is repetitive but the other one can be handy in a lot of ways.

tomjn:
Once they’re added they’re very difficult to remove. I don’t believe we
should be adding commands via vagrant itself, and I don’t believe these are
the kinds of commands we should add

Mte90:
I think that the script command can simplify a lot because doesn't need to explain what is ssh and how to execute commands inside a VM in vagrant, This can simplify a lot the approach of new users to VVV and open new ideas for the future.

LoreleiAurora:
100% in agreement with tom here. Under no circumstances should this project add commands directly to vagrant. If commands are going to be added they should be within our own tooling.

Mte90:
I don't agree with that choice, is the same to do a bash script that execute `vagrant -c ssh ...` but is integrated with VVV and can get the access to the config.yml content.

I did a Customfile that include those commands https://gist.github.com/Mte90/9810f7c2775f1fd79ee049b3d259dff0 it just require a change to Vagrantfile to expose the `vvv_config` variable as global like this pr also does (line 260).

I am using the vagrant api used by vagrant plugins to add new commands, so I am following the standards and there are no hacks.

Issue #2019, opened 2019-12-05
Site provision - vvv_hosts is not populated as expected


mathieuhays:
I'm using the develop branch, I've pulled the latest changes. The provision used to work before I pulled the latest changes.

## Expected Behavior

Generate nginx config file with the server_name line filled-in correctly.

## Current Behavior

The server_name variable in the nginx config file is empty.

## Your Environment
```
__ __ __ __
\ V\ V\ V / Varying Vagrant Vagrants
 \_/\_/\_/  v3.3.0-git::develop

Platform: platform-darwin13.4.0 shell:/bin/zsh vagrant-hostsupdater vagrant-disksize shared_db_folder_disabled , VVV Path: "/Users/math/Projects/VVV"
Vagrant: v2.2.5, VirtualBox: v5.2.30r130521
```

## Logs

vvv-nginx.conf in my provision folder:
```
server {
    listen       80;
    listen       443 ssl;
    server_name  {vvv_hosts};
    root         {vvv_path_to_site}/public_html;

    # Nginx logs
    error_log    {vvv_path_to_site}/log/error.log;
    access_log   {vvv_path_to_site}/log/access.log;

    # This is needed to set the PHP being used
    set          $upstream {upstream};

    ssl_certificate /vagrant/certificates/www.mathieuhays.co.uk/dev.crt;
    ssl_certificate_key /vagrant/certificates/www.mathieuhays.co.uk/dev.key;

    # Nginx rules for WordPress, rewrite rules, permalinks, etc
    include      /etc/nginx/nginx-wp-common.conf;
}
```

generated config file in `/etc/nginx/custom-sites`
```
server {
    listen       80;
    listen       443 ssl;
    server_name  ;
    root         /srv/www/www.mathieuhays.co.uk/public_html;

    # Nginx logs
    error_log    /srv/www/www.mathieuhays.co.uk/log/error.log;
    access_log   /srv/www/www.mathieuhays.co.uk/log/access.log;

    # This is needed to set the PHP being used
    set          $upstream php;

    ssl_certificate /vagrant/certificates/www.mathieuhays.co.uk/dev.crt;
    ssl_certificate_key /vagrant/certificates/www.mathieuhays.co.uk/dev.key;

    # Nginx rules for WordPress, rewrite rules, permalinks, etc
    include      /etc/nginx/nginx-wp-common.conf;
}
```

site configuration in `config/config.yml`
```
sites:
[...]

PR #2020, opened 2019-12-06
replace the escaping on site names so that they work agan with dots

tomjn:
## Summary:

Switches to the faster/simpler bash based string substitution

## Checks

* [x] I've tested this PR with Vagrant **vXX** and VirtualBox **vXX** on **Operating System**


tomjn:
Fixes #2019

PR #2034, opened 2019-12-17
Fix #2027

### LINKED ISSUE #2027: microk8s MOTD popup on SSH
When I SSH'd into my VM I got this:

Screenshot 2019-12-09 at 16 47 57

I'm sure microk8s is interesting, but it's not VVV, lets get this removed from MOTD

Mte90:
ref #2027 

This remove the motd of the screen, changes tested.

### DIFF
--- a/provision/provision.sh
+++ b/provision/provision.sh
@@ -274,8 +274,8 @@ cleanup_terminal_splash() {
   if [[ -f /etc/update-motd.d/50-landscape-sysinfo ]]; then
     rm /etc/update-motd.d/50-landscape-sysinfo
   fi
-  if [[ -f /etc/update-motd.d/80-live-patch ]]; then
-    rm /etc/update-motd.d/80-live-patch
+  if [[ -f /etc/update-motd.d/80-livepatch ]]; then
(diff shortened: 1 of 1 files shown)


provision/provision.sh
```diff
@@ -274,6 +274,9 @@ cleanup_terminal_splash() {
   if [[ -f /etc/update-motd.d/50-landscape-sysinfo ]]; then
     rm /etc/update-motd.d/50-landscape-sysinfo
   fi
+  if [[ -f /etc/update-motd.d/80-live-patch ]]; then
+    rm /etc/update-motd.d/80-live-patch
```
msaggiorato: For my, without running this specifically i have the `80-livepatch` file. Maybe we should remove that aswell?

Mte90: Well this command will remove that file

msaggiorato: But the command here removes the file named 80-live-patch. I mentioned 80-livepatch.

Mte90: you are right... I didn't saw that. fixed.

PR #2035, opened 2019-12-17
Fix potentially broken nginx_upstream.

### LINKED ISSUE #2028: Validate Nginx Upstream values
At the moment you can specify a different version of PHP like this:

```yaml
nginx_upstream: php73
```

But there's nothing validating that `php73` actually exists, and a user could put in nonsensical values such as:

```yaml
nginx_upstream: banana
```

The end result is a provision failure without any explanation

msaggiorato:
This fixes #2028.

Accepts php or any valid upstream in the /etc/nginx/upstreams directory.

Example output:

image

### DIFF
--- a/provision/provision-site.sh
+++ b/provision/provision-site.sh
@@ -78,7 +78,7 @@ function vvv_provision_site_nginx() {
   sed -i "s#{vvv_hosts}#${VVV_HOSTS}#" "/etc/nginx/custom-sites/${DEST_NGINX_FILE}"
 
   if [ 'php' != "${NGINX_UPSTREAM}" ] && [ ! -f "/etc/nginx/upstreams/${NGINX_UPSTREAM}.conf" ]; then
-    echo -e "${YELLOW} * Upstream ${NGINX_UPSTREAM} not found. Defaulting to php.${CRESET}"
+    echo -e "${YELLOW} * Upstream value '${NGINX_UPSTREAM}' doesn't match a valid upstream. Defaulting to 'php'.${CRESET}"
     NGINX_UPSTREAM='php'
   fi
   sed -i "s#{upstream}#${NGINX_UPSTREAM}#" "/etc/nginx/custom-sites/${DEST_NGINX_FILE}"
--- a/provision/provision-site.sh
+++ b/provision/provision-site.sh
@@ -76,6 +76,11 @@ function vvv_provision_site_nginx() {
   sed -i "s#{vvv_path_to_site}#${VM_DIR}#" "/etc/nginx/custom-sites/${DEST_NGINX_FILE}"
   sed -i "s#{vvv_site_name}#${SITE}#" "/etc/nginx/custom-sites/${DEST_NGINX_FILE}"
   sed -i "s#{vvv_hosts}#${VVV_HOSTS}#" "/etc/nginx/custom-sites/${DEST_NGINX_FILE}"
+
+  if [ 'php' != "${NGINX_UPSTREAM}" ] && [ ! -f "/etc/nginx/upstreams/${NGINX_UPSTREAM}.conf" ]; then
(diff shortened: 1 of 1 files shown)


provision/provision-site.sh
```diff
@@ -76,6 +76,11 @@ function vvv_provision_site_nginx() {
   sed -i "s#{vvv_path_to_site}#${VM_DIR}#" "/etc/nginx/custom-sites/${DEST_NGINX_FILE}"
   sed -i "s#{vvv_site_name}#${SITE}#" "/etc/nginx/custom-sites/${DEST_NGINX_FILE}"
   sed -i "s#{vvv_hosts}#${VVV_HOSTS}#" "/etc/nginx/custom-sites/${DEST_NGINX_FILE}"
+
+  if [ 'php' != "${NGINX_UPSTREAM}" ] && [ ! -f "/etc/nginx/upstreams/${NGINX_UPSTREAM}.conf" ]; then
+    echo -e "${YELLOW} * Upstream ${NGINX_UPSTREAM} not found. Defaulting to php.${CRESET}"
+    NGINX_UPSTREAM='php'
```
tomjn: There is no `php` upstream, instead it should default to `php72` as that's what VVV core installs. We should also print out a list of the valid nginx upstream values

provision/provision-site.sh
```diff
@@ -76,6 +76,11 @@ function vvv_provision_site_nginx() {
   sed -i "s#{vvv_path_to_site}#${VM_DIR}#" "/etc/nginx/custom-sites/${DEST_NGINX_FILE}"
   sed -i "s#{vvv_site_name}#${SITE}#" "/etc/nginx/custom-sites/${DEST_NGINX_FILE}"
   sed -i "s#{vvv_hosts}#${VVV_HOSTS}#" "/etc/nginx/custom-sites/${DEST_NGINX_FILE}"
+
+  if [ 'php' != "${NGINX_UPSTREAM}" ] && [ ! -f "/etc/nginx/upstreams/${NGINX_UPSTREAM}.conf" ]; then
```
tomjn: We don't need the first bit as `php` isn't a valid upstream value, just the final file check for the `.conf` file is enough

tomjn: I think this should do the trick:

```suggestion
  if [ ! -f "/etc/nginx/upstreams/${NGINX_UPSTREAM}.conf" ]; then
```

provision/provision-site.sh
tomjn:
```suggestion
    NGINX_UPSTREAM='php72'
```

provision/provision-site.sh
```diff
@@ -76,6 +76,11 @@ function vvv_provision_site_nginx() {
...
+    echo -e "${YELLOW} * Upstream ${NGINX_UPSTREAM} not found. Defaulting to php.${CRESET}"
```
tomjn:
```suggestion
    echo -e "${YELLOW} * Upstream value '${NGINX_UPSTREAM}' doesn't match a valid upstream. Defaulting to 'php72'.${CRESET}"
```

provision/provision-site.sh
msaggiorato: There is a php upstream. Checked it personally.

https://github.com/Varying-Vagrant-Vagrants/VVV/blob/develop/config/nginx-config/nginx.conf#L88-L91

msaggiorato: It's also the default upstream https://github.com/Varying-Vagrant-Vagrants/VVV/blob/develop/Vagrantfile#L189

provision/provision-site.sh
msaggiorato: I initially had it like this, but covered for php, since that was the default in Vagrantfile. (read other thread for context and links to references)

tomjn: hmmm perhaps we need to double check `php` works as a value

PR #2036, opened 2019-12-20
Improve tideways integration

msaggiorato:
This avoids some warnings being printed before when tideways was not installed. Eg:

image

And also adds specific tideways commands to turn it on or off.

Avoids the problem that was inherent of having tideways installed, that either xdebug or tideways needed to be turned on at any given point.

I've created a new command to check if a utility is installed. It was duplicated across the provisioners, so a command made sense.

### DIFF
--- a/config/homebin/is_utility_installed
+++ b/config/homebin/is_utility_installed
@@ -1,0 +1,10 @@
+#!/bin/bash
+VVV_CONFIG=/vagrant/config.yml
+
+utilities=$(shyaml get-values "utilities.${1}" 2> /dev/null < ${VVV_CONFIG})
+for utility in ${utilities}; do
+if [[ "${utility}" == "${2}" ]]; then
+  exit 0
+fi
+done
+exit 1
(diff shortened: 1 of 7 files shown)


config/homebin/is_utility_installed
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+VVV_CONFIG=/vagrant/config.yml
+
+utilities=$(shyaml get-values "utilities.${1}" 2> /dev/null < ${VVV_CONFIG})
```
tomjn: A micro-optimisation would be to check for the folder since we know the name of the utility, but I doubt that'll save very much time

config/homebin/tideways_off
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+if [[ ! $(is_utility_installed core tideways) ]]; then
+  echo "Tideways is not installed"
+  exit 0
+fi
+
+echo "Enabling tideways_xhprof/xhgui"
```
tomjn:
```suggestion
echo "Disabling tideways_xhprof/xhgui"
```

config/homebin/is_utility_installed
msaggiorato: I'd say it's on the ms range, right?

PR #2042, opened 2020-01-06
Make bash commands available for root

### LINKED ISSUE #2041: is_utility_installed not available inside provisioners
I did a quick update to latest `develop` and reprovisioned, the latest version of the core dev template gave me this in the terminal output:

```
    default: /tmp/vagrant-shell: line 76: is_utility_installed: command not found
```

msaggiorato:
This fixes #2041 .

I've done research on how all the profiles are set up in Ubuntu, and Linux in general.

Here is the information that has been the most helpful for me:

https://unix.stackexchange.com/a/29811

I've also considered things on this answer:
https://askubuntu.com/a/866169

I've decided to move PATH changes, and overall autocomplete initialization to bash_aliases, which is executed in most shell types. Also copying the aliases (not the profile) to the root folder now to run on root.

I've tested this and works with no problem.

### DIFF
--- a/config/bash_aliases
+++ b/config/bash_aliases
@@ -27,10 +27,10 @@ elif [ -d "/srv/www/wordpress-develop/public_html/src/" ]; then
 fi
 
 # add autocomplete for grunt
-eval "$(grunt --completion=bash)"
+[ ! type "grunt" > /dev/null 2>&1 ] && eval "$(grunt --completion=bash)"
 
 # add autocomplete for wp-cli
-. /srv/config/wp-cli/wp-completion.bash
+[ -s "/srv/config/wp-cli/wp-completion.bash" ] && . /srv/config/wp-cli/wp-completion.bash
 
 # PHPCS path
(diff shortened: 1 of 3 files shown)


Mte90:
The code seems good, just to be sure that the code was moved to bash_alias just because this file is used in more places then the bash_profile.
I am wondering if create more issues about the consistency of the file name that in this version includes more stuff and not just only aliases.

msaggiorato:
I've explored where this file gets called from, and it's sourced from the
root profile and the bash_profile just like any other bash file.

I thought this was less problematic than replacing the root user .profile
file which does a lot of other things that we'd have to duplicate, and
potentially maintain across Ubuntu versions.

Also, in the file history I've seen that this file was used up to 2013 in
VVV.

I don't think this causes any issues, but if after testing it does, then
I'd be more than happy to help with any fixes to this bug I've accidentally
introduced.

Mte90:
Yes probably is better to not alter the .profile file for the root user, let's see what @tomjn think about it.

PR #2049, opened 2020-01-20
Update README

achaacha:
A lot of documentation is incorrect and extremely confusing for fresh installers/new users of VM. 
If fresh install downloads VB 6.1.x [Windows], VVV sets default provider to Hyper-V instead of VB, even if the Windows OS is compatible with Hyper-V.

The only way of being aware that VB 6.1.x is incompatible is by adding --provider=virtualbox at the end of vagrant up, to which we will receive the error explaining that it only is compatible with 6.0 and below. There is no documentation of this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,8 +22,8 @@ The online documentation contains more detailed [installation instructions](http
 
 ## Minimum System requirements
 
-* [Vagrant](https://www.vagrantup.com) 2.2.4+
-* [Virtualbox](https://www.virtualbox.org) 5.2+
+* [Vagrant](https://www.vagrantup.com) 2.2.6 or below.
+* [Virtualbox](https://www.virtualbox.org) 6.0.x or below.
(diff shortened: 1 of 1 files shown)


tomjn:
VVV doesn't set the provider to Hyper-V, it goes with Vagrants default, unless you uncomment the provider option in `config.yml`

Also, VB 6.1.x  isn't supported by Vagrant yet as v2.2.7 hasn't been released yet

achaacha:
> VVV doesn't set the provider to Hyper-V, it goes with Vagrants default, unless you uncomment the provider option in `config.yml`
> 
> Also, VB 6.1.x isn't supported by Vagrant yet as v2.2.7 hasn't been released yet

Ah, I see, that makes sense.
I didn't touch the config files with my fresh install, it automatically went with Hyper-V first. But I'm guessing that's because it didn't recognize the VB 6.1.x.

PR #2057, opened 2020-01-29
use the full path to is_utility_installed in the site provisioner

tomjn:
## Summary:

Fixes a problem in the site provisioner

## Checks

* [x] I've tested this PR with Vagrant **v2.2.6** and VirtualBox **v6** on **MacOS**

### DIFF
--- a/provision/provision-site.sh
+++ b/provision/provision-site.sh
@@ -73,7 +73,7 @@ function vvv_provision_site_nginx() {
   fi
   sed -i "s#{upstream}#${NGINX_UPSTREAM}#" "/etc/nginx/custom-sites/${DEST_NGINX_FILE}"
 
-  if [[ $(is_utility_installed core tls-ca) ]]; then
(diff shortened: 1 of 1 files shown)


provision/provision-site.sh
```diff
@@ -73,7 +73,7 @@ function vvv_provision_site_nginx() {
   fi
   sed -i "s#{upstream}#${NGINX_UPSTREAM}#" "/etc/nginx/custom-sites/${DEST_NGINX_FILE}"
 
-  if [[ $(is_utility_installed core tls-ca) ]]; then
+  if [[ $(/srv/config/homebin/is_utility_installed core tls-ca) ]]; then
```
msaggiorato: I would have rathered test this a bit further to get the command to run properly on the provisioners aswell without including the path.

Mte90: In the meantime is patched for who use the develop branch but I think that if we can do some tests an another pr can address this.

PR #2069, opened 2020-02-10
Add php-ldap to provision.sh

olipayne:
## Summary:
Added php-ldap to provision.sh

## Checks
* [✓] I've tested this PR with Vagrant 2.2.7 and VirtualBox 6.0.16 on macOS 10.15.3
* [✓] This PR is for the `develop` branch not the `master` branch.
* [✓] I've updated the changelog.
* [✓] This PR is complete and ready for review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ permalink: /docs/en-US/changelog/
 ### Enhancements
 
 * Improvements to the ruby code in the vagrant file
+* Added php-ldap as standard
 
 ### Bug Fixes
 
--- a/provision/provision.sh
+++ b/provision/provision.sh
@@ -163,6 +163,7 @@ apt_package_install_list=(
   php-ssh2
   php-xdebug
   php-yaml
+  php-ldap
   php7.2-bcmath
   php7.2-curl
   php7.2-gd


tomjn:
What does this do? In the meantime it can be added via a utility, it
doesn’t have to be in the main core provisioner

olipayne:
> What does this do? In the meantime it can be added via a utility, it doesn’t have to be in the main core provisioner

I need php-ldap installed on the box for the plugin I'm developing to work, and I assumed it would be best next to all of the other php-*. Please let me know if it should go elsewhere. I see no harm in it being a part of the whole provision process however.

tomjn:
I’m on a train at the moment, but a utility will do fine for now, look it
up on the website and it’ll show what needs doing

As for the main provisioner, PHP extensions sometimes come with strings
attached, eg pcov, which we added and removed not long ago. Turns out it
can’t be turned on at the same time as tideways or xdebug

olipayne:
Thanks for the nudge in the right direction, anyone who does want php-ldap can use this utility - https://github.com/olipayne/vvv-php-ldap

Mte90:
If you want you can do a pr to https://github.com/Varying-Vagrant-Vagrants/varyingvagrantvagrants.org/blob/ced3f1527a12849e6ec7b0b33da2993cff3538ca/docs/en-US/utilities.md so will be listed in the docs website :-)

PR #2073, opened 2020-02-13
Adjust the syntax of if statements calling the is_utility_installed script

h4r1m4u:
## Summary:

Addresses problem with SSL certificates not being correctly added into Nginx site configs described in issue #2072. 

## Checks

* [x] I've tested this PR with Vagrant **v2.2.7** and VirtualBox **v6.0.16r135674** on **Windows 8.1**
* [x] This PR is for the `develop` branch not the `master` branch.
* [ ] I've updated the changelog.
* [x] This PR is complete and ready for review.

Issue #2072, opened 2020-02-13
SSL certificates not correctly added to Nginx site configs due to if [[ $(/srv/config/homebin/is_utility_installed core tls-ca) ]] always evaluating to false


h4r1m4u:
## What was The Command Used To Provision

```shell
vagrant reload --provision
```

## What Kind of VVV Provision Was This

This was an attempt to upgrade to a new version of VVV

## Logs/What Broke

I've updated to VVV 3.3.0 and have been having issues with the provisioned SSLs on my sites. I'm getting the NET::ERR_CERT_COMMON_NAME_INVALID warning and when I examine the certificates in the browser, they're all issued to vvv.test instead of mysite.test. In other words, my sites are basically using the default SSL and the browser doesn't like the domain mismatch. 

After digging into it a bit, I found out that the cert information is not correctly inserted into the Nginx site configs in /etc/nginx/custom-sites/. After some more digging, I found the block responsible for setting up the certs in `provision/provision-site.sh`:

```shell
  if [[ $(/srv/config/homebin/is_utility_installed core tls-ca) ]]; then
    sed -i "s#{vvv_tls_cert}#ssl_certificate /srv/certificates/${VVV_SITE_NAME}/dev.crt;#" "/etc/nginx/custom-sites/${DEST_NGINX_FILE}"
    sed -i "s#{vvv_tls_key}#ssl_certificate_key /srv/certificates/${VVV_SITE_NAME}/dev.key;#" "/etc/nginx/custom-sites/${DEST_NGINX_FILE}"
  else
    sed -i "s#{vvv_tls_cert}##" "/etc/nginx/custom-sites/${DEST_NGINX_FILE}"
    sed -i "s#{vvv_tls_key}##" "/etc/nginx/custom-sites/${DEST_NGINX_FILE}"
  fi
``` 

My bash knowledge is very limited, so I'm not sure if this is a bug, but it seems that the `if [[ $(/srv/config/homebin/is_utility_installed core tls-ca) ]]` statement always evaluates to false. 

I have tls-ca added in the core utilities in my `config.yml` and if I run `is_utility_installed core tls-ca` from command line in vagrant directly, it correctly exits with 0 status. Likewise, if I run it with some non-existing utility, it exits with 1 status. So the `is_utility_installed` script itself is working correctly. It's only in the if statement in the `provision-site.sh` script where the output always evaluates to false, regardless of whether the utility is actually installed or not. 

My `config/config.yml`:

[...]

PR #2074, opened 2020-02-14
Generate .rnd file if it doesn't exist.

msaggiorato:
This fixes an issue with the latest box (and also a couple of versions before), which didn't have a proper /root/.rnd file which is necessary for generating a certificate.

Seems to address #2072 , right now tls-ca is working for me without the change in syntax proposed in #2073 .

I got this error while provisioning from scratch with 3.2.0 and 3.3.0 with the latest box.

```
 * Sign the certificate using the above private key...
Can't load /root/.rnd into RNG
139738907833408:error:2406F079:random number generator:RAND_load_file:Cannot open file:../crypto/rand/randfile.c:98:Filename=/root/.rnd
```

### DIFF
--- a/provision/provision.sh
+++ b/provision/provision.sh
@@ -614,6 +614,11 @@ tools_install() {
 
 nginx_setup() {
   # Create an SSL key and certificate for HTTPS support.
+  if [[ ! -e /root/.rnd ]]; then
+    echo " * Generating Random Number for cert generation..."
+    vvvgenrnd="$(openssl rand -out /root/.rnd -hex 256 2>&1)"
+    echo "$vvvgenrnd"
+  fi
   if [[ ! -e /etc/nginx/server-2.1.0.key ]]; then
     echo " * Generating Nginx server private key..."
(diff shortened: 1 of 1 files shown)


h4r1m4u:
I tried this and it doesn't fix the #2072 issue for me. The 'if' block in `provision/provision-site.sh` always executes the 'else' branch and the provisioned sites use the default vvv.test certificate instead of their own certs.

Mte90:
I don't get that error on provisioning but I don't think that is a problem to add that check for the machine to avoid issues for others, what do you think @tomjn?

tomjn:
Does this belong in the nginx setup? I don't have a `/root/.rnd` in my VM

tomjn:
Related https://github.com/OpenVPN/easy-rsa/issues/261 and https://security.stackexchange.com/questions/177509/purpose-of-randfile-in-openssl/177512#177512

msaggiorato:
For me, not having this file made the SSL configuration of nginx not work at all. That's why I attempted this.

This is on a fresh provision (after a full destroy, and with the latest box).

PR #2083, opened 2020-03-13
Add a new set_wp_version homebin script

stevegrunwell:
## Summary:

This PR adds a `set_wp_version` script to the VM's `/srv/config/homebin` directory which acts as a wrapper around `svn switch` for the configured WP_TESTS_DIR copy of WordPress (read: `wordpress-trunk` on typical installations).

### Usage

```sh
# Switch to the WordPress 5.3 branch
$ set_wp_version 5.3

# Checkout the 5.2.2 tag
$ set_wp_version 5.2.2

# Reset to WordPress trunk
$ set_wp_version trunk
```

## Checks

* [x] I've tested this PR with Vagrant **v2.2.7** and VirtualBox **v6.1** on **macOS 10.15.3**


Mte90:
My j2c is better to define also in the command name that is for SVN because VVV can use the git repo for the trunk and avoid confusions.

tomjn:
`set_wp_version` implies it can be used to change a sites WP version, which isn't the case.

This looks useful, but it bothers me that it's even necessary, and it's always bothered me that we have a single test WP folder no matter how many copies of core you set up.

tomjn:
hmmm `WP_TESTS_DIR` doesn't point to a WordPress install

https://github.com/Varying-Vagrant-Vagrants/VVV/blob/c1bfe2fc8a1f64e777483cfb215e3e2974c9fd4b/config/bash_aliases#L16-L20

It normally points at a `/srv/www/wordpress-trunk/public_html/tests/phpunit/` folder, not `/srv/www/wordpress-trunk/`, so this script doesn't change the version of WP used

Since this is technically a type of provisioning, wouldn't it make sense to use the provisioner?

tomjn:
I think we need a re-evaluation of exactly what `WP_TESTS_DIR` and its purpose, WP Cores test commands all rely on executing phpunit specific docker stuff

stevegrunwell:
> It normally points at a `/srv/www/wordpress-trunk/public_html/tests/phpunit/` folder, not `/srv/www/wordpress-trunk/`, so this script doesn't change the version of WP used

Yeah, ran into this during testing, so the `cd` line turned into:

```sh
cd "$WP_TESTS_DIR" && cd "$(svn info --show-item wc-root)"
```

Since svn doesn't seem to have a "set the current working directory for the duration of the command" option, we literally go into `/srv/www/wordpress-trunk/public_html/tests/phpunit/`, then go "hey, who's in charge around here?" and `cd` to the root of the SVN repo.

> I think we need a re-evaluation of exactly what `WP_TESTS_DIR` and its purpose, WP Cores test commands all rely on executing phpunit specific docker stuff

Totally fair. Would you like me to close out the PR, or keep it open for discussion? I kept forgetting the command for `svn switch`, so I wrote this out of necessity and thought "hmm, wonder if other people are having trouble testing against non-trunk releases?".

tomjn:
> Totally fair. Would you like me to close out the PR, or keep it open for discussion? 

Lets keep it open, this must be useful or you wouldn't have written it. There's also the option of it going in the `bin` folder of the site template for WP Core dev. I wonder if you could help us and look into it for us and make some recommendations? I loathe having to hardcode stuff, at the moment if you don't have a `wordpress-trunk` or `wordpress-develop` site does that mean no WP unit testing can occur with the official framework?

tomjn:
I'm going to close this as it doesn't make sense as is. Open to a PR on the custom site template develop though

PR #2084, opened 2020-03-14
Generic PHP Debug Mod Switcher

tomjn:
## Summary:

When we tried to add pcov, it wasn't until afterwards I realised it wasn't compatible with XDebug, it seems they all need a particular handler installed, but PHP will only use 1 of them.

To prevent the proliferation of `xdebug_on` `xdebug_off` `tideways_on` `tideways_off` `pcov_on` `pcov_off` etc etc, I've written a switcher script that includes some improvements:

 - You can pass `none` and it will return you to vanilla PHP
 - It will disable everything then enable what you wanted
 - It won't try to disable things that aren't enabled
 - It sets the pipefail etc options
 - Reduces duplication across every script by having just one script

Here's what it looks like in action:
```
vagrant@vvv:~$ switch_php_debugmod xdebug
 * Disabling active debug PHP mods
 * Disabling if present: 'xdebug, xhgui, tideways_xhprof, pcov'
 * Disabling active module: 'xdebug'
 * Enabling 'xdebug'
 * Restarting PHP FPM's so the change takes effect
vagrant@vvv:~$ 
```
Screenshot 2020-03-14 at 16 28 19

There are some future improvements that can be made:

 - It doesn't need to disable the thing that's being enabled, in the above example it disables everything then enables xdebug, but if xdebug is already enabled it should see that and just disable the others
 - I'd like to be able to detect when you're activating something that doesn't exist, at the moment it generates a warning which is a bit unsightly. E.g:
```
vagrant@vvv:~$ switch_php_debugmod bananas
 * Disabling active debug PHP mods
 * Disabling if present: 'xdebug, xhgui, tideways_xhprof, pcov'
 * Disabling active module: 'xdebug'
 * Enabling 'bananas'
WARNING: Module bananas ini file doesn't exist under /etc/php/7.2/mods-available
WARNING: Module bananas ini file doesn't exist under /etc/php/7.2/mods-available
 * Restarting PHP FPM's so the change takes effect
vagrant@vvv:~$ 
```

## Checks

* [x] I've tested this PR with Vagrant **v2.2.7** and VirtualBox **v6.1** on **MacOS Catalina**


Mte90:
I don't know if it is the case to merge also with the support to check if themodule exists.

tomjn:
@Mte90 we don't check for that at the moment, I'm not 100% sure of how we should go about doing that anyway, as `php -m` etc only tell us if it's active. We could look in the mods available folder, but I don't know that's 100% foolproof

PR #2085, opened 2020-03-15
fix #2076

### LINKED ISSUE #2076: Ensure root user has the default credentials in MariaDB
As:

* https://varyingvagrantvagrants.org/docs/en-US/default-credentials/
* https://varyingvagrantvagrants.org/docs/en-US/references/sql-client/

We need to switch to wp/wp after https://mariadb.com/kb/en/authentication-from-mariadb-104/ where root/root is not available on MariaDB

Mte90:
Try to fix https://github.com/Varying-Vagrant-Vagrants/VVV/issues/2076

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ permalink: /docs/en-US/changelog/
 * Improvements to the ruby code in the vagrant file
 * Improved the log folder names from `20200225-182126` to `2020.02.25-18-21-26`
 * Added a `switch_php_debugmod` to replace the `xdebug_on` `tideways_off` style scripts
-* Set always the default password for MySQL root user during provision
+* Checks the default password for MySQL root user during provision
 * Improved the provider examples in `default-config.yml`
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ permalink: /docs/en-US/changelog/
 * Improvements to the ruby code in the vagrant file
 * Improved the log folder names from `20200225-182126` to `2020.02.25-18-21-26`
 * Added a `switch_php_debugmod` to replace the `xdebug_on` `tideways_off` style scripts
+* Set always the default password for MySQL root user during provision
 
 ### Bug Fixes
 
--- a/provision/provision.sh
+++ b/provision/provision.sh
@@ -761,16 +761,13 @@ mysql_setup() {
     cp -f  "/srv/config/mysql-config/root-my.cnf" "/home/vagrant/.my.cnf"
     chmod 0644 "/home/vagrant/.my.cnf"
     echo " * Copied /srv/config/mysql-config/root-my.cnf          to /home/vagrant/.my.cnf"
-    
+
     mysql -u root -proot -e "SHOW DATABASES" 2$> /dev/null
     if [ $? -eq 1 ]; then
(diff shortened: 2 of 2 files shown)


provision/provision.sh
```diff
@@ -761,9 +761,13 @@ mysql_setup() {
     cp -f  "/srv/config/mysql-config/root-my.cnf" "/home/vagrant/.my.cnf"
     chmod 0644 "/home/vagrant/.my.cnf"
     echo " * Copied /srv/config/mysql-config/root-my.cnf          to /home/vagrant/.my.cnf"
-    
-    echo " * Setting the default database password for the root user"
-    mysqladmin -u root password root
+
+    if [[ "ERROR 1045 (28000): Access denied for user 'root'@'localhost' (using password: YES)" != "$(mysql -u root -proot)" ]]; then
```
tomjn: It's enough to check the return code of the command

provision/provision.sh
```diff
@@ -761,9 +761,13 @@ mysql_setup() {
...
+      echo " * Setting the default database password for the root user"
+      mysqladmin -u root password root
+    else
+      echo " * The default database password for the root user is already configured"
```
tomjn: I don't think we need this line, instead we should say that `mysqladmin` was or wasn't succesful in the other conditional

CHANGELOG.md
```diff
@@ -13,8 +13,10 @@ permalink: /docs/en-US/changelog/
 * Improvements to the ruby code in the vagrant file
 * Improved the log folder names from `20200225-182126` to `2020.02.25-18-21-26`
 * Added a `switch_php_debugmod` to replace the `xdebug_on` `tideways_off` style scripts
+* Set always the default password for MySQL root user during provision
```
tomjn:
```suggestion
* Checks the default password for MySQL root user during provision
```

provision/provision.sh
Mte90: Well in this case can detect this error, if there are other errors with the return code we cannot debug easily.

tomjn: If there's an error other than the access denied it could cause issues that are difficult to debug, lets check the return code instead  and print any errors out

provision/provision.sh
Mte90: Behaviour changed let me know for the wording if can be improved, I have not so much fantasy.

Mte90:
it is not yet ready and need testing

Mte90:
With the last code I am getting:
```
default: ERROR
    default:  2002 (HY000)
    default: : Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2)
```
Any hints?

tomjn:
So this PR shuts down MariaDB then restarts it in safe mode, but never undos this after the password has been set :( As a result all the follow up provisioners fail with:


>    default: ERROR 1290 (HY000) at line 1: The MariaDB server is running with the --skip-grant-tables option so it cannot execute this statement

Mte90:
In Italy we say "4 eyes are better then 2", dammit how much time lost for this stupid thing.

tomjn:
You need to start it again after you've stopped it!

Mte90:
It does automatically in the lines after that, there is a check about the mysql status, if is not running automatically start again.

PR #2109, opened 2020-04-09
Fix sudo homebin path

msaggiorato:
I figured out how to remove the requirement for full paths on the provisioners for using utilities under homebin.

I've added a fix at the beginning of the provisioner which makes the paths available to sudo aswell (by adding the homebin path to the secure_path parameter on `/etc/sudoers`)

Also fixed the PATH additions on bash_aliases so that they don't have a trailing slash, and that we check first if the path we're adding is already included or not.

In order to test this, you can SSH in and do the following:

```
is_utility_installed
sudo is_utility_installed
sudo su
is_utility_installed
```

None of the above commands should throw any errors (after a provision of course since the fix is implemented there)

After this is fixed, I can do a separate PR to remove the full path requirement everywhere.

This is related to #2036 and #2042 and several other issues where the homebin commands didn't work properly as expected.

### DIFF
--- a/provision/provision.sh
+++ b/provision/provision.sh
@@ -25,7 +25,15 @@ sed -i -E \
   -e "s|(.*Defaults.*secure_path.*?\".*?)(\")|\1:/srv/config/homebin\2|" \
   /etc/sudoers
 
-# source bash_aliases before anything else so that PATH is properly configured
+# add homebin to the default environment, clean first and then append at the end
+sed -i -E \
+  -e "s|:/srv/config/homebin||" \
+  -e "s|/srv/config/homebin:||" \
+  -e "s|(.*PATH.*?\".*?)(\")|\1:/srv/config/homebin\2|" \
+  /etc/environment
+
+# source bash_aliases before anything else so that PATH is properly configured on
+# this shell session
 . "/srv/config/bash_aliases"
 
 export DEBIAN_FRONTEND=noninteractive
--- a/provision/provision.sh
+++ b/provision/provision.sh
@@ -18,12 +18,14 @@ start_seconds="$(date +%s)"
 # fix no tty warnings in provisioner logs
 sudo sed -i '/tty/!s/mesg n/tty -s \\&\\& mesg n/' /root/.profile
 
-# fix commands from homebin not working with sudo, clean first and then append at the end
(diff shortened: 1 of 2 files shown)


provision/provision.sh
```diff
@@ -18,6 +18,14 @@ start_seconds="$(date +%s)"
 # fix no tty warnings in provisioner logs
 sudo sed -i '/tty/!s/mesg n/tty -s \\&\\& mesg n/' /root/.profile
 
+# fix commands from homebin not working with sudo, clean first and then append at the end
```
tomjn, line 21: What is this command doing? 🤔

Mte90: Quite cryptic indeed, maybe is taking all the aliases and putting that inside the sudoers file?

msaggiorato: The sed is editing a setting for the sudo command, which basically resets the PATH when you do `sudo command`. If you do `sudo su` then you're logged in as root, and the bash_aliases is loaded normally from the root profile. 

But when you run a command with `sudo` (which seems to be the way that scripts are executed by vagrant) then the path used is the one on the `secure_path` setting of the `/etc/sudoers` file. More info on `man sudoers` or https://www.sudo.ws/man/1.8.15/sudoers.man.html

What this command does is check for the presence of the hombin path (at the beggining or anywhere else), remove it, and then append it at the end of that same setting. 

If the setting (for whatever reason) doesn't exist, then the `sed` would do nothing. Also, this sed syntax basically executes 3 expressions on one run (to avoid multiple seds).

The reason I'm then sourcing the bash_aliases into the current shell, is because the change on the sudoers file doesn't apply until you restart the current shell.

Does this make a bit more sense now?

msaggiorato:
I've turned this into a draft, as `vagrant ssh -c 'command'` form to run commands in homebin still doesn't work.

msaggiorato:
Ok, i've improved the comments explaining each command added here, and also now it's possible to run commands like:

```
vagrant ssh -c "xdebug_on"
```

Which were not possible before.

provision/provision.sh
tomjn: Thanks, that's a lot clearer :)

tomjn:
This looks good but I need to test it locally to confirm

Issue #2026, opened 2019-12-09
Hyper-V and hosts... only adds ip and vvv (neither test nor wordpress)


Suzamax:
The expected behavior is to add lines such as:
```
172.17.253.30 vvv.test
172.17.253.30 two.wordpress.test
172.17.253.30 one.wordpress.test
```

But the Vagrant host writes only this:
```
172.17.253.30	vvv	# VAGRANT: 011921ab163d6f804df72b570e7734eb (default) / 42256961-c544-49ab-b2c6-25ed0094720a
```

Have I done something wrong or is it a bug?

tomjn:
Hello! Can we see a copy of your config file? It's impossible to debug this without it, a copy of the provisioner output would also save a lot of time ( it avoids me asking you 10-20 questions if i can just see what it says in the terminal/console )

Suzamax:
Of course! It's almost default.

Also, I had trouble while starting the VVV for the first time (`vagrant up --provider hyperv` and `vagrant up --provision`), I'm going to try another time with another VVV, and see if the problem persists.

Config file below:

```yaml
---
sites:
  wordpress-one:
    skip_provisioning: false
    description: "A standard WP install, useful for building plugins, testing things, etc"
    repo: https://github.com/Varying-Vagrant-Vagrants/custom-site-template.git
    hosts:
      - one.wordpress.test
    custom:
      wpconfig_constants:
        WP_DEBUG: true
        WP_DEBUG_LOG: true
        WP_DISABLE_FATAL_ERROR_HANDLER: true # To disable in WP 5.2 the FER mode

  wordpress-two:
    skip_provisioning: false
    description: "A standard WP install, useful for building plugins, testing things, etc"
    repo: https://github.com/Varying-Vagrant-Vagrants/custom-site-template.git
    custom:
      delete_default_plugins: true
      install_plugins:
        - query-monitor
    hosts:
      - two.wordpress.test

  wordpress-trunk:
    skip_provisioning: true # provisioning this one takes longer, so it's disabled by default
    description: "An svn based WP Core trunk dev setup, useful for contributor days, Trac tickets, patches"
    repo: https://github.com/Varying-Vagrant-Vagrants/custom-site-template-develop.git
    hosts:
      - trunk.wordpress.test

  wordpress-meta-environment:
    skip_provisioning: true # disabled by default, this takes a long time to provision
    description: "An environment useful for contributions to the WordPress meta team."
[...]

PR #2112, opened 2020-04-10
Switch to vagrant-hostmanager

tomjn:
## Summary:

hostsupdater sets our host files on the host OS, but the status of that project is unknown. It looks like the owner might have gone AWOL, and there are Windows issues sitting unfixed, PR's unreviewed, etc.

Given we have Windows users with issues ( see #2026 ), we should look into an alternative, this PR tries to swap hostsupdater for hostmanager, with a fallback to the original.

If you have both installed, hostsupdater will still try to add the `vvv` domain to the hosts file, I could find no way to eliminate this.

PR is currently working locally on an existing VM in MacOS with Virtualbox:

Screenshot 2020-04-10 at 10 50 21

I'd like to see the results of testing this PR on Windows and Linux, in particular from those users with issues

## Checks

* [xv] I've tested this PR with Vagrant **v2.2.7** and VirtualBox **v6.0.14** on **MacOS**

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -373,14 +373,14 @@ Vagrant.configure('2') do |config|
   end
 
   # Auto Download Vagrant plugins, supported from Vagrant 2.2.0
-  unless Vagrant.has_plugin?('vagrant-hostmanager')
-    if File.file?(File.join(vagrant_dir, 'vagrant-hostmanager.gem'))
-      system('vagrant plugin install ' + File.join(vagrant_dir, 'vagrant-hostmanager.gem'))
-      File.delete(File.join(vagrant_dir, 'vagrant-hostmanager.gem'))
-      puts "#{yellow}VVV has completed installing the vagrant-hostmanager plugins. Please run the requested command again.#{creset}"
+  unless Vagrant.has_plugin?('vagrant-hostsupdater')
+    if File.file?(File.join(vagrant_dir, 'vagrant-hostsupdater.gem'))
+      system('vagrant plugin install ' + File.join(vagrant_dir, 'vagrant-hostsupdater.gem'))
+      File.delete(File.join(vagrant_dir, 'vagrant-hostsupdater.gem'))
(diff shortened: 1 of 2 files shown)


Vagrantfile
```diff
@@ -376,14 +373,14 @@ Vagrant.configure('2') do |config|
   end
 
   # Auto Download Vagrant plugins, supported from Vagrant 2.2.0
-  unless Vagrant.has_plugin?('vagrant-hostsupdater')
-    if File.file?(File.join(vagrant_dir, 'vagrant-hostsupdater.gem'))
-      system('vagrant plugin install ' + File.join(vagrant_dir, 'vagrant-hostsupdater.gem'))
-      File.delete(File.join(vagrant_dir, 'vagrant-hostsupdater.gem'))
-      puts "#{yellow}VVV has completed installing local plugins. Please run the requested command again.#{creset}"
+  unless Vagrant.has_plugin?('vagrant-hostmanager')
+    if File.file?(File.join(vagrant_dir, 'vagrant-hostmanager.gem'))
```
Mte90, line 377: This means that we need to update the cd usb generator

tomjn: Indeed, the instructions on the docs site and on .org will need adjusting too

Mte90:
I will test it but maybe a ticket to track that we need to update the doc and cd usb generator so we don't forget to do it.

tomjn:
The main point of this is to improve things for Windows, we need to confirm that's the case

Mte90:
@benlumia007 can you test this pr on windows?

benlumia007:
i was gonna say
```
config.hostmanager.include_offline = false
```
because if you destroy the box, the previous hosts are still available. unelss u guys wanted to set it to true.

benlumia007:
please note after doing some more test, if you do vagrant reload --provision if u have a new site that u added after the first vagrant up. it won't update the hosts file. 

the plugin for vagrant-hostmanager only hooks to vagrant up and vagrant destroy.

PR #2113, opened 2020-04-10
Remove deprecated code from trusty

msaggiorato:
This is throwing an error on every provisioning complaining that /etc/init doesn't exist. This is because upstart existed in trusty (VVV 2) but not anymore in bionic (VVV 3).

### DIFF
--- a/provision/provision.sh
+++ b/provision/provision.sh
@@ -630,10 +630,6 @@ nginx_setup() {
 
   echo " * Setup configuration files..."
 
-  # Used to ensure proper services are started on `vagrant up`
-  echo " * Copying /srv/config/init/vvv-start.conf               to /etc/init/vvv-start.conf"
-  cp -f "/srv/config/init/vvv-start.conf" "/etc/init/vvv-start.conf"
-
   # Copy nginx configuration from local
   echo " * Copying /srv/config/nginx-config/nginx.conf           to /etc/nginx/nginx.conf"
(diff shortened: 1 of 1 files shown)


Mte90:
The file still exists, so we have to remove it, maybe also with the folder itself.

tomjn:
Are we sure we have a sufficient replacement for this on Ubuntu 18? I suspect the only reason these services start is because of the post up script

msaggiorato:
The services start on the post up script for sure, but this haven't worked on bionic (i suspect) from the beggining.

If you look at your provisioning logs you should find:

 ```
* Setup configuration files...
 * Copying /srv/config/init/vvv-start.conf               to /etc/init/vvv-start.conf
 * Copying /srv/config/nginx-config/nginx.conf           to /etc/nginx/nginx.conf
cp: cannot create regular file '/etc/init/vvv-start.conf': No such file or directory
 * Copying /srv/config/nginx-config/nginx-wp-common.conf to /etc/nginx/nginx-wp-common.conf
 * Copying /srv/config/nginx-config/php7.2-upstream.conf to /etc/nginx/upstreams/php72.conf
 * Rsync'ing /srv/config/nginx-config/sites/             to /etc/nginx/custom-sites
```

Even on reprovisionings. This is because the /init folder just doesn't exist on bionic. The only case where we'd need this is if someone was still using a VVV 2 box and provisioning with VVV 3, in which case they have larger problems than this file not existing.

PR #2114, opened 2020-04-10
Improve logging logic and make it reusable across provisioners.

msaggiorato:
I'm doing a bunch of changes to the logging here in an effort to simplify the code of the provisioners.

Each commit message has details about the changes if you need.


tomjn:
I don't mind it being a function then sourcing the file if it helps avoid the full path

msaggiorato:
The point of it being on a file is to avoid the code actually being duplicated across provisioners. 

Another idea is to temporary patch the PATH on provisioners to look into the helpers `/srv/provisioners/helpers`. This would be done at the top of each provisioner, and we could also have other provisioner helpers available as commands and reusable, like noroot, or yml related functions, etc.

tomjn:
I mean do what `provision-network-checks.sh` does, define them all as functions in a file, then just load that file at the top when you want to use it in another script, like in PHP with `include`

Mte90:
So basically the idea on this pr and others that we can do is instead to do script to use as commands, create like libraries with various commands that we can use when we need.

PR #2116, opened 2020-04-13
Improve logging of provisioners, and initial tests environment

msaggiorato:
This is another step towards having unit testing usability for VVV itself.

I'm putting it as draft because this PR currently contains code already in #2114 (and also some further improvements and cleanup)

Posting here so that @tomjn @Mte90 can check this. Still not ready to be merged.


provision/run-provisioners.sh
```diff
@@ -0,0 +1,242 @@
...
+
+get_config_type() {
+  local value=$(shyaml get-type "${1}" 2> /dev/null < ${VVV_CONFIG})
+  echo "${value}"
+}
+
+get_config_keys() {
+  local value=$(shyaml keys "${1}" 2> /dev/null < ${VVV_CONFIG})
+  echo "${value:-$2}"
+}
```
tomjn, line 26: These can go inside the provision helpers

provision/run-provisioners.sh
```diff
@@ -0,0 +1,242 @@
...
+  # copy over version and config files
+  cp -f /home/vagrant/version /vagrant
+  cp -f /srv/config/config.yml "${VVV_CONFIG}"
+
+  sudo chmod 0644 "${VVV_CONFIG}"
+  sudo chmod 0644 /vagrant/version
+  sudo chmod 0644 /vagrant/provisioned_at
+
+  # change ownership for /vagrant folder
+  sudo chown -R vagrant:vagrant /vagrant
```
tomjn, line 84: We may need to create the `vagrant` user, or, create our own user and add it to the necessary groups. If we run this in a docker container or on a host, `vagrant` won't be available as a user/group

Vagrantfile
```diff
@@ -165,22 +165,13 @@ vvv_config['hosts'] = [] unless vvv_config['hosts'].is_a? Hash
 vvv_config['hosts'] += ['vvv.test']
 
 vvv_config['sites'].each do |site, args|
-  if args.is_a? String
```
tomjn, line 168: I don't think we should remove the defaults

msaggiorato: I left only what was needed on the ruby side of things.

provision/run-provisioners.sh
msaggiorato: Got that covered in another branch I have locally, where the user to use is settable on the main provisioner.

msaggiorato:
With this last commit, we have a replacement to `--provision-with="site-wordpress-one"`

Before

`vagrant up --provision-with="site-wordpress-one"`

Now

`VVV_PROVISION="site-wordpress-one" vagrant up --provision`

tomjn:
> With this last commit, we have a replacement to --provision-with="site-wordpress-one"

Problem is people have written scripts which are going to break with this, this would  require us to jump to 4.0

provision/run-provisioners.sh
```diff
@@ -0,0 +1,307 @@
...
+    if [[ $? -ne "0" ]]; then
+      return 0
+    fi
+    bash /srv/provision/provision-utility-source.sh "${name[$i]}" "${repo[$i]}" "${branch[$i]}"
+    PROVISION_SUCCESS=$?
+    provisioner_end "utility-source-${name[$i]}"
+  done
+}
+
+function utility() {
```
tomjn, line 228: We should rename this to `provision_utilities`, and then take the inner loop and replace it with a `provision_utility` function so that we can provision a single utility if we really want to

tomjn:
> Do you have an example of what you mean? I'm not sure i follow.

You  know when you have a class in PHP,  and you declare it in `classname.php`, but right at the very bottom of the file is:

```php
new ClassName();
```

Now it's impossible to test the class, anytime you try to load it a new object gets created :/

Similarly with our bash functions, we want to separate defining functions and using functions.

So for example, we  could have a `provision-all.sh`:

```bash
source "./provisioners.sh"

provisioner_init
pre_hook
main
dashboard
utility_sources
utility
sites
post_hook
```

That way we can run `provision-all.sh` on github actions etc, and we can write tests for those functions.

Then, we could keep all our other provisioners, e.g. `provision.sh` could look something like this:

```bash
source "./provisioners.sh"
main
```

So files can either define provisioners, or run provisioners

Vagrantfile
tomjn: One of the things I was hoping for was to JSON encode the final config option as Vagrant saw it and write it out, either for debugging purposes or just to act as a source of ground truth. I don't mind provisioners checking for this stuff too, that just adds resilience

msaggiorato: Could we leave this to another PR? We could improve the yml parsing logic in bash to print this file, but since it doesn't exist yet, could we do this separately?

Technically this debug data won't be tied to Vagrant/Ruby, so we need to do it on the VM/Container itself.

provision/run-provisioners.sh
msaggiorato: Done

msaggiorato:
I've implemented most of the changes requested (minus some renaming on functions and features desired but not existing until now). 

- `--provision-with` works as before this PR
- provisioner definition, and provisioner runners are separate now.
- still kept the simplification on the vagrantfile

This should be testeable already, right?

Vagrantfile
tomjn: Oh I agree JSON encoding should be another PR, but it wouldn't be possible with these bits removed. We should keep the vagrant based defaults. The bash based stuff should add to it, rather than replace it

Vagrantfile
```diff
@@ -680,91 +644,27 @@ Vagrant.configure('2') do |config|
...
-                          name,
-                          args['repo'].to_s,
-                          args['branch']
-                        ]
+  ARGV.each do |arg|
+    match = arg.match(/(--provision-with=)(.*)/i);
+    if match
+      _prefix, req_prov = match.captures
+      config.vm.provision "#{req_prov}", type: 'shell', keep_color: true, path: File.join('provision', 'run-provisioners.sh'), args: [ req_prov ]
+    end
```
tomjn, line 653: can't we just keep our provisioners defined? It feels weird to do things like this, lets just have standard vagrant provisioners. Otherwise, we're going to need to implement white lists to avoid people doing silly things like `vagrant provison --provision-with="bananas in pyjamas are walking down the stairs"`.

Also it's really handy when you reprovision and it outright tells you what the provisioner names are so you can copy paste them into `--provision-with=`

Vagrantfile
```diff
@@ -680,91 +644,27 @@ Vagrant.configure('2') do |config|
...
   end
 
+  config.vm.provision 'default', type: 'shell', keep_color: true, path: File.join('provision', 'run-provisioners.sh'), args: ( ENV.key?('VVV_PROVISION') ? ENV['VVV_PROVISION'] : [] )
```
tomjn, line 656: We shouldn't use a single file for all provisioners when using `vagrant`. Each provisioner needs to be its own provisioner

msaggiorato: What's the difference between having the VM be responsible for running the provisioners than to have `vagrant` be responsible for it?

Vagrantfile
msaggiorato: --provision-with already fails if you run some random provisioner that doesn't exist. In this case, it would just run nothing.

The idea of this PR is to detach as much as possible from vagrant, and just have it as a tool to create the VM and handle hosts files and shared folders. 

I've moved logic to a single place (the VM and bash) and removed the old ruby logic, if we keep the logic to run provisioners in ruby, and in bash, we have unnecessary code duplication. We have simple workarounds (just like this one) to keep backwards compatibility.

Vagrantfile
tomjn: Because now it's not possible to run just a single provisioner without using a weird trick to manually parse the command line arguments. It's a huge code smell. We should just use normal provisioners and do things how they're supposed to be done. If we have 5 provisioners we should declare 5 provisioners.

As for the VM running provisioners, that's a different thing entirely, *Vagrant* is running the provisioners so we shouldn't mess with that. Having a single file to run the provisioners was meant to be used for testing and docker containers.

provision/tests/provisioners.sh
```diff
@@ -0,0 +1,142 @@
...
+}
+
+function provision_utility() {
+  local group=$1
+  local utility=$2
+  VVV_LOG="utility-${group}-${utility}"
+  bash /srv/provision/provision-utility.sh "${group}" "${utility}"
+}
+
+function sites() {
```
tomjn, line 105: This needs the same change that was made to utilities

So instead of `sites`, we have `provision_sites` and `provision_site` or something to that effect

provision/tests/provisioners.sh
```diff
@@ -0,0 +1,142 @@
...
+  local sites=($(get_config_keys sites))
+  local site
+  for site in ${sites[@]}; do
+    local skip_provisioning=$(get_config_value "sites.${site}.skip_provisioning" "False")
+    if [[ $skip_provisioning == "True" ]]; then
+      continue
+    fi
+    local repo=$(get_config_type "sites.${site}")
+    if [[ "${repo}" == "str" ]]; then
```
tomjn, line 114: I've forgotten where `str` fits in here 🤔 maybe i'm having brainfuzz

provision/tests/provisioners.sh
```diff
@@ -0,0 +1,142 @@
...
+    local branch=$(get_config_value "sites.${site}.branch" "master")
+    local vm_dir=$(get_config_value "sites.${site}.vm_dir" "/srv/www/${site}")
+    local nginx_upstream=$(get_config_value "sites.${site}.nginx_upstream" "php")
+
+    VVV_LOG="site-${site}"
+    bash /srv/provision/provision-site.sh "${site}" "${repo}" "${branch}" "${vm_dir}" "${skip_provisioning}" "${nginx_upstream}"
+  done
+}
+
+function main() {
```
tomjn, line 128: `main_provisioner`? or `provision_main`? hmmm

provision/tests/run-provisioners.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+. "/srv/provision/tests/provisioners.sh"
+
+pre_hook
+main
+dashboard
+utility_sources
+provision_utilities
+sites
```
tomjn, line 10: Lets make sure all of these are prefixed with `provision_`
```suggestion
provision_main
provision_dashboard
provision_utility_sources
provision_utilities
provision_sites
```

provision/tests/provisioners.sh
tomjn:
```suggestion
function provision_main() {
```

provision/tests/provisioners.sh
tomjn:
```suggestion
function provision_sites() {
```

provision/tests/provisioners.sh
```diff
@@ -0,0 +1,142 @@
...
+
+function dashboard() {
+  local dashboard_repo=$(get_config_value "dashboard.repo" "https://github.com/Varying-Vagrant-Vagrants/dashboard.git")
+  local dashboard_branch=$(get_config_value "dashboard.branch" "master")
+
+  VVV_LOG="dashboard"
+  bash /srv/provision/provision-dashboard.sh "${dashboard_repo}" "${dashboard_branch}"
+}
+
+function utility_sources() {
```
tomjn, line 41:
```suggestion
function provision_utility_sources() {
```

provision/tests/provisioners.sh
```diff
@@ -0,0 +1,142 @@
...
+  # run after the shell commands laid out in provision.sh or provision-custom.sh should be
+  # put into this file. This provides a good opportunity to install additional packages
+  # without having to replace the entire default provisioning script.
+  if [[ -f "/srv/provision/provision-post.sh" ]]; then
+    VVV_LOG="post"
+    bash /srv/provision/provision-post.sh
+  fi
+}
+
+function dashboard() {
```
tomjn, line 33:
```suggestion
function provision_dashboard() {
```

provision/tests/provisioners.sh
msaggiorato: This checks for the type of the key, if you have:

```
sites:
  my-site: "https://github.com/something/else"
```

then sites.my-site is a str, not an object, therefore that's the repo. If it's not a str, then you likely have the repo key inside that object.

tomjn: Oooh I didn't see `get_config_type`

tomjn:
Looks good, I need to do some test provisions locally before we can merge

Mte90:
Tested now up/provision/halt with no errors
I have the feeling that is more fast but maybe it's me.

PR #2125, opened 2020-04-19
Fix logs for utility provisioners.

msaggiorato:
This fixes an issue brought up by @tomjn , in which the utility provisioner printed more output than before #2116 .

After researching, I've realized that none of the site provisioners print any output (at all) before #2116 because they were executed as separate commands (in a child process) as opposed to being sourced into the current shell.

For now i'm opening this PR as a draft, but we should consider instead of doing this, to mod the utility provisioners to output things to stdout when needed, so that we still get messages like:

```
 * Copying phpMyAdmin into place
 * Cleaning up after phpMyAdmin
 * phpMyAdmin setup complete
```

Which were not printed (but logged) before #2116 .

I'll do some research and maybe do this PR to vvv-utilities.

### DIFF
--- a/provision/provision-utility.sh
+++ b/provision/provision-utility.sh
@@ -5,7 +5,7 @@
 PROVISIONER="/srv/provision/utilities/${1}/${2}/provision.sh"
 
 if [[ -f $PROVISIONER ]]; then
-    ( source "${PROVISIONER}" )
+    ( bash "${PROVISIONER}" >> "${VVV_CURRENT_LOG_FILE}" )
     SUCCESS=$?
 	if [ "${SUCCESS}" -eq 0 ]; then
 		provisioner_success; exit;
--- a/provision/provision-helpers.sh
+++ b/provision/provision-helpers.sh
@@ -17,6 +17,7 @@ if [[ -f /vagrant/config.yml ]]; then
 fi
 
 export VVV_CONFIG
+export VVV_CURRENT_LOG_FILE=""
 
 function containsElement () {
   declare -a array=(${2})
@@ -138,6 +139,7 @@ function log_to_file() {
 	# pipe to file
 	exec > >(tee -a "${logfile}" )
 	exec 2> >(tee -a "${logfile}" >&2 )
+	VVV_CURRENT_LOG_FILE="${logfile}"
 }
 export -f log_to_file
 


msaggiorato:
I've created https://github.com/Varying-Vagrant-Vagrants/VVV/pull/2126 and https://github.com/Varying-Vagrant-Vagrants/vvv-utilities/pull/64 as a more elegant solution than what we had before #2116 .

If the solution on those 2 PRs is accepted, then I guess this PR can be closed.

I'm leaving things up for discussion.

provision/provision-utility.sh
```diff
@@ -5,7 +5,7 @@
 PROVISIONER="/srv/provision/utilities/${1}/${2}/provision.sh"
 
 if [[ -f $PROVISIONER ]]; then
-    ( source "${PROVISIONER}" )
+    ( bash "${PROVISIONER}" >> "${VVV_CURRENT_LOG_FILE}" )
```
tomjn, line 8: This just pipes everything to the file, that's what `log_to_file` was meant to do, is the issue not that `log_to_file` is not doing what it's supposed to do? What about things that appear on `stderr`? Those are meant to appear in the terminal but wouldn't here

tomjn:
The problem is that `log_to_file` should redirect all stdout to the log file, but print stderr to both the log file and the terminal.

|             | stdout             | stderr             |
|-------------|--------------------|--------------------|
| master/goal |  logfile           | terminal + logfile |
| develop     | terminal + logfile | terminal + logfile |
| this PR     | logfile            | logfile            |

provision/provision-utility.sh
msaggiorato: Let me run some tests, but this is precisely (minus the bash, which was implicit) before #2116 . Check it out here. 

### Before

Log to file

https://github.com/Varying-Vagrant-Vagrants/VVV/blob/f63b32159a949053644589545ae0d0ed9d4133f2/provision/provision-utility.sh#L3-L9

Provisioner execution

https://github.com/Varying-Vagrant-Vagrants/VVV/blob/f63b32159a949053644589545ae0d0ed9d4133f2/provision/provision-utility.sh#L18-L23

Because the provisioners are executable, this will spawn a new bash, and avoid stdout of this script, but append it into the $logfile

### Now on this PR

Log to file logic

https://github.com/Varying-Vagrant-Vagrants/VVV/blob/82f384ba59ced9f0b0c0603adf308445167a74b3/provision/provision-helpers.sh#L131-L142

Provisioner execution

https://github.com/Varying-Vagrant-Vagrants/VVV/blob/82f384ba59ced9f0b0c0603adf308445167a74b3/provision/provision-utility.sh#L8-L10

Like possible to be seen here, i'm just reverting functionality as it was, to avoid outputing information of the actual utility provisioner to the bash, same as before, but still outputing useful information as to timing.

msaggiorato:
Just to make my comment to the [thread above](https://github.com/Varying-Vagrant-Vagrants/VVV/pull/2125#discussion_r411526857) more clear, and also to proof that this PR effectively brings back the old functionality, i've done the following on the tls-ca provisioner log:

image

Which effectively outputs to stderr, and got this results:

Terminal:

image

Logfile

image

So this should be proof that the output is the same as before. 

### Further explanation

Also to clarify on my comment on the [thread](https://github.com/Varying-Vagrant-Vagrants/VVV/pull/2125#discussion_r411526857), the following:

https://github.com/Varying-Vagrant-Vagrants/VVV/blob/82f384ba59ced9f0b0c0603adf308445167a74b3/provision/provision-utility.sh#L8

Pipes stdout to file, but not stderr, as explained [here](http://www.tldp.org/LDP/abs/html/io-redirection.html). Which means stderr would be sent to whatever stderr is set to, which due to log_to_file here:

https://github.com/Varying-Vagrant-Vagrants/VVV/blob/82f384ba59ced9f0b0c0603adf308445167a74b3/provision/provision-helpers.sh#L141

Is piped to the tee command, and then back to the standard stderr (or at least that's what i understand from that).

PR #2126, opened 2020-04-19
Make current log file available to provisioners.

msaggiorato:
This will make the current log file available to subprovisioners. Useful on utilities that may want to hide something from stdout, but not from the logs.

### DIFF
--- a/provision/provision-helpers.sh
+++ b/provision/provision-helpers.sh
@@ -17,6 +17,7 @@ if [[ -f /vagrant/config.yml ]]; then
 fi
 
 export VVV_CONFIG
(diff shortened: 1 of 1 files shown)


tomjn:
Utilities can't output to stdout, all output except stderr gets piped to the log file by default, I don't see how this PR changes that

msaggiorato:
This PR is part of #2125 right now, just posted it here in case we decided to merge the PR i made in Varying-Vagrant-Vagrants/vvv-utilities#64 . I guess we could now close this in favor of whatever we end up doing in #2125 .

Issue #2135, opened 2020-04-26
Installing Packages via config.yml


tomjn:
To save the writing of utilities for things provided by package managers, perhaps we can support ruby/npm/composer packages out of the box, with something like this:

```yaml
packages:
  composer:
    humanmade/coding-standards: *
  gems:
    - wordmove
  npm:
    - sass
```

Related to #2129

Mte90:
I like this idea, the big problem is about conflicts and logs but is something that can be interesting to add.

PR #2138, opened 2020-04-26
add gem binaries to Path if the folder exists

tomjn:
Without this, binaries for installed gems aren't available in path

## Checks

* [ ] I've tested this PR with Vagrant **vXX** and VirtualBox **vXX** on **Operating System**
* [ ] This PR is for the `develop` branch not the `master` branch.

### DIFF
--- a/config/bash_aliases
+++ b/config/bash_aliases
@@ -31,6 +31,14 @@ if [[ $PATH != *"/srv/www/phpcs/scripts"* ]]; then
 	export PATH="$PATH:/srv/www/phpcs/scripts"
 fi
 
+
+# Ruby Gems
+if [ -d "$HOME/.gem/bin" ] ; then
+  if [[ $PATH != *"${HOME}/.gem/bin"* ]]; then
+	  export PATH="$PATH:${HOME}/.gem/bin"
+  fi
+fi
+
 # Vagrant scripts
 if [[ $PATH != *"/srv/config/homebin"* ]]; then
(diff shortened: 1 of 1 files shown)


Mte90:
Any idea how to test it?
I think that we need to install ruby and gem right?

tomjn:
correct, you need to install ruby via snap/linuxbrew, then install a gem, and check it's on PATH

msaggiorato:
Shouldn't this be handled somewhere else? Or have a way to allow utilities to hook on the PATH? Like extra files loaded alongside this on a specific path that utilities can append files to? Similar to how a lot of configurations work in linux, or how the MOTD files are loaded.

That way we can have priorities so that some things happen before others. Seems like an overkill just for this tho, but it would be a nice extensibility option for utilities.

tomjn:
If we add the ability to install ruby gems from the `config.yml` file in the core provisioner, it won't be useful if none of those gems are in PATH, this is a pre-requisite for #2135

msaggiorato:
Oh I see.

But shouldn't we add something to the provisioner to install ruby alongside this? Just to review the feature as a whole I mean.

tomjn:
Only if ruby gems are specified, which means it has to be a part of the PR that implements that issue.

But in the meantime, people using `wordmove` and other gems have to manually install only to find that the binaries aren't in PATH, even if they manually install ruby themselves. So adding it to the bash profile saves them having to fork VVV to get ruby gems working

Issue #2131, opened 2020-04-23
Extend 'skip_provisioning' setting to cover shared folders and databases


shoreline-chrism:
**Is your feature request related to a problem? Please describe.**

Sometimes, I'll use one VVV folder to house a few custom sites in one config.  When I use 'skip_provisioning', I will still see databases imported and folders for sites that I would like to skip.

When I use the 'skip_provisioning' setting in the custom VVV config file, I expect that nothing of that site will be provisioned to the VM.  The nginx rules, hosts and site provisioner are a great start, but is there a limitation with the shared folders and database imports?

**Describe the solution you'd like**

In order to save resources, it would be great if 'skip_provisioning' can also skip sharing the folders and importing the database into the VM. 

**Describe alternatives you've considered**

I usually customize my sites via the vvv-init.sh file but since that's not being run, I can't remove DBs and/or shared folders through that.

Mte90:
This already happens about nginx https://github.com/Varying-Vagrant-Vagrants/VVV/blob/develop/provision/provision-site.sh#L225

The issues is for the db import but I did a pr that fix this https://github.com/Varying-Vagrant-Vagrants/VVV/pull/2142

About the sharing as the whole www folder is shared I don't think that we can fix it because is how works the project. In this way has just few folders to share instead of every. We did that change few months ago to speed up and simplify the booting process.

shoreline-chrism:
Thank you for looking into this.  It's fine that this addresses just the database imports because I believe that is the biggest offender in terms of taking up resources on the VM when you have multiple sites on one VVV instance. 

I will test this branch out shortly but I am not a qualified tester so I won't be running any unit tests or anything. I'm a user of VVV with basic scripting knowledge.

shoreline-chrism:
![skip_provision_works](https://user-images.githubusercontent.com/9121766/81094225-32cfa700-8ed1-11ea-82d3-395dc38dd208.PNG)

I see the warning notes about skipping the import, however the sites were still in my phpmyadmin when I ran `vagrant up --provision`. Here's a screeshot:

![skip_provisioning_still_imported](https://user-images.githubusercontent.com/9121766/81094663-d620bc00-8ed1-11ea-9367-f474cc5c00b0.PNG)

Is there some other place in the provisioning that imports all of the files in the `database/sql/backups` folder? Might the settings `db_backup` and `db_restore` be interfering here?  Both of mine are set to `true`:

```yml
# General VVV options
general:
[...]

PR #2142, opened 2020-05-05
Do not restore the database for a site if the site has skip_provisioning as true

Mte90:
## Summary:

Fix https://github.com/Varying-Vagrant-Vagrants/VVV/issues/2131 about the database side.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ permalink: /docs/en-US/changelog/
 * Always set the database root user password to avoid having the default invalid password on fresh installs
 * Don't spider and recurse domains when checking for network connections
 * Swap the MariaDB apt mirror used for a more reliable source
+* Not import the database if the site has `skip_provisioning` as true
 
 ## 3.3.0 ( 2020 Feb 26th )
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@ permalink: /docs/en-US/changelog/
 * Always set the database root user password to avoid having the default invalid password on fresh installs
 * Don't spider and recurse domains when checking for network connections
 * Swap the MariaDB apt mirror used for a more reliable source
-* Not import the database if the site has `skip_provisioning` as true
+* Do not restore the database for a site if the site has `skip_provisioning` as true
 
 ## 3.3.0 ( 2020 Feb 26th )
 
--- a/database/sql/import-sql.sh
+++ b/database/sql/import-sql.sh
@@ -47,10 +47,18 @@ then
 	# get rid of the extension
 	pre_dot=${file%%.sql}
 	# get rid of the ./
-  db_name=${pre_dot##./}
-	if [[ True == $(get_config_value "sites.${db_name}.skip_provisioning" "False") ]]; then
-		vvv_warn " * Skipping importing ${db_name} database"
-		continue
+	db_name=${pre_dot##./}
+	db_configured=$(get_config_value "sites.${db_name}.db_name")
+	if [[ "False" == $db_configured ]]; then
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ permalink: /docs/en-US/changelog/
 * Always set the database root user password to avoid having the default invalid password on fresh installs
 * Don't spider and recurse domains when checking for network connections
 * Swap the MariaDB apt mirror used for a more reliable source
+* Not import the database if the site has `skip_provisioning` as true
```
msaggiorato, line 30: I'd say that the changelog would be "Do not restore the database for a site if the site has `skip_provisioning` as true"

database/sql/import-sql.sh
```diff
@@ -48,6 +48,10 @@ then
 	pre_dot=${file%%.sql}
 	# get rid of the ./
   db_name=${pre_dot##./}
+	if [[ True == $(get_config_value "sites.${db_name}.skip_provisioning" "False") ]]; then
```
msaggiorato, line 51: Wouldn't True need to be on quotes (since it's a string)?

tomjn:
I don't believe it's possible to implement this, and I question wether we should. It assumes that a database is created for each site, and only that database. But that assumption isn't necessarily true

database/sql/import-sql.sh
Mte90: I copied that code from another provisioner.

Mte90:
I have like 20 database backups in VVV of older webistes that not exists anymore, with this change I can confirm that is more fast the provisiong.
Also that some templates like the meta one that create like 4 db are not reconized, so they are imported anyway.

Looking at the original template there is also the `db_name` https://github.com/Varying-Vagrant-Vagrants/custom-site-template but this code doesn't use that value. I can add a check for this as first otherwise use the site name as fallback, so people with issue can change it.

benlumia007:
i think the skip_provisioning also causing sites to create certifications if the skip is true.

msaggiorato:
What happens if a database name doesn't match any entry in the config.yml file? Would it be imported?

This will still allow the flow that @tomjn mentions of people potentially having 2 databases related to a site.

Mte90:
Not imported, there is the `db_name` parameter, if it is defined it will be used that so if someone use a different db is better that configure it in the config file.

msaggiorato:
But db_name currently doesn't allow a list, it's just a string right? So the use case that @tomjn mentioned which is usually possible with the current "dirty" implementation is not possible on the new "clean" one.

We need to somehow accomodate for this use case, but still avoid restoring things that should not be restored. See what I mean?

Mte90:
Well if a template install and create more database they will be ignored on this process so they will be imported.

tomjn:
The more I think about this, the more I think that we should provide general database back up options rather than site specific ones. Otherwise we'd be better off giving every site their own user+pass and set of databases which is a lot of work and doesn't help existing users

I'd suggest `config.yml` have options for:


 - back up specific database
 - skip specific database
 - default database backup value ( e.g. if it's not in the above two options what should we do?

Then again for restore, under a new `db_backup:` section.

E.g. something like this that covers all bases:

```yaml
general:
  db_backup:
    default: true
    include:
      - firstdb
      - second
    exclude:
      - fourth
      - fifth
  db_restore:
    default: true
    include:
      - firstdb
      - second
    exclude:
      - fourth
      - fifth
```

There's also the question of changing our default backup behaviour, e.g. only backup on `vagrant destroy`, and present manual restore options for each database.

Mte90:
Closed for https://github.com/Varying-Vagrant-Vagrants/VVV/issues/2225

PR #2151, opened 2020-05-15
fix #2150 dpkg frontend lock

### LINKED ISSUE #2150: Unable to acquire the dpkg frontend lock
## What was The Command Used To Provision

vagrant up --provision

## What Kind of VVV Provision Was This

This was a reprovision of an already working VVV

1.   The issue seems to happen each morning, after a fresh boot of the system (iMac)
2.  If the system has been running for 30mins+ and I run a Vagrant Up, it usually works fine.
3.  Have previously tried a clear-out of VVV, Vagrant & VB & re-install (around time of 3.2.0 relase)
4. 

Linked above

## Your Environment

Linked above

Mte90:
## Summary:
fix https://github.com/Varying-Vagrant-Vagrants/VVV/issues/2150

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@ permalink: /docs/en-US/changelog/
 * Always set the database root user password to avoid having the default invalid password on fresh installs
 * Don't spider and recurse domains when checking for network connections
 * Swap the MariaDB apt mirror used for a more reliable source
-* Fix issues with apt locking file at every provision
+* Fixed an issue with the dpkg lock file not being cleaned up sometimes
 
 ## 3.3.0 ( 2020 Feb 26th )
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ permalink: /docs/en-US/changelog/
 * Always set the database root user password to avoid having the default invalid password on fresh installs
 * Don't spider and recurse domains when checking for network connections
 * Swap the MariaDB apt mirror used for a more reliable source
+* Fix issues with apt locking file at every provision
 
 ## 3.3.0 ( 2020 Feb 26th )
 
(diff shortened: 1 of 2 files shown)


msaggiorato:
What's the condition that #2150 happens on? Seems like a failed provisioner right? Or a machine that was force halted during the install process.

If that is the case, should we be removing the locks? Or forcing the user to force recreate the machine, just to avoid any potential problems with failed provisions.

Mte90:
I think that is a provisioner fail previously but is the only way to fix it is to remove the file. On my debian system I do the same.

macgraphic:
I added the line to provision.sh
`  # fix https://github.com/Varying-Vagrant-Vagrants/VVV/issues/2150
  rm /var/lib/dpkg/lock*`

But I am still getting the same error on a 'vagrant up --provision'
`    default:  * Installing apt-get packages...
    default: E: Could not get lock /var/lib/dpkg/lock-frontend - open (11: Resource temporarily unavailable)
    default: E: Unable to acquire the dpkg frontend lock (/var/lib/dpkg/lock-frontend), is another process using it?
    default:  * Installing apt-get packages returned a failure code, cleaning up apt caches then exiting`

To correct or to start VVV, I then run `vagrant reload --provision`

Mte90:
I tested now the pr (because I had that problem) and worked.
Did you added that line to the right line? Are you using the develop branch or switched to the pr one?

tomjn:
Is there an `apt` or `dpkg` based way of clearing it? Removing the file works but if there's a better way, similar to how you can run `svn cleanup` then that might help avoid any strange issues encountered

tomjn:
Also can we get a changelog update?

Mte90:
Changelog updated and no there isn't any way.
Remove that files is the only way to fix it with apt (debian user talking), the lock files are crate by linux processes but they are not removed if the process itself is killed (like libreoffice or vim) so you have to remove it manually.

PR #2157, opened 2020-05-18
Enable gzip compression

paulschreiber:
## Summary:
This was disabled in 2013. A quick test shows it works fine.

Previously:
https://github.com/Varying-Vagrant-Vagrants/VVV/commits/52944454


tomjn:
I'm having to tether for internet for a day or two, I won't be able to test this out until after Wednesday

tomjn:
I was concerned the underlying issue hadn't been fixed, and I remember seeing it mentioned elsewhere, but it seems all traces are gone. I'm going to merge this, and if someone has issues we can undo it and gather more info

PR #2158, opened 2020-05-18
Expire static assets after 100 days

paulschreiber:
## Summary:
Adds an `Expires:` header for static assets.

Primary purpose is to silence the error in Lighthouse performance audits.


Mte90:
Well in development can make sense to serve in this way.

tomjn:
hmm is this something we want on all sites regardless? Would it make more sense for this to be configurable? It can already be done via a `vvv-nginx-custom.conf` file via the site template

Mte90:
What are the problems of implementing this? As today browser have an option to disable caching in the developer tools if someone needs but can break the workflow for others.

paulschreiber:
I'm not sure what you mean by _what are the problems of implementing this_?

tomjn:
I can foresee people complaining their assets won't refresh because of this. Keep in mind Lightroom is intended for live production sites, but VVV is serving a local dev build.

People might also disagree on 100 days and want to match their own production servers, or differ on days for formats. By putting this in nginx WP common, we've forced all VVV users to conform to 100 days for those formats regardless of what they configure for themselves

tomjn:
Instead, perhaps this would have been better being made to the custom site template? In an ideal world the only things in nginx wp common are the PHP fast fpm parts. But if we move things out into custom site template, that'll break things for people with custom nginx configs who are now missing important things

paulschreiber:
I picked 100 days because that appears to be the minimum threshold for Lighthouse to stop showing warnings. Setting it to 90 days still shows a warning.

Happy to have you move this to another file/template.

tomjn:
@paulschreiber I'd put it in here:

https://github.com/Varying-Vagrant-Vagrants/custom-site-template/blob/master/provision/vvv-nginx-default.conf

PR #2160, opened 2020-05-21
Update changelog, linking to PRs of each change.

msaggiorato:
This should make clear the changes we've introduced.

I think i didn't miss anything. I know this is a change in how we've been structuring the changelog, but i think it's useful to link to the PRs of each change to more easily track changes.

Let me know if this is useful.


tomjn:
@msaggiorato not sure how we have conflicts but can you resolve? I suspect it's because lines were added to the top rather than the bottom of the lists and changes have been merged since

PR #2174, opened 2020-06-02
Avoid stdout on all provisioners (except main)

### LINKED ISSUE #2173: Main Site Provisioner Outputs All Output not just Errors
## Expected Behavior

When provisioning, only errors are shown during provisioning. Normal output appears in the log file alongside errors.

## Current Behavior

When provisioning, all output is shown in the terminal, wether it's an error or not

## Possible Solution

Unsure what's causing this at the moment, will need investigation

msaggiorato:
This is an attempt to fix #2173 and also clarify a bit of the process to pipe data to a file.

While researching this, i noticed there's a bug in the current implementation which is kind of a pain in the ass to fix, but was not introduced by my refactor.

This basically is a concurrency issue with tee. The way it currently works is that each output (stdout and stderr) are piped to FIFO queues, which is later fed to tee. The script only waits for the queue to be written, but not until tee finishes writing (appending actually) to the file. This is the reason that in some cases the log lines might be out of order, or in a weird situation. Link to an explanation of this [here](https://stackoverflow.com/questions/58035279/why-does-exectee-output-display-out-of-order/58036027#58036027).

I've tried addressing this, but didn't have time to fully fix this, and I think this may be a bit out of the scope of the issue i'm trying to fix (yet it's an issue with the tee piping technique implemented)

After research, the code before my refactor behaved the same as it did before this PR. This PR effectively behaves like @tomjn suggested it should.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ permalink: /docs/en-US/changelog/
 * Swap the MariaDB apt mirror used for a more reliable source ( partially #2140 and in a217369 )
 * Fixed an issue with the dpkg lock file not being cleaned up sometimes ( #2151 )
 * Fix issues with the sad bear showing at the end of provisioning despite provisioners being succesful ( #2161 )
+* Fix provisioners printing all output to console (not just errors) ( #2174 )
 
 ## 3.3.0 ( 2020 Feb 26th )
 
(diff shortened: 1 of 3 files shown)


tomjn:
I'll test this later today, do you have any links for research on the `tee` issue?

msaggiorato:
Added link to the PR description.

Issue #2178, opened 2020-06-08
PHP intl extension is not installed


paulschreiber:
## Expected Behavior
PHP `intl` extension is installed.

## Current Behavior
PHP `intl` extension is not installed.

## Your Environment
VVV 3.4.0 (develop branch)

Mte90:
The fun part is that I had a bug in one of my plugin in case there was this module. As I was using the fallback the problem wasn't appearing.
If the pr was a week before maybe I had spotted first 😂

PR #2179, opened 2020-06-08
install php7.2-intl extension

paulschreiber:
## Summary:

Install PHP 7.2 intl extension

## Checks

Partial fix for #2178.


tomjn:
This will need doing on the utility repo too, these changes will only impact VVV core when no additional versions have been installed

tomjn:
Also the changelog needs to be updated

Issue #2177, opened 2020-06-06
Why is NVM removed on provision if it exists?


coolamit:
I was looking to update the provision script to install NVM if it does not exist and then I came across this snippet in `provision/provision.sh`

```bash
  if [[ -f ~/.nvm ]]; then
    echo " * .nvm folder found, switching to system node, and removing NVM folders"
    nvm use system
    rm -rf ~/.nvm ~/.npm ~/.bower /srv/config/nvm
    echo " * NVM folders removed"
  fi
```

It seems that if anyone installs NVM manually via terminal after provisioning VVV, it will be removed next time VVV is re-provisioned which is bound to happen if a new domain/site is to be added or existing one is to be removed.

Seems kinda odd/weird.

Any idea why this was done?

I find NVM a useful tool as different projects could have different Node.js requirements based on when they were set up and upgrading to latest version right away might not be feasible for them for any number of reasons including lack of time, incompatibility of some sorts, etc. In such cases NVM allows to have more than one version of Node.js available which can be switched from project to project as per requirements.

I can submit a PR for `provision/provision.sh` to install NVM if this is something you guys are ok with having in VVV. In my opinion VVV should have NVM.

tomjn:
We used to have nvm but removed it, this code tries to clean up after old
versions. Happy for it to be removed or adjusted.

Perhaps it only runs if nvm isn’t installed?

Mte90:
Maybe this is part of https://github.com/Varying-Vagrant-Vagrants/VVV/issues/2135 ?

tomjn:
Also happy for nvm to be added to the core utilities though I can’t
maintain it myself so it’ll need clear test instructions

coolamit:
> Perhaps it only runs if nvm isn’t installed?

It runs if NVM is installed because it looks for install directory which by default is created at `~/.nvm` and that code removes it if its found.

Mte90:
One idea can be create a flag in config.yml that disable the removal of old stuff like this one.
Our provision has a lot of checks to remove tools and files of old versions, can be helpful also for our plans of testing include something similar.

coolamit:
> Maybe this is part of #2135 ?

If it were to be, I think it would be more suited as part of utilities than packages. However that is a pretty neat idea. :)

> Also happy for nvm to be added to the core utilities

[...]

PR #2185, opened 2020-06-09
Added NVM utility

coolamit:
This adds `nvm` as a core utility in default config. Its disabled by default and has to be enabled in config prior to provisioning VVV.

This PR also removes the code from `provision/provision.sh` which looks for and deletes `nvm` folder.

Ref: #2177 
Ref: https://github.com/Varying-Vagrant-Vagrants/vvv-utilities/pull/67

### DIFF
--- a/provision/provision.sh
+++ b/provision/provision.sh
@@ -452,14 +452,6 @@ tools_install() {
   # Disable xdebug before any composer provisioning.
   sh /srv/config/homebin/xdebug_off
 
-  echo " * Checking for NVM"
-  if [[ -f ~/.nvm ]]; then
-    echo " * .nvm folder found, switching to system node, and removing NVM folders"
-    nvm use system
-    rm -rf ~/.nvm ~/.npm ~/.bower /srv/config/nvm
-    echo " * NVM folders removed"
-  fi
-
   if [[ $(nodejs -v | sed -ne 's/[^0-9]*\(\([0-9]\.\)\{0,4\}[0-9][^.]\).*/\1/p') != '10' ]]; then
     echo " * Downgrading to Node v10."
     apt remove nodejs -y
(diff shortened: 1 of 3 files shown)


Mte90:
I think that we should keep the compatibility for old version on removing NVM.
Maybe we can add a check to execute this code if the NVM utility is not installed, this code should achieve that:

```
if is_utility_installed core nvm; then
```

provision/provision.sh
```diff
@@ -452,14 +452,6 @@ tools_install() {
   # Disable xdebug before any composer provisioning.
   sh /srv/config/homebin/xdebug_off
 
-  echo " * Checking for NVM"
-  if [[ -f ~/.nvm ]]; then
```
Mte90, line 456: Wrap this with `if is_utility_installed core nvm; then`

coolamit: You mean this

```bash
echo " * Checking for NVM"
  if [[ ! is_utility_installed core nvm && -f ~/.nvm ]]; then
    echo " * .nvm folder found, switching to system node, and removing NVM folders"
    nvm use system
    rm -rf ~/.nvm ~/.npm ~/.bower /srv/config/nvm
    echo " * NVM folders removed"
fi
```

tomjn: We could, but I'd rather implement an official method of removal for utilities than add cleanup code for every utility to core. We should merge this as is

coolamit: @tomjn In that case maybe have every utility add a `uninstall.sh` which could be run to remove that utility? Because every utility would have different un-installation since installation isn't standard for all of them.

tomjn:
That’d be the idea but let’s not have that block this. We can just remove
the cleanup code here and update the change log then focus on the utility PR

coolamit:
@tomjn Changelog updated

Issue #2139, opened 2020-04-27
Minimize the Contributor day box


tomjn:
Our box building script minimizes the size of the vagrant box by first zeroing out free space and removing unnecessary artefacts, significantly reducing the size of the final file. Our contributor day USB generator however doesn't do this before generating the box file though.

We should fix that!

See:

https://github.com/Varying-Vagrant-Vagrants/BoxBuilder/blob/master/ubuntu-18.04/scripts/minimize.sh

and

https://github.com/Varying-Vagrant-Vagrants/BoxBuilder/blob/master/ubuntu-18.04/scripts/cleanup.sh

PR #2200, opened 2020-06-19
Add/box scripts

tomjn:
## Summary:

Adds the 2 box minimizing scripts to the homebin folder. This should assist with https://github.com/Varying-Vagrant-Vagrants/VVV/issues/2139

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ permalink: /docs/en-US/changelog/
 
 * Cleaned up leftover `nvm` removal code from main provisioner ( #2185 )
 * Added support for `vagrant-goodhosts`, we recommend using this in the future instead of `vagrant-hostsupdater`
+* Added `box-cleanup.sh` and `box-minimize.sh` scripts. Run these before creating a vagrant box to reduce disk size. These are only intended for box file creation.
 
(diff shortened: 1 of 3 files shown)


Mte90:
Documentation for this?

tomjn:
This isn’t something I want end users to use, these are the vagrant box
minimising scripts aka magically make the final file smaller. They do what
they say on the tin

msaggiorato:
If people shouldn't be using this, shouldn't this be on the USB Generator somehow?

tomjn:
> If people shouldn't be using this, shouldn't this be on the USB Generator somehow?

sure but then what if I wanted to make a box for other reasons? Now it's here, and it's in the USB generator, and it's in the box builder repo, so many duplicated copies all over the place. You get a copy of the script, and you get a copy of the script, and every repo gets a copy of the script, and so many copies...

Or we can put it here and have 1 place for the script.

**These scripts make adjustments to the VM so that when it's boxed up it makes a smaller file.**

The messages are for people who might think it's for other things, like making their VM smaller, or they might run these scripts because they want to keep their VM clean. All of which are bad ideas. That's why these scripts won't be documented on the docs site. The only docs they need are the changelog, otherwise these are for internal use.

PR #2201, opened 2020-06-19
shyaml has a "default" parameter on the command, use it

msaggiorato:
The idea of handling defaults is to handle empty strings, but if it's specifically set to empty, our old way of reading specific values fail

[Here's](https://github.com/0k/shyaml#handling-missing-paths) docs related to what I used

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ permalink: /docs/en-US/changelog/
 ### Bug Fixes
 
 * Fix mysql root password reset ( #2182 )
+* Fix empty string yml value reading on site provisioner ( #2201 )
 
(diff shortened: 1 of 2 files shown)


tomjn:
Does this work though? Shyamls default values aren't as straight forward as they appear, especially where multiple values are concerned, there are various situations were a default value can be passed, but a default value is not usable

msaggiorato:
Hey Tom, I see you approved and merged this after asking if it was reliable, did you test those edge cases you mentioned?

From what i've seen the results should be the same, just more reliable since by using shyaml we can return the default only if a key is undefined. But if it's empty we should not return the default, and return the empty value.

tomjn:
I think it was that the function already took a default but didn’t do
anything with it so I merged

PR #2206, opened 2020-07-05
Enable the VM hostname to be set via config.yml

stevegrunwell:
## Summary:

Enable the VM's hostname to be set via the `config.yml` file, useful when a developer might have multiple instances of VVV on a single host machine.

This change technically affects two settings within the VM:

1. The hostname (`/etc/hostname`) within the virtual machine
2. The name assigned to the VM within Vagrant (currently, VVV has been defaulting to "default")

      ```sh
      $ vagrant global-status
      id       name             provider   state    directory
      -------------------------------------------------------------------------
      f0fb7bf  my-vvv-hostname  virtualbox running  /Users/steve/Vagrant/VVV
      ```

## Checks

* [x] I've tested this PR with Vagrant **v2.2.9** and VirtualBox **v6.1** on **macOS 10.15**

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -437,7 +437,14 @@ Vagrant.configure('2') do |config|
     end
   end
 
-  config.vm.hostname = 'vvv'
+  # Enable hostnames to be set via vm_config.hostname, but default to "vvv".
+  if defined? vvv_config['vm_config']['hostname'] then
+    config.vm.hostname = vvv_config['vm_config']['hostname']
+    config.vm.define vvv_config['vm_config']['hostname']
+  else
+    config.vm.hostname = 'vvv'
+    config.vm.define 'vvv'
+  end
 
   # Specify disk size
   #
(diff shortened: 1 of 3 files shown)


tomjn:
Are you sure that this won't cause all existing users to create brand new VMs on their next provision? Keep in mind as well that hostsupdater will add this value to the hosts file ( it's something it does wether you like it ti or not :( )

Vagrantfile
```diff
@@ -437,7 +437,14 @@ Vagrant.configure('2') do |config|
     end
   end
 
-  config.vm.hostname = 'vvv'
+  # Enable hostnames to be set via vm_config.hostname, but default to "vvv".
+  if defined? vvv_config['vm_config']['hostname'] then
```
tomjn, line 441: This should be set using a default value further up the file so it's always defined

tomjn:
hmmm, so if the user has an existing VM and changes this setting, they'll have a brand new VM created the next time they `vagrant up`, and they'll get 2 VMs in the same folder, with their old VM innaccessible until they change the option back.

This is why we didn't rename the box to fix the vagrant output saying `default:` instead of `vvv:` as it'd cause this problem.

For this PR to be merged, there needs to be a way to avoid this problem, either by exiting if a VM already exists with a different name, or some other way of handling it that isn't destructive or causes problems.

Mte90:
Maybe a more simple solution can be to not let an already VVV provisioned machine to change the name.

Mte90:
@stevegrunwell we need your opinion to move further this pr.

tomjn:
To move this forward:

 - we should rename it from hostname to `name`, the hostname is just a side effect of the VMs name
 - a way too only apply this if the current VM does not exist to avoid a particularly disastrous scenario that could generate multiple VMs in the same folder that cannot be accessed

tomjn:
Closing this as it breaks VMs that have already been created

PR #2216, opened 2020-07-15
Prevent use of sudo vagrant up, #2215

### LINKED ISSUE #2215: Prevent use of sudo vagrant up
Some users are typing `sudo vagrant up`, this is very bad! Once this is done it's possible to have 2 VMs occupying the same folder, and it leads to a non-trivial cleanup process.

We need to prevent this from happening, no VVV user should be using `sudo` to start the VM. VVV should check super early in the vagrantfile if the user is running as `root` on non-Windows environments, and exit early if that's the case, displaying an appropriate message indicating the seriousness of this mistake, and linking to this issue.

Mte90:
## Summary:

fix #2215, basically stop the execution of vagrant with an alert if the user is root or is running with root permission.

Require a bit of testing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ permalink: /docs/en-US/changelog/
 * Added support for `vagrant-goodhosts`, we recommend using this in the future instead of `vagrant-hostsupdater`
 * Added `box-cleanup.sh` and `box-minimize.sh` scripts. Run these before creating a vagrant box to reduce disk size. These are only intended for box file creation.
 * Fixed an issue preventing backups of databases whose names contained reserved words ( #2213 )
+* Prevent use of sudo vagrant up ( #2215 )
 
 ### Bug Fixes
 
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -123,6 +123,12 @@ if show_logo
   puts splashfirst
 end
 
+case (Process.uid)
+when 0
+  puts "#{red}VVV (and Vagrant) cannot be executed as root user or with sudo permissions!#{creset}\n\n"
+  exit
+end
+  
 # Load the config file before the second section of the splash screen
 
 # Perform file migrations from older versions
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -125,8 +125,16 @@ end
 
 case (Process.uid)
 when 0
-  puts "#{red}VVV (and Vagrant) cannot be executed as root user or with sudo permissions!#{creset}\n\n"
-  exit
+	puts "#{red}     ! DANGER  !"
+	puts "#{red} ! ▄▀▀▀▄▄▄▄▄▄▄▀▀▀▄ !  You should never use sudo or root with vagrant.#{creset}"
+	puts "#{red}  !█▒▒░░░░░░░░░▒▒█    It causes lots of problems :(#{creset}"
+	puts "#{red}!   █░░█░▄▄░░█░░█ !   #{creset}"
+	puts "#{red}     █░░█░░█░▄▄█    ! We're really sorry but you may need to do painful#{creset}"
+	puts "#{red}  !  ▀▄░█░░██░░█      cleanup commands to fix this.#{creset}"
+	puts " "
+	puts "#{red}If vagrant does not work for you without sudo, open a GitHub issue instead#{creset}"
+	puts "#{red}In the future, this warning will halt provisioning to prevent new users making this mistake.#{creset}"
+  # exit
 end
   
 # Load the config file before the second section of the splash screen


tomjn:
I think there's 3 things that need to happen before merge:

1. confirm it does what it needs to on Mac and linux with `root` and as `sudo`
2. Replace the text with a teddy bear andd a bigger message that links here
3. I wonder if we can provide a bypass for users to get a way to recover their data to convert back to non-sudo? Like `sudo vagrant up --i-need-to-recover-data-to-stop-using-sudo`, so that we don't end up with users in a stuck state with a `sudo` provisined VM they can't get rid of

Mte90:
On Linux I got an error that say that I cannot use vagrant as root user. It was localized in italian so I think that was by vagrant itself.
Maybe on OSX is different?

tomjn:
hmm that's true:

```
❯ sudo vagrant status
The VirtualBox VM was created with a user that doesn't match the
current user running Vagrant. VirtualBox requires that the same user
be used to manage the VM that was created. Please re-run Vagrant with
that user. This is not a Vagrant issue.

The UID used to create the VM was: 501
Your UID is: 0
```

I'm not sure how some users are able to use `sudo` but it does happen. At least we have code for it now.

I had some more thoughts, we can get this merged with just item 3, then follow up with another PR for enhancements

Mte90:
About point 3 I don't think that is easy. This should do a chown/chmod to all the files, breaking git and all the others subdirectories like in VVV and also change the permission of the VM that are in another folder.

tomjn:
All the things the provisioners provision have the same owner and chmod that we specify in the vagrant file, the VM knows no better and can't because of ho.w virtualbox works.

Otherwise, `root` owns those files, and will already own them when `sudo vagrant up` returns out error message. If we tryt o auto-fix this it won't work, as we need to know who to change ownership to, and, we would need too cleanup things outside the VVV folder, there is no automated fix for it.

The problem then becomes, how do I as a user that used `sudo vagrant up` in the past recover my database so I can perform cleanup without data loss? We can't tell this user that they can't use `sudo` because they need it to retrieve their data and run `sudo vagrant destroy` to begin the cleanup process. But they can't run `vagrant up` or `vagrant destroy` as they'll run into file permission problems. So now the user is stuck, and there are no instructions we can give them to resolve the issue.

So we need an option as an escape hatch to let the user run through the recovery process to undo the damage `sudo` caused. We don't want a plain `--allow-root` option because then they'll add it and continue as normal. Instead Reacts approach with `dangerouslySetInnerHTML` seems to be the appropriate path forward.

Thus, if someone is using `sudo vagrant` we can tell them to do something similar to this:

```shell
sudo vagrant up --i-need-to-recover-data-to-stop-using-sudo
.... vagrant output and a big warning from us
sudo vagrant ssh -c "db_backup" --i-need-to-recover-data-to-stop-using-sudo
sudo vagrant destroy --force --i-need-to-recover-data-to-stop-using-sudo
sudo rm -rf .vagrant
.. cleanup virtualbox folders
vagrant up --provision
```

Mte90:
I don't know how many people we have that are part of this usercase, I mean implement this recovery solution as example for 1? of VVV users is worth it?
I think that for those the best solution is to keep the previous version do a backup and later update also because with this alert of this script will be blocked to execute VVV with sudo, they will be forced to change their use.
I don't know what happens when they try to execute Vagrant after it was used sudo, maybe there are some errors but reverting to the previous version should fix everything, after all will be executed as root that have all the permissions.

tomjn:
> I don't know how many people we have that are part of this usercase, I mean implement this recovery solution as example for 1? of VVV users is worth it?

I've come across 2, and I know the moment we merge this at least 1 will pop up and we'll have to handle the situation. It'll be someone who installed via a `zip` too just to make our lives difficult.

Also, if someone is in this situation, going back to an older version could cause problems that prevent them completing the steps, perhaps due to a bug we don't know about yet, or because in 2 years we decide to change how something works in a non-compatible manner.

Mte90:
I think that anyway to fix for this cases is a restart from scratch and a manual reset of the permissions of the files.
Handle from VVV can create other issues and will be difficult to manage.

tomjn:
To do that you would need to run `sudo vagrant destroy` and `sudo vagrant halt` which wouldn't work because of this code. We'd then need to walk a user who may not be knowledgeable about this step by step to do manual instructions. I know at least one user went with `sudo` because they thought if you were doing anything on Linux you needed to `sudo` and VMs sounded like a system thing so `sudo` made sense.

Making those users jump through extra hoops of downgrading, and forcing them to trawl through the changelog to figure out which version they need to downgrade to, and hold their hand at every step is painful for all involved. If it were easy for them and us then they would never have used `sudo` in the first place.

So we need to be more forgiving if only for our own sanity, remember we have to support these users so we're making things harder for ourselves too.

Imagine if someone had ran `sudo vagrant up` at a contributor day at WCUSA, and had just updated via a button in their editor by accident, and is complaining we broke their VVV.

Mte90:
I understand the issue but I don't think that is failure proof.
For me, with sudo vagrant create also the VM with that permission so I am wondering why they are not getting this alert in vagrant. 
Maybe they are using an old vagrant version and an old VVV version?

To me the best option it is to ship this change as it is now or not (because vagrant already has this alert), wait when there will be this requests because create now a tool or whatever that fix for these users without a real case for testing is not the best way to consider this case closed.

tomjn:
> but I don't think that is failure proof

There is no failure proof solution, we can't automate it, so we have to make it as easy as possible to fix this when it happens, and in the case of MacOS users, make it possible to fix it at all.

Unless we're able to build out the parameter immediatley when users show up, we'll have developers unable to work unless they go to a competitor. If we can build it out immediatley then we should just do it now.

In the meantime there are other mitigations that can happen, such as putting a flag inside the VM and making the dashboard show only a big warning with instructions on migrating off of `sudo`. In this scenario we need to be forgiving and provide as many paths as possible to getting users out of this predicament. It is not the time to be restrictive. Otherwise it's easier to modify the vagrant file to delete this block of code than it is to downgrade fix the use of `sudo`.

Right now, we know for a fact that if this PR is merged, then users who use `sudo` on MacOS have no way of recovering. If we were going to merge it then it would need to be changed so that it didn't prevent `sudo` but warned instead, meaning we would need to deprecate the use of `sudo` over 2 major releases. I'm not willing to make that kind of change, so the PR can't be merged without this parameter being implemented.

tomjn:
We can follow this up with another PR that exits when this happens, and adds the parameter to allow recovery steps to be taken for those who've already made this mistake ( case in point someone saw our conversation and raised it in Slack that they'd been using `sudo` too )

tomjn:
Screenshot 2020-07-20 at 12 26 53


That's what this looks like. Note the message at the end. It's not vagrant preventing you from using `sudo`, it's complaining the user ID is different. This is because I did not use `sudo` to create my VM. Instead, if I had ran `sudo vagrant up --provision` when I first created the VM, then I would get this message when I tried to use `vagrant` without `sudo`

tomjn:
Additionally, I don't see any reason to require the parameter or exit if `sudo vagrant halt` or `sudo vagrant destroy` are used. We want to destroy `sudo` created VMs. It's creating `sudo` VMs we want to avoid

PR #2228, opened 2020-08-19
Ignore windows user for administration warnings

Mte90:
Detect windows and ignore for warning

## Checks

* [ ] I've tested this PR


tomjn:
The check was using bitwise and instead of logical and, so I refactored it to use `unless` and the vagrant utils library instead

tomjn (approved):

PR #2230, opened 2020-08-25
Refactor main provisioner

msaggiorato:
This is a draft PR to finally make the main provisioner truly modular.

This is still ongoing and not finished, just pushing it here as draft for preview purposes.

### New features

- [x] Hook system to trigger function execution, even tho the hook system would initially only be used in the main provisioner, it would be made available to all provisioners to use as desired.
- [x] Package installation (apt-get) function has been moved to a reusable function, so that other provisioners that install packages and need this can reuse this.
- [x] All software being provisioned have it's own folder now with it's own "plugin" provisioner. These provisioners hook into the flow of things and perform actions at specific points (similar to how WP hooks work).

Things to do in order for this to be considered mergeable
- [x] Convert HEREDOC usage to file inclusion through something like `cat file.list >> append-to-this-file.list`
- [ ] Improve messagging (maybe)
- [ ] Improve file structure (maybe)
- [ ] Fix hook system to work with negative priorities (not a priority right now, but maybe a nice to have)
- [ ] Test test test test

To be worked on separate PRs after this one is merged:
- [x] Improving each "plugin" provisioner to be less hardcoded (eg: default PHP version).
- [ ] Allow removal of something that should not be provisioned.
- [ ] Improve coding style by doing a shellcheck pass on VSCode (bash linting)
- [ ] SVN check should be turned into a separate command


provision/core/deprecated.sh
```diff
@@ -0,0 +1,83 @@
+#!/bin/bash
+#
+# core/deprecated.sh
+#
+# This file contains deprecated functions that may be removed later.
+
+depreacted_distro() {
```
tomjn, line 7: deprecated*

provision/core/php/provision.sh
```diff
@@ -0,0 +1,140 @@
...
+  # Graphviz
+  #
+  # Set up a symlink between the Graphviz path defined in the default Webgrind
+  # config and actual path.
+  echo " * Adding graphviz symlink for Webgrind..."
+  ln -sf "/usr/bin/dot" "/usr/local/bin/dot"
+}
+export -f graphviz_setup
+
+vvv_add_hook after_packages graphviz_setup 20
```
tomjn, line 81: Should this be in here? Perhaps this should move to the  webgrind provisioner

provision/core/php/provision.sh
```diff
@@ -0,0 +1,140 @@
...
+
+phpfpm_setup() {
```
tomjn, line 83: Take a look at the PHP v8 PR on the utilities repo, it swaps a lot of these out for version variables, and simplifies file names, it'll make bumping the default from 7.2 to 7.3 easier

provision/core/vvv/provision.sh
```diff
@@ -0,0 +1,104 @@
...
+  echo " * Installing Shyaml for bash provisioning.."
+  sudo pip install shyaml
+}
+export -f shyaml_setup
+
+vvv_add_hook after_packages shyaml_setup 0
+
+vvv_add_hook services_restart "service ntp restart"
+
+function vvv_finalize_svn_check() {
```
tomjn, line 57: I'd really love it if this was in the utilities repo instead and opt in, it's useless to most people now

tomjn: Or maybe even a new script in homebin alongside `xdebug_on` etc

provision/core/php/provision.sh
msaggiorato: It was here so I left it. I didn't get to cleanup yet. Just moving things around so far.

Glad to axe it.

provision/core/php/provision.sh
Mte90: maybe we can create a config file for those version default constants and avoid hardcoded stuff? just an idea

msaggiorato: Looking at the PHP provisioners now, they could use some cleanup to reuse partially the same code 🤔. 

But to what concerns this PR, the PHP 8 provisioner would be definetely helpful to a separate PR here and looking into potentially changing the default PHP version installed through a setting in config.yml.

provision/core/composer/provision.sh
```diff
@@ -0,0 +1,50 @@
...
+  chown -R vagrant:www-data /usr/local/src/composer
+  chown -R vagrant:www-data /usr/local/bin
+
+  # COMPOSER
+
+  export COMPOSER_ALLOW_SUPERUSER=1
+  export COMPOSER_NO_INTERACTION=1
+
+  echo " * Checking for Composer"
+  exists_composer="$(which composer)"
```
Mte90, line 17: I like this way to keep the consistency and explain what the code does

provision/core/env.sh
```diff
@@ -0,0 +1,106 @@
...
+    -e "s|:/srv/config/homebin||" \
+    -e "s|/srv/config/homebin:||" \
+    -e "s|(.*PATH.*?\".*?)(\")|\1:/srv/config/homebin\2|" \
+    /etc/environment
+}
+
+cleanup_terminal_splash() {
+  # Dastardly Ubuntu tries to be helpful and suggest users update packages
+  # themselves, but this can break things
+  if [[ -f /etc/update-motd.d/00-header ]]; then
```
Mte90, line 27: Wondering if we can do something to remove all of them and simplify this

provision/core/grunt/provision.sh
```diff
@@ -0,0 +1,51 @@
+
+function install_grunt() {
+  echo " * Installing Grunt CLI"
+  npm_config_loglevel=error npm install -g grunt grunt-cli --no-optional
```
Mte90, line 4: maybe create an alias so we can simplify, like `npm-log-no-optional`

provision/core/mailhog/provision.sh
```diff
@@ -0,0 +1,53 @@
...
+    echo " * Installing MHSendmail"
+    curl --silent -L -o /usr/local/bin/mhsendmail https://github.com/mailhog/mhsendmail/releases/download/v0.2.0/mhsendmail_linux_amd64
+    chmod +x /usr/local/bin/mhsendmail
+  fi
+
+  if [[ ! -e /etc/systemd/system/mailhog.service ]]; then
+    echo " * Mailhog service file missing, setting up"
+    # Make it start on reboot
+    tee /etc/systemd/system/mailhog.service <<EOL
+[Unit]
```
Mte90, line 23: maybe extract this as file now that we have a mailhog folder so the script will be simplified?

provision/core/mariadb/provision.sh
```diff
@@ -0,0 +1,155 @@
...
+
+if ! vvv_apt_keys_has 'MariaDB'; then
+  # Apply the MariaDB signing keyg
+  echo " * Applying the MariaDB signing key..."
+  apt-key add /srv/config/apt-keys/mariadb.key
+fi
+
+if ! vvv_src_list_has "nginx.org"; then
+  cat <<VVVSRC >> /etc/apt/sources.list.d/vvv-sources.list
+# MariaDB 10.4 Amsterdam
```
Mte90, line 39: maybe also the same for this list in a file?

provision/core/php/provision.sh
```diff
@@ -0,0 +1,140 @@
...
+  # Copy php-fpm configuration from local
+  echo " * Copying /srv/config/php-config/php7.2-fpm.conf   to /etc/php/7.2/fpm/php-fpm.conf"
```
Mte90, line 85: Wondering if now that we have a folder we can automatize this with a loop and simplify

provision/core/wp-cli/provision.sh
```diff
@@ -0,0 +1,30 @@
+wp_cli_setup() {
+  echo " "
+  echo " * Installing/updating WP-CLI"
+  # WP-CLI Install
+  local exists_wpcli
+
+  # Remove old wp-cli symlink, if it exists.
+  if [[ -L "/usr/local/bin/wp" ]]; then
```
Mte90, line 8: maybe move this to the deprecated file?

provision/provision-helpers.sh
```diff
@@ -204,3 +223,45 @@ function get_config_keys() {
   echo "${value:-$2}"
 }
 export -f get_config_keys
+
+#
+# hook engine
+#
+vvv_add_hook() {
+  if [[ "${1}" =~ [^a-zA-Z_] ]]; then
+    echo "Disallowed hookname"
```
Mte90, line 232: maybe use vvv_warn or something that help with debug?

Mte90 (changes requested):

provision/core/grunt/provision.sh
Mte90: maybe for this is better a @tomjn opinion

provision/core/mailhog/provision.sh
tomjn: We could create a utility for this but we can't remove mailhog from VVV core in v3. We could deprecate and remove in v4.

Of all the things in VVV core installed by default, I think Mailhog is important enough to come out of the box

provision/core/grunt/provision.sh
tomjn: If this is only used in this file then I don't think it's too important

provision/core/env.sh
tomjn: we don't need to cleanup MOTD, but it does mean when you SSH in you'd get a tonne of ubuntu advertisements and random stuff about cloud services etc. So I'd prefer to keep it

provision/core/mailhog/provision.sh
Mte90: For sure, I mean we can remove the code of that file from the bash script and put as file that is copied.
So it is simplified.

provision/core/mariadb/provision.sh
tomjn: A `vvv-sources.list` that gets copied into place sounds good

provision/core/env.sh
Mte90: Me too but like remove all the files from that folder can be an idea?

provision/core/php/provision.sh
tomjn: Not quite, the files themselves have references to the PHP versions. We can simplify this though by copying a `php-fpm.conf` to `php-fpm.conf` rather than a `php7.2-fpm.conf`.

provision/core/wp-cli/provision.sh
tomjn: I think it makes sense as WP CLI wasn't deprecated, this is just migration code

provision/provision-helpers.sh
tomjn: Agreed, perhaps also mention the hook name in quotes so it can be tracked

provision/core/mariadb/provision.sh
Mte90: maybe `mariadb-vvv-sources.list` so every tool can have their own file

provision/provision-helpers.sh
tomjn: Perhaps this?

```suggestion
    vvv_warn "Invalid hookname '${1}', hooks must only contain the characters A-Z and a-z"
```

provision/core/mailhog/provision.sh
tomjn: Agreed

provision/core/php/provision.sh
tomjn: I don't know how I feel about changing the default version via `config.yml`, as it impacts a lot of the tooling such as PHPMyAdmin or the dashboard, at the moment we have the luxury of pushing the default upwards ourselves while users can set PHP per site

provision/core/mariadb/provision.sh
tomjn: Maybe, but we need to be mindful that `vvv-sources.list` exists, otherwise `apt` will complain about the same source appearing more than once ( I currently get this for PHP locally at the moment )

msaggiorato: I could have the file, and do the same check i have right now, but instead of using HEREDOC using cat.

provision/core/deprecated.sh
msaggiorato: Fixed

msaggiorato:
Updated the PR description to have a rough roadmap of this PR. Feel free to comment or suggest anything else.

I'll likely work on this tomorrow, got a ton of work to go through today.

provision/core/mariadb/provision.sh
tomjn: A specific file would be better, force copying a `vvv-sources.list` into place avoids a lot of checks and potential bugs. We can always copy additional source files if we want to append extra sources

msaggiorato: Done

provision/core/mailhog/provision.sh
msaggiorato: Done

provision/core/php/provision.sh
msaggiorato: Did file rename here

provision/core/php/provision.sh
msaggiorato: Defaulted via variable (that can be extended in the future to be set in a broader scope, so that it can be referenced by other provisioners as the default version).

That last change is out of the scope of this PR tho.

provision/core/php/provision.sh
msaggiorato: Lets axe this on another PR if needed.

provision/core/vvv/provision.sh
msaggiorato: Added to the main PR to have it in mind as a separate PR.

msaggiorato:
Ok, turning this into "Ready for review".

I've provisioned a site on develop, switched branches, and reprovisioned, and the default sites work.

I've provisioned from scracth, and also the default sites work.

Feel free to leave feedback about any wording change that may be needed before merge.

Mte90:
The code is fine to for me, we need just to test it.
After approve it we can see how to improve the code.

Mte90 (dismissed):
Tested and no errors reported in an already installed system.
Maybe we need a test on a new machine from scratch?

PR #2231, opened 2020-08-30
Fix provision-site fail situations

msaggiorato:
You cannot return from a sourced script, only from functions, so we should exit.


Mte90:
This kill the whole process or just that script? Just to be sure :-)

msaggiorato:
Each provisioner run on a separate shell. So by exiting we're only affecting the current shell.

The exit code 1 tho, may make the full provisioning stop (as vagrant may think this is non recoverable)

This would need some testing on these fail situations to see how it behaves.

Mte90 (approved):

Mte90:
Waiting for a check by tomjn

tomjn (approved):

tomjn:
@msaggiorato we'll need a follow up PR to update the changelog

PR #2233, opened 2020-08-31
GH Action - Run Provisioners

msaggiorato:
This is something i've been working over the weekend.

It will attempt to simulate the same environment the provisioners get inside the VM, so that it runs all the provisioners.

So far, this only runs the provisioners (using part of the refactor we've made a couple months ago) and make the full logs available for download as artifacts.

Ideally, we could add extra steps here with some sort of test suite as a separate PR.

------

Couple of things I tried and didn't work:

- Running this on an official ubuntu:18.04 docker image fails due to several reasons, the main ones being: no systemd available, and no access to /etc/hosts (as that's managed by docker)
- Running the provisioner from vagrant (full test), doesn't work because GH Actions doesn't allow a VM to run in a VM (due to lack of VT-x support).
- Running on CircleCI get similar results as above, same limitations.
- Using symlinks to "simulate" mounts has a bunch of permission issues. bindfs basically forces user and group on the mounted location, therefore there's no ownership/permission issues (similar to how synced folders work in VBox).

### DIFF
--- a/.github/workflows/vvv-provisioning.yml
+++ b/.github/workflows/vvv-provisioning.yml
@@ -89,6 +89,7 @@ jobs:
       # - Check VM hostfile
 
       - name: Prepare Output
+        if: ${{ always() }}
         run: |
           sudo cp -rf "$GITHUB_WORKSPACE/log" "/vvv-logs"
           MYUID=$(id -u -n)
@@ -96,6 +97,7 @@ jobs:
           sudo chown -R $MYUID:$MYGID "/vvv-logs"
 
       - uses: actions/upload-artifact@v2
+        if: ${{ always() }}
         with:
           name: logs
           path: "/vvv-logs"
--- a/.github/workflows/vvv-provisioning.yml
+++ b/.github/workflows/vvv-provisioning.yml
@@ -72,9 +72,27 @@ jobs:
           vbox_mount "$GITHUB_WORKSPACE/log/php" "/var/log/php" "root" "syslog" "666"
           vbox_mount "$GITHUB_WORKSPACE/log/provisioners" "/var/log/provisioners" "root" "syslog" "666"
 
-      # Runs the main provisioner as sudo
-      - name: Run all provisioners
-        run: sudo bash /srv/provision/tests/run-provisioners.sh
+      # Runs the provisioners in the expected order
+      - name: Run provison-pre.sh
+        run: sudo bash -c '. "/srv/provision/tests/provisioners.sh" && pre_hook'
+
+      - name: Run provison.sh
+        run: sudo bash -c '. "/srv/provision/tests/provisioners.sh" && provision_main'
+
+      - name: Run provison-dashboard.sh
+        run: sudo bash -c '. "/srv/provision/tests/provisioners.sh" && provision_dashboard'
+
+      - name: Run provison-utility-source.sh
+        run: sudo bash -c '. "/srv/provision/tests/provisioners.sh" && provision_utility_sources'
+
+      - name: Run provison-utility.sh
+        run: sudo bash -c '. "/srv/provision/tests/provisioners.sh" && provision_utilities'
+
+      - name: Run provison-site.sh
+        run: sudo bash -c '. "/srv/provision/tests/provisioners.sh" && provision_sites'
+
+      - name: Run provison-post.sh
+        run: sudo bash -c '. "/srv/provision/tests/provisioners.sh" && post_hook'
 
       # At this point, we would run some extra tests
       # TODO: Ideas
--- a/.github/workflows/vvv-provisioning.yml
+++ b/.github/workflows/vvv-provisioning.yml
@@ -1,6 +1,6 @@
 # This is a basic workflow to help you get started with Actions
 
-name: CI
+name: VVV Provisioning
 
 # Controls when the action will run. Triggers the workflow on push or pull request
 # events but only for the develop branch
@@ -8,12 +8,13 @@ on:
   push:
     branches: [ develop ]
   pull_request:
-    branches: [ develop ]
+    branches: [ develop, stable ]
 
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
   # This workflow contains a single job called "build"
   build:
+    name: Run Provisioner
     # The type of runner that the job will run on
     runs-on: ubuntu-18.04
 
@@ -70,6 +71,12 @@ jobs:
       # Runs the main provisioner as sudo
       - name: Run all provisioners
         run: sudo bash /srv/provision/tests/run-provisioners.sh
+
+      # At this point, we would run some extra tests
+      # TODO: Ideas
+      # - Add screenshots of provisioned sites
+      # - CURL mailhog API to see if it's working or not
+      # - Check VM hostfile
 
       - name: Prepare Output
         run: |
--- a/.github/workflows/vvv-provisioning.yml
+++ b/.github/workflows/vvv-provisioning.yml
@@ -41,8 +41,12 @@ jobs:
 
           # vbox_mount function to sumulate synced folders
           function vbox_mount() {
+            local perms="-p ug=rwx,o=r";
+            if [ ! -z "${5}" ]; then 
+              perms="-p ${5}"
+            fi
             sudo mkdir -p "$2"
-            sudo bindfs --chown-ignore --chgrp-ignore --chmod-ignore -u "$3" -g "$4" "$1" "$2"
+            sudo bindfs --chown-ignore --chgrp-ignore --chmod-ignore -u "$3" -g "$4" "$1" "$2" $perms
           }
 
           # create srv folder
@@ -62,11 +66,11 @@ jobs:
           vbox_mount "$GITHUB_WORKSPACE/config" "/srv/config" "vagrant" "vagrant"
           vbox_mount "$GITHUB_WORKSPACE/provision" "/srv/provision" "vagrant" "vagrant"
           vbox_mount "$GITHUB_WORKSPACE/certificates" "/srv/certificates" "vagrant" "vagrant"
-          vbox_mount "$GITHUB_WORKSPACE/www" "/srv/www" "vagrant" "www-data"
-          vbox_mount "$GITHUB_WORKSPACE/log/memcached" "/var/log/memcached" "root" "syslog"
-          vbox_mount "$GITHUB_WORKSPACE/log/nginx" "/var/log/nginx" "root" "syslog"
-          vbox_mount "$GITHUB_WORKSPACE/log/php" "/var/log/php" "root" "syslog"
-          vbox_mount "$GITHUB_WORKSPACE/log/provisioners" "/var/log/provisioners" "root" "syslog"
+          vbox_mount "$GITHUB_WORKSPACE/www" "/srv/www" "vagrant" "www-data" "ug=rwx,o=r,o+D"
+          vbox_mount "$GITHUB_WORKSPACE/log/memcached" "/var/log/memcached" "root" "syslog" "666"
+          vbox_mount "$GITHUB_WORKSPACE/log/nginx" "/var/log/nginx" "root" "syslog" "666"
+          vbox_mount "$GITHUB_WORKSPACE/log/php" "/var/log/php" "root" "syslog" "666"
+          vbox_mount "$GITHUB_WORKSPACE/log/provisioners" "/var/log/provisioners" "root" "syslog" "666"
 
       # Runs the main provisioner as sudo
       - name: Run all provisioners
--- a/.github/workflows/vvv-provisioning.yml
+++ b/.github/workflows/vvv-provisioning.yml
@@ -34,19 +34,12 @@ jobs:
           sudo groupadd -g 2000 vagrant
           sudo useradd -u 2000 -g vagrant -m vagrant
 
-          # Install bindfs, which would allow us to "mount" folders with specific permissions
-          # similar to vagrant shares
-          sudo apt-get update
-          sudo apt-get install bindfs -y
-
-          # vbox_mount function to sumulate synced folders
-          function vbox_mount() {
-            local perms="-p ug=rwx,o=r";
-            if [ ! -z "${5}" ]; then 
-              perms="-p ${5}"
+          # vvv_symlink function to sumulate synced folders
+          function vvv_symlink() {
+            if [ ! -d "${1}" ]; then 
+              sudo mkdir -p "${1}"
             fi
-            sudo mkdir -p "$2"
-            sudo bindfs --chown-ignore --chgrp-ignore --chmod-ignore -u "$3" -g "$4" "$1" "$2" $perms
+            sudo ln -sf "${1}" "${2}"
           }
 
           # create srv folder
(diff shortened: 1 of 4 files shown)


.github/workflows/vvv-provisioning.yml
```diff
@@ -0,0 +1,91 @@
...
+          function vbox_mount() {
+            sudo mkdir -p "$2"
+            sudo bindfs --chown-ignore --chgrp-ignore --chmod-ignore -u "$3" -g "$4" "$1" "$2"
+          }
+
+          # create srv folder
+          sudo -u "vagrant" mkdir -p "/srv"
+
+          # copy files provided by vagrant
+          sudo cp -f "$GITHUB_WORKSPACE/config/default-config.yml" "$GITHUB_WORKSPACE/config/config.yml"
```
tomjn, line 52: Is this necessary? The vagrantfile copies `config/config.yml` to `/vagrant/config.yml` and that's what the provisioners work off of, in this case could we not copy it directly there? Or even a special version with a cut down site list?

.github/workflows/vvv-provisioning.yml
```diff
@@ -0,0 +1,91 @@
...
+          vbox_mount "$GITHUB_WORKSPACE/certificates" "/srv/certificates" "vagrant" "vagrant"
+          vbox_mount "$GITHUB_WORKSPACE/www" "/srv/www" "vagrant" "www-data"
+          vbox_mount "$GITHUB_WORKSPACE/log/memcached" "/var/log/memcached" "root" "syslog"
+          vbox_mount "$GITHUB_WORKSPACE/log/nginx" "/var/log/nginx" "root" "syslog"
+          vbox_mount "$GITHUB_WORKSPACE/log/php" "/var/log/php" "root" "syslog"
+          vbox_mount "$GITHUB_WORKSPACE/log/provisioners" "/var/log/provisioners" "root" "syslog"
+
+      # Runs the main provisioner as sudo
+      - name: Run all provisioners
+        run: sudo bash /srv/provision/tests/run-provisioners.sh
```
tomjn, line 73: If it's possible I'd like some of the provisiooners to be their own steps so we can see froom GH Actions which one failed, otherwise it'll always fail on `Run all provisioners` requiring us to look through all the artefacts

tomjn (commented):
This tries to simulate running from the host, and mounted folders, but I dn't think we need to do that. We should be simulating the guest. So symlink `/srv/www`  to the `www`  folder, or, use `/srv/www` as is outright, no mapping of any kind. That's the vagrantfile/docker files job when setting up the environment, which GH Actions has already done for us. We're not testing the creation of a VM or container here, we're testing provisioners.

I'm also not bothered about permissions. We can set the initial permissions of the folders then just make sure things run as the user they need to be.

I also think we can do some prep work for this, removing reliance on the `vagrant` user which  is a bit weird in a docker context, or at least making it pluggable, or unnecessary. E.g. relying on a `vvv` user group.

There may also be an alternative lighter weight image of Ubuntu 18 that doesn't contain PHP out of  the box

tomjn:
I'd also like us to deprecate `/vagrant`

.github/workflows/vvv-provisioning.yml
msaggiorato: We do not run vagrantfile here. Therefore this is required unless we find a way to run VirtualBox in a container.

.github/workflows/vvv-provisioning.yml
msaggiorato: I'll look into separating this

msaggiorato:
What I am doing is trying to run on the VM provided by GH (this is not a docker image). Effectively simulating the environment the provisioners get on our VM at the time of the first run.

The main drawback is the built in software on the environment provided by GH. I'm specifically removing PHP versions here to test the install process of those.

I haven't found any official image that had proper support for systemd (which is what services rely on), and also I couldn't find any way to have access to the docker container /etc/hosts file (it's locked by default by docker unless it's specifically set as a mount point ad far as I saw).

I've just now found an issue with the mounted folders which I have almost solved and been testing right now.

I'm totally onboard to run this on docker entirely. But we'd need to overcome the issues above (and also mentioned on the PR description)

Again as described above. Using bindfs is the best way I found to simulate the current initial guest state and avoid permission issues that come up because of having a different scenario and that are non issues in real life cases.

.github/workflows/vvv-provisioning.yml
tomjn: I know, but since there is no host of any kind, why copy the default config to `config/config.yml` then copy that file again to `/vagrant/config.yml`? The provisioners never load `config/config.yml`, that's the vagrantfiles job, but since there is no vagrant file, we don't need it. Stop emulating the actions the host takes, and instead emulate the actions inside the environment.

tomjn:
> The main drawback is the built in software on the environment provided by GH. I'm specifically removing PHP versions here to test the install process of those.

It's just a container, you can choose a different image for your GH Action to run inside

For systemd: https://stackoverflow.com/questions/53383431/how-to-enable-systemd-on-dockerfile-with-ubuntu18-04

> Again as described above. Using bindfs is the best way I found to simulate the current initial guest state and avoid permission issues that come up because of having a different scenario and that are non issues in real life cases.

At some point, our provisioners will run in an environment where the ownership and permissions aren't enforced via VirtualBox, exposing any bugs in our provisioners. We should test for that now.

Keep in mind, the VirtualBox behaviour you're trying to reproduce doesn't exist in Hyper-V or VMWare.

Set the initial ownership and make sure the provisioners run as the correct users. If there's a permission problem as a result, then that's a bug in VVV, not the action. It's the provisioner that should be fixed.


For `/vagrant` symlink things in but don't rely on it. I'd like a new place for the files in that folder. That work can go in a separate PR though

.github/workflows/vvv-provisioning.yml
msaggiorato: https://github.com/Varying-Vagrant-Vagrants/VVV/blob/develop/provision/provision.sh#L44-L51

You can see that if the main provisioner clear whatever is on /vagrant/config.yml, and copies the file from the config/config.yml location. This file does not exist on the repo, and does not exist until created via Vagrantfile. If we don't do that step manually at any point, then we cannot run the provisioners.

One thing I thought tho, is to actually move the code that copies the file from default-config.yml to config.yml to the main provisioner, or something that runs before the main provisioner, as it's something still tying us to Vagrant.

I'm honestly trying to do at least as possible to get the provisioners to run on the current state. This clearly also exposed some things that could be moved to the provisioners to simplify the Vagrantfile further, and remove dependency to it.

msaggiorato:
> At some point, our provisioners will run in an environment where the ownership and permissions aren't enforced via VirtualBox, exposing any bugs in our provisioners. We should test for that now.
>
> Keep in mind, the VirtualBox behaviour you're trying to reproduce doesn't exist in Hyper-V or VMWare.
>
> Set the initial ownership and make sure the provisioners run as the correct users. If there's a permission problem as a result, then that's a bug in VVV, not the action. It's the provisioner that should be fixed.

I agree with this, yet I didn't think that the PR resulting action fails with the current state of the project was good. I can rollback to my initial symlink implementation if you want tho.

> For /vagrant symlink things in but don't rely on it. I'd like a new place for the files in that folder. That work can go in a separate PR though

I agree, subsequent cleanup could take place on separate PRs.

tomjn:
> I agree with this, yet I didn't think that the PR resulting action fails with the current state of the project was good. That's the reason why I'm using bindfs at this point.

We can fix those bugs upstream or as part of this PR, either way a bugs a bug. Just because it doesn't happen on VirtualBox doesn't mean it doesn't happen

> I can rollback to my initial symlink implementation if you want tho. We'd need to agree what would be the proper user to run the provisioners as tho. By default vagrant runs everything as root, yet that may not be ideal.

I've no issue with provisioners running as `root` since that's the current behaviour. Symlinking is probably best

.github/workflows/vvv-provisioning.yml
tomjn: Ah I see, lets resolve this then

msaggiorato:
Ok, rolled back to symlinks, without any permission forcing before the provisioners are run. Here are some issues that arise:

- Site provisioner is not able to create logs, since the repo is cloned as root.
- PHPCS is not able to be installed.
- Site provisioners don't fail properly when there are errors, therefore there's no errors on the action steps (and won't be any hard errors on the vagrant provisioning either)

I still need to get a clean docker image with the systemd replacement you recommended. I have seen that particular project during my research, but since it was something "simulated" and not "official" from ubuntu images, i skipped using it. I'll look into this.

Do we have a VVV docker hub or something? In case we want to push something like "vvv/ubuntu:18.04" to use as a base for these tests, and potentially in the future our docker version?

tomjn:
> Site provisioner is not able to create logs, since the repo is cloned as root.

Since this is testing frm the _guests_ point of view, we should drop the symlinks for the log folders, and use `/var/log` instead

> PHPCS is not able to be installed.
> Site provisioners don't fail properly when there are errors, therefore there's no errors on the action steps (and won't be any hard errors on the vagrant provisioning either)

Any details on these?

> I still need to get a clean docker image with the systemd replacement you recommended. I have seen that particular project during my research, but since it was something "simulated" and not "official" from ubuntu images, i skipped using it. I'll look into this.

See the stackoverflow link I mentioned further up, systemd isn't on docker out of the box intentionally, I don't believe it's a matter of switching image.

> Do we have a VVV docker hub or something? In case we want to push something like "vvv/ubuntu:18.04" to use as a base for these tests, and potentially in the future our docker version?

No, but I don't think it's necessary here

msaggiorato:
> Since this is testing frm the guests point of view, we should drop the symlinks for the log folders, and use /var/log instead

Check [here](https://github.com/Varying-Vagrant-Vagrants/VVV/pull/2233/checks?check_run_id=1055672892#step:9:5) for details about the error i'm referring to here. It's not the provisioner logs, it's the log folder inside each site.

> Any details on these?

Check [here](https://github.com/Varying-Vagrant-Vagrants/VVV/pull/2233/checks?check_run_id=1055672892#step:5:1103) for PHPCS and [here](https://github.com/Varying-Vagrant-Vagrants/VVV/pull/2233/checks?check_run_id=1055672892#step:9:5) for the site provisioners not failing appropiately (which may be desired, just need discussion).

> See the stackoverflow link I mentioned further up, systemd isn't on docker out of the box intentionally, I don't believe it's a matter of switching image.

Yes, that's what i'll be looking into. According to that project we need to do something like [this](https://github.com/gdraheim/docker-systemctl-images/blob/master/ubuntu-apache2.dockerfile) to get `service *` like commands. It might be a good opportunity to generate an image too, since sudo is not installed by default on the ubuntu official images aswell (as I've learned during my tests on this).

This is the reason I asked for a docker hub account we can push images to, as it's going to be required, since we cannot run on an image generated from a raw Dockerfile. We need to have it pushed, and I wanted to avoid having things belonging to the project under my name or something.

.github/workflows/vvv-provisioning.yml
```diff
@@ -0,0 +1,101 @@
...
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - uses: actions/checkout@v2
+
+      # Make Symlinks
+      - name: Create Vagrant Like Environment
+        run: |
+          # uninstall pre installed packages (to test if utilities work)
```
Mte90, line 29: To move on this pr we need to test utilities? We can do another issue for that task?
Maybe in the utilities repo?

tomjn:
> Check here for details about the error i'm referring to here. It's not the provisioner logs, it's the log folder inside each site.

I suspect the folder has incorrect ownership and permissions. `bindfs` and symlinks aren't the fix for that, correct ownership and permissions are


Also, for systemd, I do not want extra images. I would rather we added wrapper functions, e.g. something like this:

```shell
vvv_reload_nginx() {
    if systemd is available
        service nginx reload
    else
       /usr/sbin/nginx -s reload
}
```

msaggiorato:
Could we try supervisor to manage services? (as stated [here](https://docs.docker.com/config/containers/multi-service_container/) in official Docker docs).

We could use the approach mentioned above by Tom to keep compatibility with VMs that are not reprovisioned on the new version and therefore keep this backwards compatible.

```shell
vvv_reload_nginx() {
    if supervisor is available
        supervisorctl restart nginx
    else
        service nginx restart
}
```

According to my research, supervisor is popularly used as a process manager in docker environments (which again, is not the desired usage for docker). 

We'd need to write small configuration files for each service, but according to what i saw in the repos, there's not many services to be restarted (at least in core):

* PHP
* nginx
* mariadb
* ntp
* memcached
* mailhog

On vvv-utilities

* mongodb
* more PHP versions

Here's some example configuration for supervisor for MySQL

https://github.com/dayreiner/centos7-MariaDB-10.1-supervisord/blob/4202639d8ff300236a1480e2268e298e9aa1628f/config/supervisord.conf#L20-L24

Thinking we could use the "group" feature of supervisor to restart all registered PHP versions without hardcoding them.

Let me know what do you think of this?

tomjn:
@msaggiorato right now the goal here should be an MBP for GH actions , less so much docker support.

With that in mind I have no qualms with:

 - manually restarting the services explicitly
 - Only restarting those installed in the base setup, and ignoring those in utilities

That means no utility testing or installation.

In the future we can change things around for actual Docker support, e.g. not running the MariaDB provisioner and instead using a MariaDB container, or the Mailhog container

Mte90:
I agree let's start with a minimal version so we can see what other changes we have to do.

J2c about supervisor, maybe we can create an alias to restart services that will check like in the snippet above.

Mte90 (approved):

msaggiorato:
Ok, here are some issues i see here:

* [Here](https://github.com/Varying-Vagrant-Vagrants/VVV/pull/2233/checks?check_run_id=1124688390#step:5:1004) we see that PHPCS fails to provision due to permission issues on the symlinks (which i've previously bypassed by using bindfs). 
* [Here](https://github.com/Varying-Vagrant-Vagrants/VVV/pull/2233/checks?check_run_id=1124688390#step:6:5) we can see that the permission issues are also there for the dashboard provisioner (permissions on the www dir).
* [Here](https://github.com/Varying-Vagrant-Vagrants/VVV/pull/2233/checks?check_run_id=1124688390#step:9:5) we see issues again with the site provisioner and permissions on the www folder.

All of these show the error, but doesn't exit the respective provisioner. This may be due to a lack of `set -e` on the provisioners, which I think we're avoiding due to other issues that may appear.

Also, it's true that these errors doesn't appear under the current stack of VVV due to the way that permissions work on synced folders (which is similar to what bindfs does).

With the decision to disregard the permission specific tricks that happen in vagrant, and make things work on a docker-like environment, the question here is:

1. Do we want to work on these permission issues on separate PRs to detach ourselves from the way that vagrant / vbox handle permissions and merge this PR as is?
2. Do we fix these permission issues on this PR even if the fix affects the provisioner themselves and not the test suite?

Let's decide as a team what's best here.

PS: I've run some tests to run exit 1 on the provisioners and it's not failing as expected, so i'm fixing that, since that should be shipped with this PR for sure.

Mte90:
Well I think that is the case of approve this and do a new pr to approach those 2 problems.

msaggiorato:
Now the proper error is displayed, and the Action actually fails when the site provisioner cannot run.

I'd say we're good to merge and create issues for the subsequent changes fixes that are necessary.

tomjn:
@msaggiorato lets get the action passing first :)

msaggiorato:
Ok, we're back in business.

Also found out this issue https://github.com/Varying-Vagrant-Vagrants/custom-site-template/pull/45 and fixed it.

PR #2238, opened 2020-09-07
Minor cleanup changes

msaggiorato:
I've found some issues while experimenting with docker provisioning, and here are some minor fixes/gotchas

Also, i've rearranged the PHP provisioner and moved anything memcached to it's own section at the bottom, so it's easier to move to another provisioner in the future.

Ideally memcached could be turned into an optional utility, that you could choose among others, eg: redis.

### DIFF
--- a/provision/core/deprecated.sh
+++ b/provision/core/deprecated.sh
@@ -5,6 +5,11 @@
 # This file contains deprecated functions that may be removed later.
 
 function deprecated_distro() {
+  local command_exist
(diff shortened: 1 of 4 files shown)


Mte90:
I agree too that memcache need to be moved as utility, is not used by all the users.
In the past the question was that in WordPress VIP needs that but I think that if we want to improve the performance and simplify move it as utility is the right way.

tomjn (approved):

PR #2264, opened 2020-10-19
WP alias with `--no-ssh`

Mte90:
In this way we can use wp-cli inside the VM without any issue.


tomjn:
Users would not though, and anybody who has forked or built their own provisioners will have their scripts broken

Mte90:
So you mean to change the alias just in the template?

tomjn:
I'm not a fan of using an alias to change what `wp` means to add hidden parameters. What we want is not possible with the current WP CLI implementation, we shouldn't introduce workarounds. It would be better to work with the WP CLI project to make it possible using just the config files.

PR #2267, opened 2020-10-21
Allow hosts yml property to default to the site name.

msaggiorato:
Allow hosts to default to the sitename, reducing the require config

This would basically allow something like this

```yml
sites:
  wordpress-one:
```

And this site will provision as `wordpress-one.test`

### DIFF
--- a/provision/provision-site.sh
+++ b/provision/provision-site.sh
@@ -124,7 +124,7 @@ function vvv_process_site_hosts() {
     fi
   else
     echo " * Adding hosts from the VVV config entry"
-    for line in $(get_hosts_list); do
+    for line in $hosts; do
       if [[ -z "$(grep -q "^127.0.0.1 $line$" /etc/hosts)" ]]; then
         echo "127.0.0.1 ${line} # vvv-auto" >> "/etc/hosts"
         echo "   - Added ${line} from ${VVV_CONFIG}"
--- a/provision/provision-site.sh
+++ b/provision/provision-site.sh
@@ -24,8 +24,7 @@ VVV_CONFIG=/vagrant/config.yml
 # Takes 2 values, a key to fetch a value for, and an optional default value
 # e.g. echo $(get_config_value 'key' 'defaultvalue')
 function get_config_value() {
-  local value=$(shyaml -q get-value "sites.${SITE_ESCAPED}.custom.${1}" "${2}" < ${VVV_CONFIG})
-  echo "${value}"
+  vvv_get_site_config_value "custom.${1}" "${2}"
 }
 
 function get_hosts() {
--- a/provision/provision-site.sh
+++ b/provision/provision-site.sh
@@ -30,12 +30,17 @@ function get_config_value() {
 
 function get_hosts() {
   local value=$(shyaml get-values-0 "sites.${SITE_ESCAPED}.hosts" 2> /dev/null < ${VVV_CONFIG} | tr '\0' ' ' | sed 's/ *$//')
-  echo "${value:-$@}"
+  echo "${value:-"${VVV_SITE_NAME}.test"}"
+}
+
+function get_hosts_list() {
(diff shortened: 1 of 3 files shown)


Mte90:
This can help in the cases when that that field is not defined.
We need to document this change.

provision/provision-site.sh
```diff
@@ -30,12 +30,17 @@ function get_config_value() {
 
 function get_hosts() {
   local value=$(shyaml get-values-0 "sites.${SITE_ESCAPED}.hosts" 2> /dev/null < ${VVV_CONFIG} | tr '\0' ' ' | sed 's/ *$//')
-  echo "${value:-$@}"
+  echo "${value:-"${VVV_SITE_NAME}.test"}"
+}
+
+function get_hosts_list() {
+  local value=$(shyaml get-values "sites.${SITE_ESCAPED}.hosts" 2> /dev/null < ${VVV_CONFIG})
```
tomjn, line 37: Can we do this the same way as `if folders=$(shyaml keys -y -q "sites.${SITE_ESCAPED}.folders" < "${VVV_CONFIG}"); then` in the folders setup?

tomjn: also no `/dev/null`

provision/provision-site.sh
```diff
@@ -120,7 +125,7 @@ function vvv_process_site_hosts() {
     fi
   else
     echo " * Adding hosts from the VVV config entry"
-    for line in $(shyaml get-values "sites.${SITE_ESCAPED}.hosts" 2> /dev/null < ${VVV_CONFIG}); do
+    for line in $(get_hosts_list); do
```
tomjn, line 128: shouldn't it be looping over `hosts` instead of `get_hosts_list`?

tomjn (changes requested):

tomjn:
Good change, I thought we already did this!

tomjn:
Also needs changelog updating

provision/provision-site.sh
msaggiorato: `get_hosts_list` is a new function i added which does precisely this. Created a function to reduce code duplication slightly.

`get_hosts` does the same, but puts this in a single line, separated by space, to be used on the nginx conf.

provision/provision-site.sh
msaggiorato: Replaced `/dev/null` with `-q` . I think the if structure here might overcomplicate the syntax, since basically the return statement returns the default if `value` is empty.

provision/provision-site.sh
tomjn: I get that, it's more that you used the function several lines up and stored the result in a variable, we might avoid duplicated calls

provision/provision-site.sh
tomjn: it'd avoid certain issues with traps and if `shyaml` returned a bad error code. Also do we need the `-q`? If `shyaml` encounters a problem we'll be unable to debug it if it's silenced

provision/provision-site.sh
msaggiorato: 🤦 Fixed

provision/provision-site.sh
msaggiorato: The problem could be simply that the path is not defined, which is why we default to the sitename. Afaik there's no default parameter for a list type of return of shyaml.

Can you elaborate on the traps issues? I'm not aware of what you mean.

tomjn: So originally I needed to know how to get a list of items but safely handle the situation when the path didn't exist, and the proposed method is what I got back from the shyaml author

msaggiorato: So what would be the errors that shyaml would throw that is not a path error?

tomjn (approved):
this might need an adjustment to the dashboard repo to handle these sites

provision/provision-site.sh
tomjn: I'm not sure of the full list, but YAML syntax errors, and outright crashes in Shyaml itself could happen

Issue #2272, opened 2020-10-22
Conditional apt package update


tomjn:
We always want to install things if they're not present, and we always want to make sure when VVV updates stuff gets provisioned, but do we have to update them every single time?

We don't have to update all the packages every single time, perhaps just at least once a day? lets make it configurable

PR #2281, opened 2020-10-23
Avoid running apt-get more than once per hour

msaggiorato:
This attempts to speed up provision by addressing #2272

### DIFF
--- a/provision/provision.sh
+++ b/provision/provision.sh
@@ -12,6 +12,7 @@
 . "/srv/config/bash_aliases"
 
 # cleanup
+mkdir -p /vvv
 mkdir -p /vagrant
 rm -rf /vagrant/failed_provisioners
 mkdir -p /vagrant/failed_provisioners
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ permalink: /docs/en-US/changelog/
 * Install WP-CLI doctor package ( #2051 )
 * Enhanced database backup terminal output ( #2256 ) 
 * Sites with no `hosts` defined will now default to `{sitename}.test` ( #2267 ) 
+* Avoid `apt-get update` to run on the main provisioner on every provision ( #2272 )
 
 ### Deprecations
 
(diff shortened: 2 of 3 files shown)


tomjn:
I was thinking more updating the packages than the sources, but this is needed too. We'll need to account for utilities as well and the VVV version though, the date isn't enough

tomjn:
I figure we store an SHA of a string of the version/utilities/etc

Mte90:
Wondering if we can extend this also to other utilities or command like npm or wp-cli download.

msaggiorato:
> Wondering if we can extend this also to other utilities or command like npm or wp-cli download.

I'll think about this, i'm sure we can do this kind of thing for everything.

provision/provision.sh
```diff
@@ -12,6 +12,7 @@
 . "/srv/config/bash_aliases"
 
 # cleanup
+mkdir -p /vvv
```
tomjn, line 15: imo this should go in `/vagrant` and the move to `/vvv` should be a separate issue where we symlink it to `/vagrant`, that way we aren't spreading things across multiple folders

msaggiorato:
I'll turn this into a draft, since I don't have time to work on this on the next couple days.

tomjn:
I think the todo list of things here should be:

1. the `/vvv` older should be a separate issue and PR, ideally using a subfolder of `/srv/` instead that's VM only ( unmapped )
2. we can't rely on just time
    - if the config files changes we should run `apt`
    - if the config file says always run `apt` then always run it
    - if the VVV version has changed we should run `apt`

So we should compare an SHA hash of:

```
{VVV version}{utility src git hashes}{current date}
```

We can then provide finder control in the future by allowing the `config` to specify how granular the date is, e.g. `2020/07` rather than `2020/07/01` for monthly updates

We should also look into appending the VVV commit hash to the version in the vagrant file so that people on `develop` get updates on `git pull`.

At that point we're probably good to extend this to the entire main provisioner

tomjn:
Additionally, only succesful provisions should update the hash so that it doesn't stay stuck in a broken state

Mte90:
any updates @msaggiorato

PR #2285, opened 2020-10-28
Fix syhaml not working

msaggiorato:
Trying a fresh provision today, and shyaml wasn't installing.

After this, it provisions and work fine. Not sure of the details of the error (likely a change upstream somewhere).

### DIFF
--- a/provision/core/vvv/provision.sh
+++ b/provision/core/vvv/provision.sh
@@ -22,6 +22,7 @@ function vvv_register_packages() {
     vim
     colordiff
     python-pip
+    python-setuptools
     lftp
 
     # ntp service to keep clock current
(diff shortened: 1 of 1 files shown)


Mte90:
Strange because there are no new release of that package since 2018 https://github.com/0k/shyaml/releases

Also setuptools and wheel are needed to generate a pip package.

msaggiorato:
Super strange indeed... Maybe I was hit with a bad cache somewhere?

Without doing this provision was definetely not working.

It's also definetely strange that nobody reported this yet.

Mte90:
I am testing a provision and I don't see any error about shyaml.

msaggiorato:
Are you trying completely from scratch? Destroying and provisioning?

Can you share the logs around shyaml installation?

Mte90:
I did a provision on my machine, now trying with a vagrant destroy.
I will let you know.

Mte90:
Ok so in a new box I got that error but when started working on custom-template and not during VVV.

The issue is one of the dependence of shyaml that require those packages.

Mte90 (approved):

PR #2310, opened 2020-12-03
Only install the PCov version we need, and other PHP install tweaks

tomjn:
Use the PHP specific version of pcov instead of installing all of them, and only copy configs if the folder exists

This needs testing, once it's confirmed to work for an existing and a new VVV instance, we can port this to the utilities.

## Checks

* [ ] I've tested this PR


Mte90 (dismissed):
Tested and there are no errors

PR #2312, opened 2020-12-04
Prevent VM creation if using VirtualBox + root/sudo

tomjn:
In theory this prevents the creation of the VM by exiting early if the user is root. It needs further testing however

## Checks

* [ ] I've tested this PR


Mte90:
I cannot test it on linux because I did't gave sudo permission to vagrant at all.

tomjn:
`sudo vagrant up` in a fresh VVV checkout of this branch should be all that's needed, or running the `vagrant` command as `root`. This PR doesn't influence the guest VM

Vagrantfile
```diff
@@ -375,6 +376,18 @@ Vagrant.configure('2') do |config|
 
   # Configurations from 1.0.x can be placed in Vagrant 1.1.x specs like the following.
   config.vm.provider :virtualbox do |v|
+
+    unless Vagrant::Util::Platform.windows?
+      if Process.uid == 0
+        machine_id_file=Pathname.new(".vagrant/machines/default/virtualbox/id")
```
msaggiorato, line 382: What about other providers (VMware or parallels)?

tomjn: We don't know how to do this check for parallels or VMWare, or even if the `sudo` problem affects those users

msaggiorato: Should we do a check for the current provider then? And skip the warning? Or at least make it different?

Mte90: maybe is the case to do a check for the provider to avoid errors or crashes?

tomjn: I don't know that such a check would avoid errors and crashes, or what such a check would look like. I'm not even sure what the concern is?

Mte90:
I cannot test it because my machine block sudo on other commands.

Vagrantfile
evertiro: The problem is a design choice in VirtualBox (i.e.: it's intentional). Parallels doesn't have the same problem (though I can't swear it has _no_ problem if started as root, I only did a basic function test). Can't speak to VMWare.

evertiro: Actually, I'd argue that we should prevent using `sudo` with VirtualBox on all platforms. It's not _intended_ to be launched as root, so even if some OSs handle it better than others, it's still using it in a way other than intended. Even beyond that, I'd argue that we should at least discourage using `sudo` at all... there's just not really a good reason to do it, and it'll cause more issues than it solves 90% of the time.

tomjn: the problem is that this would mean using VVV on Windows would become extremely difficult as you couldn't launch it using an elevated prompt, which is necessary if you want hostnames and symlinks

evertiro: OK, so flip the conditions... make it always throw that error unless platform == windows and provider != virtualbox

evertiro: There's no *good* reason to use `sudo` on any platform but Windows... making Windows-based solutions other than VirtualBox the exception rather than the rule.

evertiro:
> I cannot test it because my machine block sudo on other commands.

I'm really curious about this... what environment actively blocks `sudo` by default? And if it's not a default, why would you intentionally put yourself in that position?

Vagrantfile
tomjn: `Unless` is used because the Codacy linter says it's better

evertiro: So? `Unless` can be applied either direction...

tomjn: That if not not windows

evertiro: Wait never mind, you're already catching it for non-windows systems elsewhere. You're good.

evertiro (commented):
Sorry, that wasn't supposed to be opening a review, misclicked.

Vagrantfile
tomjn: Resolving this because:
 - the code is inside the virtualbox config section so other providers aren't affected
 - it doesn't run on Windows

Vagrantfile
```diff
@@ -329,6 +330,18 @@ ENV['LC_ALL'] = 'en_US.UTF-8'
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   # VirtualBox
   config.vm.provider :virtualbox do |v|
+    unless Vagrant::Util::Platform.windows?
+      if Process.uid == 0
+        machine_id_file=Pathname.new(".vagrant/machines/default/virtualbox/id")
```
Mte90, line 335: I tested it now after executing it on Linux. As normal user that file doesn't exist (also when I execute the machine).
On my machine if I try to execute vagrant with sudo I get an error that my user is not allowed to execute it as root.

tomjn: hmm what's the equivalent?

Mte90: I have `.vagrant.d` folder but inside doesn't have any machines folder.

tomjn: that's with a vagrant VM running? or not created? How does it know the ID of the VM? What's the layout/structure?

Mte90: I have that folder just installing it and includes the various VM and plugins for that user but is not helpful to see if the machine is under root or another user.

It isn't more easy to check what kind of user is executing the vagrantfile? https://stackoverflow.com/questions/52747306/how-to-check-for-root-prvileges-from-the-user-in-ruby

tomjn: To be clear, we are _**only**_ interested in exiting when creating a _new_ VM, this does not affect VMs that have already been created this way

tomjn: it should be checking the local `.vagrant` folder not the folder in the homedir, so for a VVV install at `~/vvv-local` it should look inside `~/vvv-local/.vagrant` not `~/.vagrant.d`. I too have a `~/.vagrant.d/` folder on MacOS that looks the same as that

We can't use the presence of `~/.vagrant.d/data/machine-index/index` though as it's a JSON file listing all known machines, unlike the `id`, whereas the `id` file is just the ID of the machine.

So if you have not created a VM for that VVV install then it is expected that the file and folder does not exist.

tomjn: The logic is:

 ```
if {vvvdir}/.vagrant/machines/default/virtualbox/id exists
    the VM exists, don't abort
else
    it doesnt exist therefore no VM exists
```

Mte90: I misconfused thinking that was a .vagrant folder in the homedir instead the VVV folder itself.

Mte90 (approved):

PR #2318, opened 2020-12-15
Modify/provisioner tweaks

tomjn:
This builds on the work in #2317 and extends the improvements across the main provisioners. It also introduces the `set -eo pipefail` in a lot of main provisioners, reduces output when it isn't needed, and fixes several installation issues related to composer and shyaml

Screenshot 2020-12-15 at 19 08 44
Screenshot 2020-12-15 at 19 08 58
Screenshot 2020-12-15 at 19 09 17

## Checks

* [x] I've tested this PR

### DIFF
--- a/config/homebin/vagrant_provision
+++ b/config/homebin/vagrant_provision
@@ -55,6 +55,7 @@ else
 	echo -e "${RED} ──────────────────────────────────────────────────${CRESET}"
 	echo -e " "
 	echo -e "${RED}    ⚠ oh noes :'( ⚠"
+	echo -e "${RED}"
 	echo -e "${RED}   ▄▀▀▀▄▄▄▄▄▄▄▀▀▀▄ ${CRESET}"
 	echo -e "${RED}   █▒▒░░░░░░░░░▒▒█    ⚠ Provisioning${CRESET}"
 	echo -e "${RED}    █░░${E}░░░░░${E}░░█     ⚠ incomplete!${CRESET}"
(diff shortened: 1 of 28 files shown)


config/homebin/vagrant_provision
```diff
@@ -65,6 +68,6 @@ else
 	echo -e "${RED} ⚠ The failed provisioners left these logs:"
 
 	# List failed provisioners, and pipe them through xargs to print them in a formatted way.
-	ls -A /vagrant/failed_provisioners/ | xargs -I "LOGFILE" echo -e "   - log/provisioners/${date_time}/LOGFILE.log"
+	ls -A /vagrant/failed_provisioners/ | xargs -I "LOGFILE" vvv_error "   - <b>log/provisioners/${date_time}/LOGFILE.log</b>"
```
Mte90, line 71: this is an html tag, why on a log file?

Mte90: Ok I saw in another file that will be replaced, maybe add something in our doc website?

tomjn: At the moment we don't document any of these functions so we'd want to figure out a solution for that, something like PHPDoc for shell scripts

Mte90: https://github.com/reconquest/shdoc There is this that is for bash, maybe is for another ticket?

tomjn: That looks like a good option :) Can you make a new ticket for it?

PR #2328, opened 2020-12-27
WIP: Add /srv/vvv

tomjn:
Switch from `/vagrant` to `/srv/vvv`

## Checks

* [ ] I've tested this PR


Mte90 (changes requested):
Screenshot_20201230_162403

It is crashing to me the provision with that branch.

provision/provision-helpers.sh
```diff
@@ -17,11 +17,18 @@ export BLUE="\033[0;38;5;4m" # 33m"
 export PURPLE="\033[0;38;5;5m" # 129m"
 export CRESET="\033[0m"
 
-
-VVV_CONFIG=/vagrant/vvv-custom.yml
-if [[ -f /vagrant/config.yml ]]; then
-	VVV_CONFIG=/vagrant/config.yml
-fi
+configs=(
```
msaggiorato, line 20: Should we turn this snippet into a function (including the export below the change here)?

This would reduce some code duplication, and potentially improve maintainability (same code is implemented in the test suite on provision/tests/provisioners.sh )

msaggiorato (changes requested):

provision/provision-helpers.sh
tomjn: I'd like that but we'll need this in utilities etc for a while eitherway

msaggiorato: This file we're commenting on is already on provision helpers. Utilities can have their own implementations that would override this exported value.

On that topic, at some point we'd need to have some sort of versioning for easier introduction of breaking changes to utilities. Eg: utilities v2, only works with VVV v3.6, or something on those lines. Might be troublesome to implement, but will allow faster development on the utilities repo.

tomjn: I think a polyfill script is probably best for that, and a minimum VVV version helper function

Mte90 (dismissed):
Tested and no errors

provision/provision-helpers.sh
Mte90: Maybe we can open a ticket for that and move on this pr?

Mte90:
Screenshot_20210406_134256
I got this I guess that this works for new VM.

tomjn:
@Mte90 that shouldn't have happened, looks like there's still work to be done here

Mte90:
As I can see that folder get root permissions instead the others have vagrant/vagrant:

Screenshot_20210407_122030

I think that a chown when the folder is created should be enough @tomjn

Mte90:
I think that is the case that someone else test that PR.

evertiro (changes requested):
No. This still fails with the SCP error (actually, the only change made after the SCP error was discovered was a merge).

Mte90:
I will do some tests for this to moving on, I just forgot it...

Mte90:
So I tested it and a `vagrant provision` generate this.

```
 ⚠ ! Running additional Vagrant code in Customfile located at /home/www/VVV/Customfile
 ⚠ ! Official support is not provided for this feature, it is assumed you are proficient with vagrant

 ⚠ ! Finished running Customfile, resuming normal vagrantfile execution

There are errors in the configuration of this machine. Please fix
the following errors and try again:

shell provisioner:
* Shell provisioner `args` must be a string or array.
```

tomjn:
converted this to a draft, I want to see the end result of the vagrantfile changes before adjusting this

PR #2329, opened 2020-12-27
Docker compatibility adjustments

tomjn:
This attempts to make our provisioners a little more comfortable during docker builds. Note that this PR is not adding docker support.

## Checks

* [ ] I've tested this PR

### DIFF
--- a/provision/core/mariadb/provision.sh
+++ b/provision/core/mariadb/provision.sh
@@ -96,7 +96,7 @@ function mysql_setup() {
   vvv_info " * Copied /srv/config/mysql-config/root-my.cnf          to /home/vagrant/.my.cnf"
 
   if [ "${VVV_DOCKER}" != 1 ]; then
-  check_mysql_root_password
+    check_mysql_root_password
   fi
 
   # MySQL gives us an error if we restart a non running service, which
@@ -132,4 +132,6 @@ function mysql_setup() {
 }
 export -f mysql_setup
 
-vvv_add_hook after_packages mysql_setup 30
+if [ "${VVV_DOCKER}" != 1 ]; then
+  vvv_add_hook after_packages mysql_setup 30
+fi
--- a/provision/core/mariadb/provision.sh
+++ b/provision/core/mariadb/provision.sh
@@ -81,12 +81,9 @@ SQL
 
 function mysql_setup() {
   # If MariaDB/MySQL is installed, go through the various imports and service tasks.
-  local exists_mysql
-
-  exists_mysql="$(service mysql status)"
-  if [[ "mysql: unrecognized service" == "${exists_mysql}" ]]; then
+  if ! command -v mysql &> /dev/null; then
     vvv_error " ! MySQL is not installed. No databases imported."
-    return
+    return 1
   fi
   vvv_info " * Setting up database configuration file links..."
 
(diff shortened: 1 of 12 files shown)


tomjn:
I've some other adjustments but they're experimental, these changes should benefit VirtualBox provisioning by fixing problems that don't prevent VVV from running under vagrant

provision/core/mariadb/provision.sh
```diff
@@ -98,19 +95,20 @@ function mysql_setup() {
   chmod 0644 "/home/vagrant/.my.cnf"
   vvv_info " * Copied /srv/config/mysql-config/root-my.cnf          to /home/vagrant/.my.cnf"
 
-  check_mysql_root_password
+  if [ "${VVV_DOCKER}" != 1 ]; then
```
msaggiorato, line 98: This whole function would only run if we're not on docker, so i'd say this check is unnecessary.

provision/core/vvv/provision.sh
```diff
@@ -71,7 +71,10 @@ function cleanup_vvv(){
   mv /tmp/hosts /etc/hosts
 }
 export -f cleanup_vvv
+
+if [ "${VVV_DOCKER}" != 1 ]; then
 vvv_add_hook finalize cleanup_vvv 15
```
msaggiorato, line 76: Indentation?
```suggestion
  vvv_add_hook finalize cleanup_vvv 15
```

provision/core/wp-cli/provision.sh
```diff
@@ -17,20 +17,26 @@ function wp_cli_setup() {
...
+    curl -s https://raw.githubusercontent.com/wp-cli/wp-cli/master/utils/wp-completion.bash -o /srv/config/wp-cli/wp-completion.bash
+    chown vagrant /srv/config/wp-cli/wp-completion.bash
   else
     vvv_info " * Updating wp-cli..."
     wp --allow-root cli update --nightly --yes
     vvv_success " * WP CLI Nightly updated"
   fi
-  vvv_info " * Installing WP CLI doctor sub-command"
-  wp --allow-root package install git@github.com:wp-cli/doctor-command.git
+  if [ "${VVV_DOCKER}" != 1 ]; then
```
msaggiorato, line 35: Why execute this only on non docker? Is this incompatible or useless in docker?

provision/provision-helpers.sh
```diff
@@ -85,10 +85,14 @@ export -f check_network_connection_to_host
...
 # @see check_network_connection_to_host
 function network_check() {
+  if [ "${VVV_DOCKER}" == 1 ]; then
+    return 0
+  fi
+
   # Make an HTTP request to ppa.launchpad.net to determine if
   # outside access is available to us. Also check the mariadb
   declare -a hosts_to_test=(
-    "https://ppa.launchpad.net"
```
msaggiorato, line 91: Is there any reason for this change on this particular PR?

provision/tests/run-provisioners.sh
```diff
@@ -1,11 +1,15 @@
 #!/bin/bash
+DEBIAN_FRONTEND=noninteractive
+TZ=Europe/London
```
msaggiorato, line 3: Is UTC available as a timezone? Or is this an exact equivalent?

msaggiorato (changes requested):

provision/core/mariadb/provision.sh
tomjn: I don't understand, the check prevents it running on docker, this is desirable behaviour

provision/provision-helpers.sh
tomjn: there is no  https ppa.launchpad.net, never was.  They only  provide TLS for private PPA's

provision/tests/run-provisioners.sh
tomjn: This is copied from elsewhere, and just for the build of the docker image, I suspect `UTC` may work but I don't know

provision/core/wp-cli/provision.sh
tomjn: Docker image builds vs Docker container scripts. We should not consider `VVV_DOCKER` to be final, think of it as an experimental internal temporary thing that's necessary until a proper system is in place

Issue #2225, opened 2020-07-29
Extended DB Backup Options


tomjn:
Originally, DB backups and restores were added so that if you destroyed the VM you could create it again without data loss. Options were added

At the moment you can back up databases to SQL files, or not back up. All or nothing:

```yaml
general:
  db_backup: true
  db_restore: true
```

t's time we give people finger control over DB backup behaviours. Here's a proposed schema:

```yaml
general:
  db_backup:
    # Compress usiing gzip, db.sql becomes db.sql.gz when true
    gzip: true

    # When do we run backups?
    on:
      halt: false
      destroy: true

    # Always back up these databases
    include:
      - firstdb
      - second

    # Never back up these databases
    exclude:
      - fourth
      - fifth

    # If a database doesn't appear in the above lists, do we back it up?
    other: true

  db_restore:
    # when do we restore?
    on:
      provision: true

    # Always restore these:
    include:
      - firstdb
      - second

    # Never restore these:
    exclude:
      - fourth
      - fifth

    # Default for databases not mentioned above
    other: true
```

I don't expect all of these to be implemented at once. So this will act as the umbrella ticket for:

 - backup include list
 - backup exclude list
 - backup default
 - restore include list
 - restore exclude list
 - restore default
 - backup gzip compression
 - backup trigger controls

Mte90:
The idea is to have db backup as default just on destroy.

Mte90:
As today we have gzip and exclude list, so we need just to expand it for the other cases.

evertiro:
How would include lists work? Isn't it a safe assumption that if something isn't _excluded_, it's _included_?

Mte90:
Me too I don't think that is necessary to have both the implementation

evertiro:
I _do_ see a point to `include` for restores. In the case of restoration, it might make sense for `include` to be treated automatically as `only`.

PR #2339, opened 2021-01-25
Disable backups by default

tomjn:
Disables backup and restore of the database by  default

## Checks

* [ ] I've tested this PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ permalink: /docs/en-US/changelog/
 * SHDocs added to core provisioners
 * Improved PHP configuration file installation
 * Adds a `vagrant` command inside the virtual machine to tell users they are still inside the VM and need to exit
+* Disable backup and restore of databases by default
 
 ### Bug Fixes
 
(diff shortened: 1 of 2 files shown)


msaggiorato:
What's the reasoning for this? I can see restore being disabled by default, but why disable backups?

tomjn:
> What's the reasoning for this? I can see restore being disabled by default, but why disable backups?

Users can run these manually or turn them back on, it seems that people are provisioning with monumentally huge databases and see multi-hour provisioning as just a fact of life. A provision shouldn't take longer than 10 minutes.

When we have more granular backup options we can rethink this, but we've been saying that for a while and nobody has stepped up to implement it.

Also, the original purpose of this was so that you could `vagrant destroy` then `vagrant up` regularly, but we no longer expect this.

Mte90:
It makes sense indeed if someone does often a vagrant destroy and it isn't the majority of the users.
I agree with that change.

tomjn:
Ideally we  would  implement part  of this:

https://github.com/Varying-Vagrant-Vagrants/VVV/issues/2225

Mte90 (approved):

evertiro:
@tomjn just making sure this hasn't been forgotten. It's been almost a month since it was approved!

tomjn:
@evertiro I was never needed for this to be merged

PR #2369, opened 2021-02-22
fix(provision-site): fixing copy paste mistake, updating to search fo…

### LINKED ISSUE #2368: Provision Site Script: function is not searching for the right file when file is not present in a standard location
## Expected Behavior
Ability to place vvv-init.sh into a project within 3 folder depths of site provision root and have vvv provision script find and run it.

## Current Behavior
Code searches for vvv-init.conf instead of vvv-init.sh when the file is not present in a standard location. 

## Possible Solution
vvv_proviion_site_script was likely created from duplicating vvv_provision_site_nginx and vvv-nginx.conf was not fully replaced with vvv-init.sh some places still referenced vvv-info.conf which is incorrect.

example: /public_html/.vvv/vvv-init.sh is not found.

spenserhale:
…r vvv-init.sh instead of vvv-init.conf

vvv_proviion_site_script was likely created from duplicating
vvv_provision_site_nginx and vvv-nginx.conf was not fully replaced with
vvv-init.sh some places still referenced vvv-info.conf which is
incorrect.

BREAKING CHANGE: Previously was not searching for the right file, so now
will start provisioning if the consumer has a vvv-init.sh within their
projects that was not in /.vvv/, /provision/, / of their project.

Closes #2368 

* [x] I've tested this PR

### DIFF
--- a/provision/provision-site.sh
+++ b/provision/provision-site.sh
@@ -247,8 +247,8 @@ function vvv_provision_site_script() {
     vvv_run_site_template_script "vvv-init.sh" "${VM_DIR}"
     SUCCESS=$?
   else
-    vvv_warn " * Warning: A site provisioner was not found at .vvv/vvv-init.conf provision/vvv-init.conf or vvv-init.conf, searching 3 folders down, please be patient..."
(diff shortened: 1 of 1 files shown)


tomjn (approved):
Ah good catch!

Issue #2309, opened 2020-11-27
Internal utility to set the PHP version of VVV


Mte90:
The idea behind this is to avoid issues when we change the internal default php version.
AS we need to change various files and version in a lot of files, something that does this for us automatically can save a lot of time.

tomjn:
It will need to:

 - read in from the config file what the preferred default PHP should be
 - check that it's available
    - This means it will need to run after the utilities have run else we could declare we want 7.0 to be the default and it would not find it because it hasn't been installed yet
 - The dashboard/XHGUI/PHPinfo/etc currently use `php` upstream, but these should be changed instead to an internal upstream that uses our preferred default so that changing the default doesn't break these


The last item can be spun off into a new ticket

Mte90:
Reading now the ticket seems that we need as always that in VVV define a PHP default version in case the one in the config is not available.

Mte90:
So as today we have https://github.com/Varying-Vagrant-Vagrants/VVV/blob/4a22238a555b939076bb41b5f73a699b94afb8a2/provision/core/php/provision.sh#L5 this variable that is set as `7.4` we can think to a new config:
```
general:
  default_php_version: 8.0
```

So the code as wrote https://github.com/Varying-Vagrant-Vagrants/VVV/issues/2309#issuecomment-734812688 require some check. 

Instead how this change can affect utilities (apart the php version running as default for the webserver for them):

* https://github.com/Varying-Vagrant-Vagrants/vvv-utilities/blob/master/webgrind/provision.sh#L22 - use composer
* https://github.com/Varying-Vagrant-Vagrants/vvv-utilities/blob/master/opcache-gui/provision.sh#L7 - user composer
* https://github.com/Varying-Vagrant-Vagrants/vvv-utilities/blob/master/tideways/provision.sh#L106 - we use php 7.4 that is enforced so we can ignore it, also if it is composer used later
* https://github.com/Varying-Vagrant-Vagrants/vvv-utilities/blob/master/wpcli-dev/provision.sh#L38 - use composer
* https://github.com/Varying-Vagrant-Vagrants/vvv-utilities/blob/master/mongodb/provision.sh#L15 - we use pecl but there is a loop that install it to any php version

So a solution is to enforce the VVV's php default version as minimum for web server and also for cli.
This can open to:

```
general:
  default_php_version_cli: 8.0
  default_php_version_nginx: 7.4
```

[...]

PR #2371, opened 2021-02-23
Upgrade default PHP to v7.4

tomjn:
Changes the default PHP from 7.3 to 7.4

## Checks

* [x] I've tested this PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ permalink: /docs/en-US/changelog/
 ### Enhancements
 
 * Improve the way that PHPCS gets provisioned to avoid conflicts with composer v2 (#2357)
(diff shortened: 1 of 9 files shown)


Mte90:
As we are working on that maybe we can also implement https://github.com/Varying-Vagrant-Vagrants/VVV/issues/2309

msaggiorato:
As @Mte90 says, this new command seems like a good chance to set a specific php version as default.

It could open up a can of worms of support tho, when some specific PHP version make provisioners or stuff stop working.

tomjn:
I don't think we should conflate upgrading 7.4 and making PHP default versions configurable. They aren't the same piece of work. I also having misgivings about making the default user configurable, it's a blunt force fix that doesn't address the actual problem, an X Y problem.

benlumia007 (approved):
looks good.

PR #2377, opened 2021-03-02
Installs the missing package for Xdebug support

evertiro:
Installs the missing package for Xdebug support

## Checks

* [x] I've tested this PR

### DIFF
--- a/provision/core/php/provision.sh
+++ b/provision/core/php/provision.sh
@@ -53,6 +53,7 @@ function php_register_packages() {
 
   # XDebug
   VVV_PACKAGE_LIST+=(
+    "php${VVV_BASE_PHPVERSION}-xdebug"
     php-xdebug
   )
 }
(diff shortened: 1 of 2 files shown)


provision/core/php/provision.sh
```diff
@@ -53,6 +53,7 @@ function php_register_packages() {
 
   # XDebug
   VVV_PACKAGE_LIST+=(
+    "php${VVV_BASE_PHPVERSION}-xdebug"
```
tomjn, line 56: hmmm this shouldn't be needed what with the line below, but I don't see the harm, I'll test this shortly

evertiro: Without that line, clone VVV, run `vagrant up`, ssh to it and run `xdebug_on`. It will give the error below. The `php-xdebug` package doesn't seem to actually function as a metapackage or version-agnostic installation as you might expect.

Screenshot-20210302-13:41:17

PR #2381, opened 2021-03-05
Modify/better php debug switcher

tomjn:
This greatly improves the `switch_php_debugmod` script so that:

 - it no longer tries to disable modules that are already disabled
 - it no longer tries to enabel modules that are not installed
 - it enables/disables modules on a per PHP basis to avoid warnings when a module does not support all installed versions ( e.g. pcov did not support 7.0 and below, and would generating warnings if those were installed )
 - improves output so it's clearer to users what's happening and which versions of PHP a debug module is active on

## Checks

* [ ] I've tested this PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ permalink: /docs/en-US/changelog/
 * SHDocs added to core provisioners
 * Improved PHP configuration file installation
 * Adds a `vagrant` command inside the virtual machine to tell users they are still inside the VM and need to exit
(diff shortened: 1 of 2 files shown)


tomjn:
Screenshot 2021-03-05 at 17 40 53

tomjn:
Do you have PHP 8.0 setup right? `conf.d` should exist

Mte90:
No my VVV instance doesn't have php 8. Just 71/72/73.

tomjn:
That'll be why then, does `/etc/php/8.0/` exist?

Issue #2386, opened 2021-03-09
Update to support Ubuntu 20.04


evertiro:
**Is your feature request related to a problem? Please describe.**

VVV is currently built around Ubuntu 18.04. Ubuntu 18.04 is getting close to ESM updates only, and 20.04 is the current LTS.

**Describe the solution you'd like**

We should start looking at making VVV work with 20.04 with the goal of switching over completely in the near future.

**Tested working on M1:**
- [x] Dashboard
- [x] MailHog
- [x] Memcached
- [x] Opcache
- [x] PCov
- [x] PHP Info
- [x] phpMyAdmin
- [x] Webgrind
- [x] WP-CLI
- [x] Xdebug

**Not working on M1:**
- [ ] Tideways (`mongodb` install fails)

evertiro:
@tomjn You might as well just assume this is my ticket. Since the existing ARM64 boxes are all 20.04+ I'm going to have to figure out what changes are necessary for 20.04 anyway unless I want to build my own box for 18.04... which seems counterintuitive given its support window. I'll just tackle this first, then move on to dealing with Silicon once this is a bit better tested.

evertiro:
Note, the VVV PPA needs to be updated for focal

evertiro:
Previous comment is erroneous; caching issue. However...

```
default:  ▷ Running after_packages hook
default:  * Installing Shyaml for bash provisioning..
default: sudo: pip: command not found
```

evertiro:
Ubuntu dropped default python2 support in 20.04. Switching to python3/pip3 fixes the above issue.

evertiro:
Now getting the following:

```
  - Installing dealerdirect/phpcodesniffer-composer-installer (v0.7.1): Loading from cache
 Extracting archivePlugin installation failed (include(/srv/www/phpcs/vendor/dealerdirect/phpcodesniffer-composer-installer/src/Plugin.php): failed to open stream: No such file or directory), rolling back
  - Removing dealerdirect/phpcodesniffer-composer-installer (v0.7.1)

Installation failed, reverting ./composer.json to its original content.

                                                                               
  [ErrorException]                                                             
  include(/srv/www/phpcs/vendor/dealerdirect/phpcodesniffer-composer-installe  
  r/src/Plugin.php): failed to open stream: No such file or directory          
                                                                               

Exception trace:
 () at phar:///usr/local/bin/composer/vendor/composer/ClassLoader.php:444
[...]

Issue #2385, opened 2021-03-09
Support for Apple Silicon devices - Parallels


evertiro:
**Is your feature request related to a problem? Please describe.**

Currently, VVV only officially supports the use of VirtualBox and Hyper-V. Vagrant, however, supports VMWare and Parallels as well, both of which have at least a work-in-progress build for Silicon devices. As of now, the only reason Silicon devices can't be used with VVV is that nobody on the dev team has one. I have one. I want to use VVV on it. I'm going to see what we have to do to make it happen.

**Describe the solution you'd like**

I want to make VVV work on Silicon devices.

**Describe alternatives you've considered**

Please don't make me rehash the long debate on remote environments I just had on Slack...

**Additional context**

This ticket is being opened so that a) I have a place to track my progress on this little project, and b) so that other users who have access to Silicon-based devices have a point of reference on where VVV currently stands with Silicon support. Please note that this is (for now) my personal "pet project" and not an officially supported feature by VVV.

tomjn:
Notes:

 - VMWare and Parallels have technical previews based on using the Apple hypervisor
 - VMWare requires a $199 yearly license as well as purchasing the vagrant plugin from Hashicorp at $79.99 for a grand total of **$279.98**
 - Parallels provides a free plugin and has a $79.99 yearly license for the Business/Pro version required

Parallels is significantly cheaper

We'd need to use an Arm64 box, as well as test that our provisioners work with Arm64 packages. A raspberry Pi should be capable of testing most of this by running the provisioners directly in an Ubuntu 18 type OS

It also looks like there are people attempting to use Parallels for vagrant work on M1 macs:

https://github.com/Parallels/vagrant-parallels/issues/382

Albeit with various issues

tomjn:
Free options so far seem to be to hope VirtualBox adds Arm64 support ( extremely unlikely ), or that we containerize VVV and use docker ( already a longterm goal )

evertiro:
I'm going to read up on the issues present on Parallels and hopefully start there given the price difference. I'd like to eventually be able to green flag both options as working with VVV.

Also... I want to know more about your docker goals. That may get added to my todo list too. I already use Docker a fair bit in my day-to-day work.

evertiro:
[...]

PR #2390, opened 2021-03-15
First pass at M1 support

evertiro:
This is a first-pass at a build working on the Apple M1 (#2385). It mostly works, with a few caveats:

* This is dependent on #2386. There are no 18.x builds for the M1. I'm not building one to support something we shouldn't be using anyway.
* ~~Mailhog still doesn't work. My change to the mailhog provisioner is a bandaid at best. The file I'm linking to is an ARM build, not an ARM64 build. It makes vagrant happy enough to not error out, but won't actually run on M1. I'm looking into building it on the M1.~~ This is both invalid and outdated... we needed aarch64, not arm64. Neither existed. Both are now hosted on my GitHub for `mhsendmail` and `MailHog`.
* It should go without saying, but this assumes that you are using Parallels. This also means that in order to run this, you need to be using Parallels with a Business/Pro license. Once I figure out the mailhog issues, I'll look into VirtualBox on M1.

Since the only two points that this changes only run if you're on an ARM64 machine, this shouldn't impact users on any other architecture.

## Checks

* [x] I've tested this PR


provision/core/git/sources-ubuntu-bionic.list
```diff
@@ -1,3 +1,3 @@
 # git lfs (large file storage plugin for git)
-deb https://packagecloud.io/github/git-lfs/ubuntu/ bionic main
-deb-src https://packagecloud.io/github/git-lfs/ubuntu/ bionic main
```
tomjn, line 3: This change isn't necessary, and wouldn't apply anyway, a `sources-ubuntu-focal.list` is the solution

provision/core/mailhog/provision.sh
```diff
@@ -10,13 +10,22 @@ function mailhog_setup() {
 
   if [[ ! -e /usr/local/bin/mailhog ]]; then
     vvv_info " * Installing MailHog"
-    curl --silent -L -o /usr/local/bin/mailhog https://github.com/mailhog/MailHog/releases/download/v1.0.0/MailHog_linux_amd64
+
+    if [[ "aarch64" == $(uname -m) ]]; then
+      curl --silent -L -o /usr/local/bin/mailhog https://github.com/evertiro/MailHog/releases/download/v1.0.0-M1/MailHog_linux_arm64
+    else
+      curl --silent -L -o /usr/local/bin/mailhog https://github.com/mailhog/MailHog/releases/download/v1.0.0/MailHog_linux_amd64
+    fi
```
tomjn, line 18: this could be separated into its own PR, Ubuntu 18 runs on Arm too ( e.g. [Raspberry Pis](https://maker.pro/raspberry-pi/projects/install-ubuntu-18044-lts-on-your-raspberry-pi-board)  )

provision/core/mailhog/provision.sh
```diff
@@ -10,13 +10,22 @@ function mailhog_setup() {
...
     vvv_success " * Mailhog binary installed"
   fi
   if [[ ! -e /usr/local/bin/mhsendmail ]]; then
     vvv_info " * Installing MHSendmail"
-    curl --silent -L -o /usr/local/bin/mhsendmail https://github.com/mailhog/mhsendmail/releases/download/v0.2.0/mhsendmail_linux_amd64
+    if [[ "aarch64" == $(uname -m) ]]; then
+      curl --silent -L -o /usr/local/bin/mhsendmail https://github.com/evertiro/mhsendmail/releases/download/v0.2.0-M1/mhsendmail_linux_arm64
+    else
+      curl --silent -L -o /usr/local/bin/mhsendmail https://github.com/mailhog/mhsendmail/releases/download/v0.2.0/mhsendmail_linux_amd64
+    fi
```
tomjn, line 28: see above comment

provision/core/vvv/sources-ubuntu-bionic.list
```diff
@@ -1,3 +1,3 @@
 # VVV mirror packages
-deb http://ppa.launchpad.net/varying-vagrant-vagrants/php/ubuntu bionic main
-deb-src http://ppa.launchpad.net/varying-vagrant-vagrants/php/ubuntu bionic main
+#deb http://ppa.launchpad.net/varying-vagrant-vagrants/php/ubuntu bionic main
+#deb-src http://ppa.launchpad.net/varying-vagrant-vagrants/php/ubuntu bionic main
```
tomjn, line 3: there's a focal source now too, and this file isn't installed on focal anyway

provision/core/git/sources-ubuntu-bionic.list
evertiro: I don't know where that one came from...

provision/core/mailhog/provision.sh
evertiro: Agreed

evertiro: This is resolved with #2423

provision/core/mailhog/provision.sh
evertiro: This is resolved with #2423

evertiro:
The M1-specific stuff from this is now in `develop`, the rest is covered by other PRs.

PR #2392, opened 2021-03-15
fix(noroot): on utilities

Mte90:
Tested and works

## Checks

* [x] I've tested this PR


tomjn:
This looks good, does it handle when the files are already owned by `root` though? We don't want provisioning to halt because it can't write to those files

Mte90:
> This looks good, does it handle when the files are already owned by `root` though? We don't want provisioning to halt because it can't write to those files

I don't know about this case as I was destroying every time the VM.

tomjn (approved):

tomjn:
Testing confirms it's not an issue

PR #2401, opened 2021-03-18
Working through PHP webroot permissions

### LINKED ISSUE #2398: Parallels: PHP can't write to webroot
Unfortunately, this could be caused by a number of different things, so this issue is going to be a long, winding journey instead of a quick sprint.

The underlying issue is that on my current setup, I can deploy a WordPress site, but any operation inside WordPress that involves direct filesystem manipulation fails. This includes installation of plugins, themes, uploading media, etc. However... the issue only exists in the web interface. Everything has the correct permissions (`vagrant:www-data` and `755`), the directories can be written to by the vagrant user through ssh and through the shared directory on the host. Running a plugin install through `wp` succeeds. Possible underlying causes:

- I am running on a MacBook Pro with an M1 chip and therefore using the Apple hypervisor
- As a result, I am unable to use VirtualBox and am working on a WIP Parallels build (#2385) 
-  There is no M1-compatible build of Ubuntu 18.04, so I'm also using the WIP 20.04 branch (#2386)

EDIT: Since this only seems to happen on M1, I think we can rule out 20.04 itself. However, there could be an issue with the specific M1 image we're using.

To clarify, this means the following need to be tested:

- [x] Does the problem exist on a Parallels build running under 20.04 Intel
  - If not, the problem lies in the Parallels Tech Preview for M1
- [x] Does the problem exist on a VirtualBox build running under 20.04 Intel
  - If not, the problem lies in the Parallels 20.04 build
  - ~~If so, the problem lies in 20.04 itself~~

evertiro:
@tomjn This was the fix for my issue with WordPress requiring SFTP to upload things. I can make this only apply when installing on an M1, but this feels like a core bug.

Fixes #2398

## Checks

* [x] I've tested this PR


tomjn:
hmmm this shouldn't cause issues, it's probably a difference between the boxes. Could we adjust the `www-data` user so it can write those files?

evertiro:
@tomjn I suspect (I can't check at the moment) that the official box we're using for most builds is giving the `www-data` group write access to the webroot. The M1 build seems to _only_ allow the owner any access to the webroot. I can't think of any reason the above change would impact anything in a local environment, but it does fix the issue on my end without changing the PHP user.

evertiro:
My previous attempt works for directories created during provisioning, but directories created by plugins are created with the wrong permissions and therefore it fails. Reverting to my previous proposal... thus far it's the only thing that's 100% consistently worked on M1.

tomjn:
The `www-data` user should be in the `www-data` group which owns those folders, so there should be no issue. Have you raised this with Parallels support?

evertiro:
The `www-data` user is in the appropriate group. Still an issue. Interestingly, if I create a folder outside of the webroot and set its ownership to `www-data:www-data` it works.

evertiro:
Fixed, major props to @benlumia007... this one's been driving me nuts for ages.

evertiro:
@tomjn We're ready for an official review I believe.

tomjn (approved):

PR #2402, opened 2021-03-19
Better site provision messaging, and fixed git cloning

tomjn:
expand and improve the logging in the site provisioner, with clearer messages and new messages

## Checks

* [x] I've tested this PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,10 +9,14 @@ permalink: /docs/en-US/changelog/
 ## 3.7.x ( 2021 )
 ### Enhancements
 
-* 
+* Improved site provisioning messages
+
 ### Bug Fixes
 
-* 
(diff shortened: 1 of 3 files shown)


tomjn:
This PR also fixes 2 problems with git cloning the main site template repo:

 - it would reset all changes _then_ pull down the remote branch, which meant if the user had committed anything locally by acccident, it caused issues rather than being overwritten
 - if the `origin` remote was ever changed then VVV would be unaware of this and pull down the new remote. A user had done this and the remote was a WP theme
 - `vvv_error` did not print out error messages in site provisioners

Good provisioning:

Screenshot 2021-03-19 at 14 58 29

Provisioning with a changed remote URL:

Screenshot 2021-03-19 at 15 04 33

Mte90 (dismissed):

PR #2412, opened 2021-03-21
Upgrade to Python3 pip3/setuptools

### LINKED ISSUE #2408: Switch from pip to pip3
We should use pip3 to install `shyaml` instead of `pip`. This will help with compatibility as `pip` is not available on Ubuntu 20, but `pip3` is.

tomjn:
closes #2408

## Checks

* [ ] I've tested this PR


tomjn:
So this fails on Ubuntu 18:

```
vagrant@vvv:~$ sudo apt-get install python3-pip python3-setuptools
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 python3-pip : Depends: python-pip-whl (= 9.0.1-2) but 9.0.1-2.3~ubuntu1.18.04.4 is to be installed
               Recommends: python3-dev (>= 3.2) but it is not going to be installed
               Recommends: python3-wheel but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
```

It seems that to fix this, `python-pip-whl` needs to be removed so that the appropriate dependency can be installed. Cleanup routines for the old packages are needed for this

tomjn:
This needs testing on existing VVV instances to confirm it upgrades from pip to pip3 correctly. I've added a package removal list to uninstall the old packages but I'm not sure if it removes the dependencies that are causing issues

Mte90 (approved):
Tested and no errors for me on upgrading a machine

PR #2429, opened 2021-04-03
Better Mailhog download handling

tomjn:
When it fails, say so

## Checks

* [ ] I've tested this PR

### DIFF
--- a/provision/core/mailhog/provision.sh
+++ b/provision/core/mailhog/provision.sh
@@ -14,9 +14,13 @@ function mailhog_setup() {
     if [[ "aarch64" == $(uname -m) ]]; then
       mailhog_bin="https://github.com/evertiro/MailHog/releases/download/v1.0.0-M1/MailHog_linux_arm64"
     fi
-    curl --silent -L -o /usr/local/bin/mailhog ${mailhog_bin}
-    chmod +x /usr/local/bin/mailhog
-    vvv_success " * Mailhog binary installed"
+    if curl --silent -L -o /usr/local/bin/mailhog ${mailhog_bin}; then
+      chmod +x /usr/local/bin/mailhog
+      vvv_success " * Mailhog binary installed"
+    else
+      vvv_error " ! MailHog failed to download"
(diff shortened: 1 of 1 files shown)


provision/core/mailhog/provision.sh
```diff
@@ -14,19 +14,27 @@ function mailhog_setup() {
     if [[ "aarch64" == $(uname -m) ]]; then
       mailhog_bin="https://github.com/evertiro/MailHog/releases/download/v1.0.0-M1/MailHog_linux_arm64"
     fi
-    curl --silent -L -o /usr/local/bin/mailhog ${mailhog_bin}
-    chmod +x /usr/local/bin/mailhog
-    vvv_success " * Mailhog binary installed"
+    if curl --silent -L -o /usr/local/bin/mailhog ${mailhog_bin}; then
```
msaggiorato, line 17: I'd do a check on $? Not being 0, rather than the curl inside the if.

msaggiorato (changes requested):

provision/core/mailhog/provision.sh
Mte90: this is something that we can implement as function to use also to download other stuff? like for vvv utilities?

tomjn: if we checked `$?` then it would fail before the if check because of thee bad error code

tomjn: As for a generic function, I think this is simple enough already, using a function would still require an if check for a message. It's out of the scope of this PR

msaggiorato: > if we checked `$?` then it would fail before the if check because of thee bad error code

Not sure I follow. Why would it fail?

tomjn: because bad exit codes end the provisioner, we have `set -eo pipefail` at the top of the file

evertiro (approved):
I'm ok with this, it works and looks pretty

Issue #2437, opened 2021-04-08
Improve Composer installation


evertiro:
Composer tries to run a self-update twice in different ways. This may result in some unnecessary slowness on 18 series machines, but the first run fatal errors on 20.x. Also, some minor cleanup.

evertiro:
Note: This also impacts the M1 box we're currently using.

PR #2438, opened 2021-04-08
Cleanup composer installation

evertiro:
Cleans up the composer installation #2437 

## Checks

* [x] I've tested this PR

### DIFF
--- a/.github/workflows/vvv-provisioning.yml
+++ b/.github/workflows/vvv-provisioning.yml
@@ -30,13 +30,18 @@ jobs:
           sudo apt-get -q --autoremove --purge remove php*
           sudo apt-get -q autoclean
 
+          # remove pre-installed composer
+          if [ -f /usr/bin/composer ]; then
+            sudo rm -f /usr/bin/composer
+          fi
+
           # create vagrant user
           sudo groupadd -g 2000 vagrant
           sudo useradd -u 2000 -g vagrant -m vagrant
 
           # vvv_symlink function to sumulate synced folders
           function vvv_symlink() {
-            if [ ! -d "${1}" ]; then 
+            if [ ! -d "${1}" ]; then
               sudo mkdir -p "${1}"
             fi
             sudo chown -R vagrant:vagrant "${1}"
--- a/provision/core/composer/provision.sh
+++ b/provision/core/composer/provision.sh
@@ -50,9 +50,11 @@ function composer_setup() {
     COMPOSER_HOME=/usr/local/src/composer noroot composer --no-ansi global require --prefer-dist --no-update --no-progress --no-interaction phpunit/phpunit:^7.5
   fi
 
-  vvv_info " * Updating global composer packages..."
-  COMPOSER_HOME=/usr/local/src/composer noroot composer --no-ansi global update --no-progress --no-interaction
-  vvv_success " * Global composer package update completed"
+  if [ -f /usr/local/src/composer/composer.json ]; then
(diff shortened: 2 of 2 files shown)


tomjn (dismissed):

evertiro:
@tomjn I think this should also include the provisioner fix. Even though the provisioner is only failing on 20.x, it _is_ a composer issue.

.github/workflows/vvv-provisioning.yml
```diff
@@ -30,13 +30,18 @@ jobs:
           sudo apt-get -q --autoremove --purge remove php*
           sudo apt-get -q autoclean
 
+          # remove pre-installed composer
+          if [ -f /usr/bin/composer ]; then
+            sudo rm -f /usr/bin/composer
```
msaggiorato, line 35: This doesn't fully install the previously installed composer. Just removes it's executable. Shouldn't we also remove it's config and such?

msaggiorato (changes requested):

.github/workflows/vvv-provisioning.yml
evertiro: No. I discussed this in a bit more on [the 20.04 issue](https://github.com/Varying-Vagrant-Vagrants/VVV/pull/2389#issuecomment-815518165), but this only runs in the GH action. We _have_ to remove the bin in some environments (20.x+ actions include that file, which breaks our installer), but our composer doesn't have issues with any other files from the base install (honestly, I don't think there really are any), so the less we have to run the faster actions will complete.

Mte90: If it is just for the GH action I don't think that is a problem.

provision/core/composer/provision.sh
```diff
@@ -44,17 +42,19 @@ function composer_setup() {
...
     COMPOSER_HOME=/usr/local/src/composer noroot composer --no-ansi global config bin-dir /usr/local/bin
     COMPOSER_HOME=/usr/local/src/composer noroot composer --no-ansi self-update --2 --stable --no-progress --no-interaction
     vvv_info " * Making sure the PHPUnit 7.5 package is available..."
     COMPOSER_HOME=/usr/local/src/composer noroot composer --no-ansi global require --prefer-dist --no-update --no-progress --no-interaction phpunit/phpunit:^7.5
   fi
 
-  vvv_info " * Updating global composer packages..."
-  COMPOSER_HOME=/usr/local/src/composer noroot composer --no-ansi global update --no-progress --no-interaction
-  vvv_success " * Global composer package update completed"
+  if [ -f /usr/local/src/composer/composer.json ]; then
```
tomjn, line 53: On MacOS this folder doesn't exist, it's stored elsewhere. We need to figure out how to do this using composer commands

tomjn: I believe `composer config -g` can be used

tomjn: perhaps this:

```sh
~/dev/vvv ᚴ:develop 
❯ composer config -g data-dir
/Users/tarendai/.composer
```

or the `home` config value

tomjn: there's also `composer global show`

Mte90: I can confirm that on debian:
```$ composer config -g data-dir
/home/mte90/.composer
```

evertiro: This is a non-issue. Even on Mac, the VVV environment itself is Linux, and that path is valid.

PR #2453, opened 2021-04-28
MailHog tweaks

tomjn:
better messages, upgraded to 1.0.1, always copy the service file

## Checks

* [x] I've tested this PR

### DIFF
--- a/provision/core/mailhog/provision.sh
+++ b/provision/core/mailhog/provision.sh
@@ -37,9 +37,11 @@ function mailhog_setup() {
     fi
   fi
 
-  vvv_info " * Adding Mailhog service file"
-  # Make it start on reboot
-  cp -f "/srv/provision/core/mailhog/mailhog.service" "/etc/systemd/system/mailhog.service"
+  if [[ -d /etc/systemd/system/ ]]; then
+    vvv_info " * Adding Mailhog service file"
+    # Make it start on reboot
+    cp -f "/srv/provision/core/mailhog/mailhog.service" "/etc/systemd/system/mailhog.service"
+  fi
 
   # Start on reboot
   if [ "${VVV_DOCKER}" != 1 ]; then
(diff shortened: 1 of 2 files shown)


Mte90 (dismissed):

provision/core/mailhog/provision.sh
```diff
@@ -10,15 +10,15 @@ function mailhog_setup() {
 
   if [[ ! -e /usr/local/bin/mailhog ]]; then
     vvv_info " * Installing MailHog"
-    mailhog_bin="https://github.com/mailhog/MailHog/releases/download/v1.0.0/MailHog_linux_amd64"
+    mailhog_bin="https://github.com/mailhog/MailHog/releases/download/v1.0.1/MailHog_linux_amd64"
     if [[ "aarch64" == $(uname -m) ]]; then
       mailhog_bin="https://github.com/evertiro/MailHog/releases/download/v1.0.0-M1/MailHog_linux_arm64"
```
evertiro, line 15: Please bump this to 1.0.1 as well

evertiro (commented):
Update the M1 release to 1.0.1 as well

evertiro (dismissed):
Assuming no architecture-specific issues I can't test, this looks good!

provision/core/mailhog/provision.sh
```diff
@@ -28,20 +28,18 @@ function mailhog_setup() {
...
   fi
 
-  if [[ ! -e /etc/systemd/system/mailhog.service ]]; then
-    vvv_info " * Mailhog service file missing, setting up"
-    # Make it start on reboot
-    cp -f "/srv/provision/core/mailhog/mailhog.service" "/etc/systemd/system/mailhog.service"
-  fi
+  vvv_info " * Adding Mailhog service file"
+  # Make it start on reboot
+  cp -f "/srv/provision/core/mailhog/mailhog.service" "/etc/systemd/system/mailhog.service"
```
msaggiorato, line 42: This without the if would fail on docker right?

evertiro: I mean without copying over the service it'll fail anywhere the service doesn't already exist...

tomjn: the if just checked if the file existed, it didn't check for docker. If we ever changed `mailhog.service` or the file was broken somehow then the check would see it existed and skip copying, and it would never get fixed

msaggiorato: I meant mostly that there's no service manager in docker mostly.

So the path doesn't exist.

PR #2478, opened 2021-05-26
Emergency fix - broken default config

evertiro:
Somehow, we managed to merge a broken default config and nobody noticed. This is an emergency patch to fix it. A YAML node can't also be a tag.


msaggiorato (approved):

Mte90:
We need to test better the various pr...

PR #2479, opened 2021-05-26
Add restore exclude option

evertiro:
Adds the next logical step for extending database options (#2225).

## Checks

* [x] I've tested this PR

### DIFF
--- a/config/default-config.yml
+++ b/config/default-config.yml
@@ -143,7 +143,10 @@ general:
       #  - wordpress-trunk
 
   # Import the databases if they're missing from backups
-  db_restore: false
+  db_restore:
+    enable: false
+    #exclude:
+    #  - wordpress-trunk
 
   # set to true to use a synced shared folder for MariaDB database storage
   db_share_type: false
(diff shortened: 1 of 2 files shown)


evertiro:
This needs feedback. It _does_ work as is, but there are two caveats.

1. The current `develop` code for database restoration mirrors the backup `exclude` option. I dropped that entirely in favor of an actual `db_restore.exclude` option to match the backup option. This is the easy solution, but may not actually be what we _want_. It seems to me that there should be some sort of cascade to how options are handled.
2. Similarly, I left out the proposed `include` option. I tried several variations of adding an `include` option, but in the end, it always made things too confusing. How do you determine which takes precedence? What's the point of an include option? If you're treating it as an "only load these things" option (and therefore disregarding the `exclude` option entirely, we have to add another layer of logic that doesn't presently exist. If you aren't treating it that way, what's the point? Everything that isn't excluded explicitly gets included anyway.

Mte90:
To me it is the case to just exclude as it is more fast and easy to update but for the rest, let's see what @tomjn suggests.

evertiro:
I would suggest that we leave out any concept of `include`, and make it such that if `db_restore.enabled` is `true` and `db_restore.exclude` isn't set, it defaults to `db_backup.exclude`.

tomjn:
closing this as it's implemented in another PR

Issue #2482, opened 2021-06-02
Vagrant Up Fails MacOS Big Sur 11.4 ntupdate: command not found | install ntpdate manually, now Vagrant Provision errors


harrisonfm:
## What was The Command Used To Provision

`vagrant provision`

## What Kind of VVV Provision Was This

This was a reprovision of an already working VVV
then...
This was an attempt to upgrade to a new version of VVV

## Logs/What Broke

Update MacOS to Big Sur 11.4, Vagrant no longer works. I update VirtualBox, VVV, get `command ntpdate  not found` errors on vagrant up.

SSH into vagrant box, install ntpdate, vagrant up passes. Vagrant provision then fails.

```vagrant provision``` https://drive.google.com/file/d/1-aoh2kzwoWSkTbHax_20i2f_Y2vQcSSQ/view?usp=sharing

```yaml
hfm-2020:
    skip_provisioning: false
    description: ""
    repo: https://github.com/Varying-Vagrant-Vagrants/custom-site-template.git
    hosts:
      - hfm.test
    custom:
      install_test_content: true
      delete_default_plugins: true
```

## Your Environment

```
__ __ __ __
\ V\ V\ V / Varying Vagrant Vagrants
 \_/\_/\_/  v3.6.2-git::master(cabaa3a)

Platform: platform-darwin19.2.0 shell:/bin/zsh vagrant-hostsupdater vagrant-goodhosts shared_db_folder_disabled, VVV Path: "/Users/harrisonjune/vvv"
Vagrant: v2.2.13, VirtualBox: v6.1.22r144080
```

Thanks for any help!

tomjn:
@harrisonfm can you update your log to include the entire log? Of both the failed provision_and_ of the failed `vagrant up`. It's missing lots of information, we really do need the full entire output of the commands.

Also, I strongly recommend you switch away from the `master` branch to either `stable` or `develop`. The `develop` branch usually has fixes that haven't been merged to the stable branch yet, your issue may have been fixed months ago.

harrisonfm:
@tomjn , thanks for responding.

I just made sure that I pulled develop and got this same issue.

The ```vagrant provision``` log is huge, 70mb: https://drive.google.com/file/d/1-aoh2kzwoWSkTbHax_20i2f_Y2vQcSSQ/view?usp=sharing 

[...]

PR #2483, opened 2021-06-02
Turn off xdebug before wp package install and update wp packages

tomjn:
Possibly related to #2482

## Checks

* [ ] I've tested this PR

### DIFF
--- a/provision/core/wp-cli/provision.sh
+++ b/provision/core/wp-cli/provision.sh
@@ -29,18 +29,19 @@ function wp_cli_setup() {
     chown vagrant /srv/config/wp-cli/wp-completion.bash
   else
     vvv_info " * Updating wp-cli..."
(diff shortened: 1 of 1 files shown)


evertiro (commented):
Since the installation failing is extremely inconsistent, I can't swear that this fixes it... but I can say that it doesn't break anything and doesn't make the problem worse. How does one approve a fix for a bug that happens maybe 1/100th of the time?

Mte90 (approved):

Mte90:
> Since the installation failing is extremely inconsistent, I can't swear that this fixes it... but I can say that it doesn't break anything and doesn't make the problem worse. How does one approve a fix for a bug that happens maybe 1/100th of the time?

Good point I think that is just the case to approve and hope that fix for that strange case and moving on other things.

PR #2484, opened 2021-06-02
Adjust phpmod disabling

tomjn:
Explicitly disable FPM and CLI instead of implicitly all

## Checks

* [ ] I've tested this PR


evertiro:
@tomjn can you explain this change? I don't see anything wrong with the changed version, but I don't see why you're making the change either.

Mte90:
I guess that is happened that in a case one wasn't disabled and in this way we are forcing it.

Mte90 (approved):

tomjn:
@evertiro it makes disabling fpm and cli explicit rather than implicit, which hopefully fixes situations where xdebug is active when it's meant to have been turned off

Issue #2224, opened 2020-07-29
Skipped Sites still mount shared folders


tomjn:
If a site declares a `local_dir` and a `vm_dir` but has `skip_provisioning: true`, then Vagrant still mounts the folder. It shouldn't.

Related to #2131

tomjn:
Noting that for most sites this won't impact them as the `/srv/www` folder is mounted as a single shared folder, rather than for each individual site, so it only impacts sites that have set a custom `vm_dir`

PR #2487, opened 2021-06-09
Skip_provisioning support for vm_dir and local_dir

Mte90:
Fix https://github.com/Varying-Vagrant-Vagrants/VVV/issues/2224

## Checks

* [ ] I've tested this PR


tomjn (changes requested):
ooo I like this `next` pattern 👍🏻  just needs a changelog update

Mte90:
Changelog updated, I preferred to wait as there are always conflicts with that.

For the code is a copy and paste from the code that works in the same file and ported also in other sections.

PR #2488, opened 2021-06-09
Set php version

Mte90:
https://github.com/Varying-Vagrant-Vagrants/VVV/issues/2309

## Checks

* [ ] I've tested this PR

### DIFF
--- a/config/homebin/get_php_version.sh
+++ b/config/homebin/get_php_version.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+#
+# Get PHP version from config.yml or fallback to the system version
+set -eo pipefail
+
+source /srv/provision/provision-helpers.sh
+
+phpversion=$(get_config_value "general.default_php")
+if [ -z "$phpversion" ] then
+  echo $phpversion
+else
+  phpversion=$(php --version | head -n 1 | cut -d " " -f 2 | cut -c 1-3)
+fi
+
+return $phpversion
--- a/config/homebin/get_php_version.sh
+++ b/config/homebin/get_php_version.sh
@@ -6,10 +6,8 @@ set -eo pipefail
 source /srv/provision/provision-helpers.sh
 
 phpversion=$(get_config_value "general.default_php")
-if [ -z "$phpversion" ] then
-  echo $phpversion
-else
+if [[ ! -z "$phpversion" ]]; then
   phpversion=$(php --version | head -n 1 | cut -d " " -f 2 | cut -c 1-3)
 fi
 
-return $phpversion
+echo "$phpversion"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ permalink: /docs/en-US/changelog/
 * Switch from Ubuntu 18.04 to 20.04 (current LTS release)
 * Simplified config folder
 * Increased the default PHP memory limit from 128MB to 256MB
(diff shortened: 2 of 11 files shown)


config/homebin/get_php_version.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+#
+# Get PHP version from config.yml or fallback to the system version
+set -eo pipefail
+
+source /srv/provision/provision-helpers.sh
+
+phpversion=$(get_config_value "general.default_php")
```
Mte90, line 8: It is missing the default from VVV.

Mte90:
Right now we already set the php version on provisioner. This was to help us to change the php version on VVV and also for the people that asks often how to change the default php version.

tomjn:
> Right now we already set the php version on provisioner.

We just allow the nginx upstream to be set, which isn't quite the same thing, e.g. the site provisioner still runs on PHP 7.4 even if `nginx_upstream` is set to `php72`

> This was to help us to change the php version on VVV and also for the people that asks often how to change the default php version.

People who ask this are just doing the problem X Y thing trying to solve other issues. E.g. xhgui not working on PHP 7.4, or a specific site requiring 7.2 or 7.3

Imagine it this way, we serve cheese and you're lactose intolerant, so we switch everybody to peanut butter and you're happy, but the person next to you has anaphylactic shock and goes to hospital.

 - ✅ a bash function is good as it simplifies provisioning
 - ❌ a `config.yml` option and a command are bad as it introduces new sources for bugs and will be misused by people trying to solve unrelated bugs

There's also the issue that if I set 7.4 as the default in the config, then VVV upstream switches to 8.0 as the default, when I update none of my sites or the dashboard will work because they use the wrong nginx upstream.

What we actually need is:

```yaml
  wordpress-two:
    php: 7.2
    hosts:
      - two.wordpress.test
```

With this we can auto-configure the upstream, change the default PHP but just for that specific site, we could even modify the terminal prompt to modify `PATH` so that `php` is mapped to the correct version when they `cd` into that directory

Mte90:
So that change is not for VVV but for the provisioner, so the php commands on that will use that version for everything.

Instead my idea was to create a tool to help us to change the default php version, so maybe we can have a constant internally and remove that stuff from nginx.

tomjn:
> So that change is not for VVV but for the provisioner, so the php commands on that will use that version for everything.

Extending the custom site template would mean conflicting instructions on setting up PHP, or, parsing the nginx upstream value to try and extract the PHP version, both of which are confusing and problematic for users. This is not the solution.

> Instead my idea was to create a tool to help us to change the default php version, so maybe we can have a constant internally and remove that stuff from nginx.

We cannot allow the default PHP to be configured via `config.yml`.

tomjn:
It seems like this PR conflates several things:

 - letting users set the default PHP version
 - making the PHP configs generic
 - adding CLI commands for changing the current default version of PHP

### letting users set the default PHP version

A hard no, this must not be possible.

### making the PHP configs generic

Are we sure that newer version of PHP won't require changes? What benefit do we have from this

### adding CLI commands for changing the current default version of PHP

I can see value in this but I think that changing the default PHP isn't the solution, but rather changing the PATH. This way, we can do it on a per session basis that's safe for multiple sessions and parallel commands.

Mte90:
So we can close this PR and update the ticket with a recap of what we need with al those details.

PR #2498, opened 2021-07-20
add db.vvv.test and api.vvv.test hosts

tomjn:
## Checks

* [ ] I've tested this PR


tomjn:
Noting that the name of the database container if we're doing docker would be what gets used, so we should be sure that we pick carefully, `db.vvv.test` may not work

tomjn:
I'm not sure this is needed, we can redo it easily in future if we change our mind

PR #2499, opened 2021-07-21
Default Nginx config, #2279

### LINKED ISSUE #2279: Default Nginx config
When a site doesn't provide an Nginx config, we should fallback to a generic Nginx config in the sites provisioner

Mte90:
Should fix https://github.com/Varying-Vagrant-Vagrants/VVV/issues/2279

## Checks

* [x] I've tested this PR


provision/provision-site.sh
```diff
@@ -84,6 +84,10 @@ function get_primary_host() {
 function vvv_provision_site_nginx_config() {
   local SITE_NAME=$1
   local SITE_NGINX_FILE=$2
+  if [ ! -f "$SITE_NGINX_FILE" ]; then
+    vvv_warn " * VVV didn't found any Nginx config file, it will use a fallback config file"
+    cp -f "/srv/provision/core/nginx/config/sites/fallback.conf" "$SITE_NGINX_FILE"
+  fi
```
tomjn, line 90: this scenario isn't possible because the code that calls `vvv_provision_site_nginx_config` already performs this check, and if nothing is found it moves on to the next thing. If nothing is found then this function is never called.

`vvv_provision_site_nginx` would need changes to support this too

tomjn: I would imagine instead this part of the `vvv_provision_site_nginx` function would need changing:

```sh
    if [[ -z $NGINX_CONFIGS ]] ; then
      vvv_error " ! Error: No nginx config was found, VVV will not know how to serve this site"
      exit 1
```

So that it called `vvv_provision_site_nginx_config` passing the fallback config as the parameter

tomjn (dismissed):

tomjn:
@Mte90 is some of it missing? I see it reverted a portion in the last commit

Mte90:
> @Mte90 is some of it missing? I see it reverted a portion in the last commit

I changed the if as you suggested.

tomjn:
@Mte90 `vvv_provision_site_nginx_config` still needs calling with the fallback config in the other function

provision/provision-site.sh
```diff
@@ -84,6 +84,11 @@ function get_primary_host() {
 function vvv_provision_site_nginx_config() {
   local SITE_NAME=$1
   local SITE_NGINX_FILE=$2
+  local NGINX_CONFIGS=$(find "${VM_DIR}" -maxdepth 3 -name 'vvv-nginx.conf');
+  if [[ -z $NGINX_CONFIGS ]] ; then
+      vvv_error " ! Error: No nginx config was found, VVV will not know how to serve this site"
+      exit 1
+  fi
```
tomjn, line 91: I'm confused by this change, it's not the job of `vvv_provision_site_nginx_config` to locate the nginx config, it's passed to it as the second parameter. That task is the responsibility of a different function

tomjn: `vvv_provision_site_nginx` is the function that finds the Nginx file, and when it finds it it calls this function. That's what needed to be modified because when that function does not find a config, it fails the provisioner. The goal of the ticket is that it does not fail the provisioner but instead falls back to a default config

Mte90:
I misunderstood your request, now should be right (I hope).

tomjn:
I think this needs the main `develop` branch merging into it so that the CI actions pass

Mte90:
Seems that GitHub doesn't find any conflicts.

tomjn:
@Mte90 more so that it pulls in fixes in develop for CI so that it can pass

tomjn:
@Mte90 can you resolve the conflicts and update this for the latest `develop` branch?

Mte90:
Conflicts fixed.

provision/provision-site.sh
```diff
@@ -135,6 +135,16 @@ function get_primary_host() {
 function vvv_provision_site_nginx_config() {
   local SITE_NAME=$1
   local SITE_NGINX_FILE=$2
+
+  if [ ! -f "$SITE_NGINX_FILE" ]; then
+    vvv_warn " * VVV didn't found any Nginx config file, it will use a fallback config file"
+    cp -f "/srv/provision/core/nginx/config/sites/fallback.conf" "$SITE_NGINX_FILE"
+  fi
+  local DEST_NGINX_FILE=${SITE_NGINX_FILE//\/srv\/www\//}
```
Mte90, line 143: after all those time I am not sure if these lines are needed as we have there on https://github.com/Varying-Vagrant-Vagrants/VVV/blob/2593a4cbba2ccdbaeb7f909a4bd5c6296d0224df/provision/provision-site.sh#L181

tomjn:
@Mte90 I tested this PR and ran into a failed provision with Nginx configs failing due to syntax errors:

```
vagrant@vvv:~$ sudo nginx -t
nginx: [emerg] directive "server_name" is not terminated by ";" in /etc/nginx/custom-sites/fallback.conf:4
nginx: configuration file /etc/nginx/nginx.conf test failed
vagrant@vvv:~$ 
```

`fallback.conf` should not have been included based on my understanding, it's just a template that's used to generate the sites actual Nginx config if the site provides no config of its own ( either because there is no provisioner or none was provided ).

To test this I did these steps:

 - I created a `www/wordpress-three/public_html` folder and put an `index.php` inside it with a generic "It works!" type message
 - I then added a `wordpress-three` site to `config.yml` that did not have a `repo` parameter, so no site provisioner was present
 - I then reprovisioned with `vagrant up --provision`

I suspect part of the problem is the location of the fallback config is in a place where it's automatically included, which is not how it should work. `fallback.conf` is a template file, much like how we have template files in the site provisioner.

tomjn:
I moved the file one folder up so it wasn't copied into place, and renamed it to `site-fallback.conf`, adjusted `provision-site.sh` accordingly, and re-ran the provision, and got this:

```
==> default: Running provisioner: site-wordpress-three (shell)...
    default: Running: /var/folders/dw/wdtnsdn154db04ymxv90t9tm0000gn/T/vagrant-shell20221113-84039-12bycqx.sh
    default:  ▷ Running the 'site-wordpress-three' provisioner...
    default:  * The site: 'wordpress-three' does not have a site template, assuming custom provision/vvv-init.sh and provision/vvv-nginx.conf
    default:  * Warning: A site provisioner was not found at .vvv/vvv-init.sh provision/vvv-init.sh or vvv-init.sh, searching 3 folders down, please be patient...
    default:  * Warning: No site provisioner was found, VVV could not perform any scripted setup that might install software for this site
    default:  ! Warning: An nginx config was not found!! VVV needs an Nginx config for the site or it will not know how to serve it.
    default:  * VVV searched for an Nginx config in these locations:
    default:    - /srv/www/wordpress-three/.vvv/vvv-nginx.conf
    default:    - /srv/www/wordpress-three/provision/vvv-nginx.conf
    default:    - /srv/www/wordpress-three/vvv-nginx.conf
    default:  * VVV will search 3 folders down to find an Nginx config, please be patient...
    default:  ! Error: No nginx config was found, VVV will not know how to serve this site
    default:  ! The 'site-wordpress-three' provisioner ran into problems, the full log is available at '/var/log/provisioners/2022.11.13_16-28-19/provisioner-site-wordpress-three.log'. It completed in 0 seconds.
```

tomjn:
Pushed up fixes

tomjn (approved):

Issue #2502, opened 2021-07-28
Add backward compatible for db_backup option 


emgk:
Did we intentionally extended `db_backup` option here https://github.com/Varying-Vagrant-Vagrants/VVV/pull/2478 _(couldn't find  useful doc related to this change)_, pulling these changes to the local setup can break the purpose of db_backup, because its configured in an old way `db_backup: false` in vvv/config.yml file, can we add backward compatibility in following files? 

https://github.com/Varying-Vagrant-Vagrants/VVV/blob/b4d2655bf01120e32cdfc6251af26a2c009eeccc/config/homebin/vagrant_destroy#L21
https://github.com/Varying-Vagrant-Vagrants/VVV/blob/b4d2655bf01120e32cdfc6251af26a2c009eeccc/config/homebin/vagrant_halt#L21
https://github.com/Varying-Vagrant-Vagrants/VVV/blob/b4d2655bf01120e32cdfc6251af26a2c009eeccc/config/homebin/vagrant_suspend#L21 

Thanks :pray:

Mte90:
So you mean add a code that check the previous parameter for old VVV instance that didn't updated the settings?
If yes , it is something that we need to do.

tomjn:
@Mte90 this was a part of the original spec, we can't break things for existing users, it should never have been merged if it that wasn't the case

tomjn:
@Mte90 I'm labelling this OMGWTFBBQ level high priority

tomjn:
@emgk that PR isn't related to this, it fixes a YAML syntax error

tomjn:
https://github.com/Varying-Vagrant-Vagrants/VVV/pull/2346 is the PR that implemented this

Based on the what the OP says It looks like nobody did the testing I asked for:

https://github.com/Varying-Vagrant-Vagrants/VVV/pull/2346#issuecomment-821257734

Mte90:
I did a pr but require some testings.

PR #2503, opened 2021-07-28
Fix 2502

Mte90:
Ref: https://github.com/Varying-Vagrant-Vagrants/VVV/issues/2502

## Checks

* [ ] I've tested this PR

### DIFF
--- a/config/homebin/vagrant_destroy
+++ b/config/homebin/vagrant_destroy
@@ -18,10 +18,13 @@ else
     fi
 
     # Check if backups are turned on or off
-    run_backups=`cat ${VVV_CONFIG} | shyaml get-value general.db_backup.enable`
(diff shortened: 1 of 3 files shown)


tomjn:
@Mte90 if you use `get-type` instead you can test the value of `db_backup`, if it's true/false you'll get `bool` back, if it's got sub-values you'll get `struct`

tomjn:
Also there's parameter expansion which will need quote wrapping

Mte90:
I should be able to improve the code using the quiet parameter of shyaml.

emgk:
I've tested the changes with all possible cases using `vagrant_halt` script, seems to be working fine, thank you guys.

tomjn (approved):
👍🏻

PR #2506, opened 2021-07-28
Time to make the long provisioning warning be a bear. (and also provide more accurate information)

bradp:
Updates the character that warns you the first provision is long to be a bear. 

tea-time 

Also updated the text, because previously, it suggests running  `vagrant up` if you only wanted to turn on VVV, but the first time you do a `vagrant up`, if you haven't provisioned, you'd see this message. So now it's a bit more clear hopefully.

I kept the output as inline in the Vagrantfile, but happy to split into a separate file if we want.

Also look how cute they are with their lil tea mug, even sitting on a little table

```shell
    default: Running: inline script
    default:     
    default:                                       A full provision will take a bit.
    default:     ▄▀▀▀▄▄▄▄▄▄▄▀▀▀▄          ▄   ▄    Sit back, relax, and have some tea.
    default:     █▒▒░░░░░░░░░▒▒█         █   █
    default:      █░░█░░░░░█░░█         ▀   ▀      If you keep seeing this message,
    default:   ▄▄  █░░░▀█▀░░░█  ▄▄  ▄▀▀█▀▀▀▀▀▀█    make sure you're running 'vagrant up'
    default:  █░░█ ▀▄░░░░░░░▄▀ █░░█ █  █      █
    default: ─────────────────────── ▀▀█      █ ───────────────
    default:                            ▀▄▄▄▄▀
    default:
==> default: Running provisioner: default (shell)...
    default: Running: /var/folders/86/w0vt5r5x4n369p808m9py51c0000gn/T/vagrant-shell20210728-66392-crly12.sh
    default:  - skipping ntpdate clock sync, not installed yet
    default:  ✔  ▷ Running the 'main' provisioner
```

## Checks

* [x] I've tested this PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ permalink: /docs/en-US/changelog/
 
 ## 3.7.2 ( 2021 )
 
+* Updated message on full provision to be more clear
+* Added mascot to full provision message
 
 ## 3.7.1 ( 2021 July 20th )
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,7 @@ permalink: /docs/en-US/changelog/
 
 ## 3.7.2 ( 2021 )
 
-* Updated message on full provision to be more clear
-* Added mascot to full provision message
+* Added a new bear to full provision message, and updated message to be more clear
 
 ## 3.7.1 ( 2021 July 20th )
 
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -694,7 +694,19 @@ Vagrant.configure('2') do |config|
     end
   end
 
-  config.vm.provision "pre-provision-script", type: 'shell', keep_color: true, inline: "echo \"\n༼ つ ◕_◕ ༽つ A full provision can take a little while!\n             Go make a cup of tea and sit back.\n             If you only wanted to turn VVV on, use vagrant up\n\""
+  # \033[38;5;4m
+  # \033[38;5;4m                                      \033[01;32mA full provision will take a bit.
+  # \033[38;5;4m    ▄▀▀▀▄▄▄▄▄▄▄▀▀▀▄          ▄   ▄    \033[01;32mSit back, relax, and have some tea.
+  # \033[38;5;4m    █▒▒░░░░░░░░░▒▒█         █   █
+  # \033[38;5;4m     █░░█░░░░░█░░█         ▀   ▀      \033[01;32mIf you keep seeing this message,
+  # \033[38;5;4m  ▄▄  █░░░▀█▀░░░█  ▄▄  ▄▀▀█▀▀▀▀▀▀█    \033[01;32mmake sure you're running 'vagrant up'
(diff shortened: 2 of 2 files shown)


bradp:
For context, here's the old version of what was there. As cute as it is, a bit out of place with all the bears

```shell
༼ つ ◕_◕ ༽つ A full provision can take a little while!
             Go make a cup of tea and sit back.
             If you only wanted to turn VVV on, use vagrant up
```

tomjn:
Some notes:

 - we have variables in the vagrant file for colours and resets that can make this nicer in code
 - kill the bottom of the coffee cup and move it over so it overlaps a bit, we can get rid of the right hand of the bear
 - max of 1 line of empty space after and before
 - we can probablly shorten or adjust the text, doesn't need to be a separate colour
 - we should check on terminals with white backgrounds

bradp:
> * we have variables in the vagrant file for colours and resets that can make this nicer in code

Unfortunately, they aren't able to be used in the context where this is set, as we're passing an inline bash command, so we don't have access to the Vagrantfile variables. Could be possible to build the string up separately and then pass it, but that's probably overkill just for these colors.

Updated bear:
Screen Shot 2021-07-29 at 3 05 48 PM

These are the default Terminal themes, and they all look good. Taking the color off the text helps, because now it's just the cyan bear, which is used in `vagrant halt` and the other hooks.
Screen Shot 2021-07-29 at 3 05 33 PM

bradp:
lmao im an idiot

tomjn:
coffee cup looks good

tomjn:
Just needs a changelog update then it can be merged

bradp:
Does the bear have a name?

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ permalink: /docs/en-US/changelog/
 
 ## 3.7.2 ( 2021 )
 
+* Updated message on full provision to be more clear
+* Added mascot to full provision message
```
msaggiorato, line 12: I would use bear instead of mascot, as it's used in other places across the changelog, and also join it in a single line.
```suggestion
* Added a new bear to full provision message, and updated message to be more clear
```

msaggiorato (changes requested):

CHANGELOG.md
bradp: 👍 Sounds good to me. Asked about the name because I was thinking about using that, but matching what's been used before is more important

🎉

tomjn: VVV bear? The bear was more me adding a bit of whimsy so that it was obvious when a vagrant up or provision had succeeded, users weren't sure if it was still provisioning or what to do next. A big ascii teddy bear seemed like a great way to indicate it had finished, and successfully

bradp: Well now we need to give them a name! I love that lil bear

tomjn: Or her! I've asked the bear and have yet to get an answer back. Perhaps one day in the dashboard you can customize your own VVV bear but that's a long way off

tomjn (approved):

Issue #2270, opened 2020-10-22
Split main provisioner into 2


tomjn:
1 provisioner for apt and system level packages like nginx php and git, another for tools such as node and composer packages

PR #2516, opened 2021-08-28
Tools Provisioner

tomjn:
Implements #2270 

## Checks

* [x] I've tested this PR

### DIFF
--- a/provision/provision.sh
+++ b/provision/provision.sh
@@ -90,7 +90,7 @@ if ! vvv_package_install ${VVV_PACKAGE_LIST[@]}; then
   exit 1
 fi
 
-vvv_info " * Running tools_install"
+vvv_info " * Running after_packages"
 vvv_hook after_packages
 
 vvv_info " * Finalizing"
--- a/provision/provision.sh
+++ b/provision/provision.sh
@@ -57,19 +57,17 @@ export VVV_PACKAGE_REMOVAL_LIST=()
 . "/srv/provision/core/nginx/provision.sh"
 . "/srv/provision/core/memcached/provision.sh"
 . "/srv/provision/core/php/provision.sh"
-. "/srv/provision/core/composer/provision.sh"
-. "/srv/provision/core/nodejs/provision.sh"
-. "/srv/provision/core/grunt/provision.sh"
 . "/srv/provision/core/mailhog/provision.sh"
-. "/srv/provision/core/wp-cli/provision.sh"
-. "/srv/provision/core/phpcs/provision.sh"
+. "/srv/provision/core/nodejs/provision.sh"
 
 ### SCRIPT
 #set -xv
 
 vvv_hook init
 
+# If you need to disable this check then something is terribly wrong, tell us on github/slack
 if ! network_check; then
+  vvv_warning " ! If this check fails despite succeeding in the browser, contact us in Slack or GitHub immediatley"
   exit 1
 fi
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ permalink: /docs/en-US/changelog/
 
 ## 3.8.x ( 2021 )
 
+### Enhancements
+
+* Split tools out into their own provisioner ( #2270 )
+
 ### Bug Fixes
 
 * Fixed backwards compatibility for enabling backups in `config.yml` via `backup: true`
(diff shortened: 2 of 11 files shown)


tomjn:
This is running ok for a VM that's already provisioned and updated to this code.

It needs testing for a brand new VVV too confirm it does not break provisioning

provision/provision.sh
```diff
@@ -57,19 +57,17 @@ export VVV_PACKAGE_REMOVAL_LIST=()
 . "/srv/provision/core/nginx/provision.sh"
 . "/srv/provision/core/memcached/provision.sh"
 . "/srv/provision/core/php/provision.sh"
-. "/srv/provision/core/composer/provision.sh"
-. "/srv/provision/core/nodejs/provision.sh"
-. "/srv/provision/core/grunt/provision.sh"
 . "/srv/provision/core/mailhog/provision.sh"
-. "/srv/provision/core/wp-cli/provision.sh"
-. "/srv/provision/core/phpcs/provision.sh"
+. "/srv/provision/core/nodejs/provision.sh"
```
Mte90, line 61: this isn't a tool after all? should be part of that provisioner?

provision/provision.sh
```diff
@@ -57,19 +57,17 @@ export VVV_PACKAGE_REMOVAL_LIST=()
...
 
 ### SCRIPT
 #set -xv
 
 vvv_hook init
 
+# If you need to disable this check then something is terribly wrong, tell us on github/slack
```
Mte90, line 68: maybe is better to add a parameter in `config.yml` so people don't need to hack with the VVV code?

Mte90 (changes requested):

provision/provision.sh
tomjn: Which? PHPCS and WP CLI are tools so they're moved to the tools provisioner, Node is a base dependency

provision/provision.sh
tomjn: We've discussed this elsewhere and the conclusion was "tell us how it failed and we can fix it", if the feature si broken we should fix it, adding an option called `bypass_broken_check: true` is just silly

provision/provision.sh
Mte90: it was node but ok

tomjn: Ah, Node isn't a tool

Mte90:
```
    default:  ▷ Running the 'tools' provisioner...
    default:  * Running tools_install
    default:  ▷ Running tools_setup hook
    default:  * Turning off XDebug to avoid Composer performance issues
    default:  ✘ Disabling active PHP v8.0 debug mod: 'xhgui'
    default:  ✘ Disabling active PHP v7.4 debug mod: 'xhgui'
    default:  ✘ Disabling active PHP v7.3 debug mod: 'xhgui'
    default:  ✘ Disabling active PHP v8.0 debug mod: 'tideways_xhprof'
    default:  ✘ Disabling active PHP v7.4 debug mod: 'tideways_xhprof'
    default:  ✘ Disabling active PHP v7.3 debug mod: 'tideways_xhprof'
    default:  * Restarting PHP FPM services so that the change takes effect
    default:  ✔ All PHP Debug mods are now turned off.
    default:  * Checking if Composer is installed
    default:  * Making sure the Composer cache is not owned by root
    default:  * Checking for GitHub tokens
    default:  * Checking for Composer updates
    default:  ▷ Running after_composer hook
    default:  * Provisioning PHP_CodeSniffer (phpcs), see https://github.com/squizlabs/PHP_CodeSniffer
    default: Loading composer repositories with package information
    default: Updating dependencies
    default: Nothing to modify in lock file
    default: Installing dependencies from lock file (including require-dev)
    default: Nothing to install, update or remove
    default: Generating autoload files
    default:  * Setting WordPress-Core as the default PHPCodesniffer standard
    default:  * The following PHPCS standards are set up:
    default:  * PHPCS provisioning has ended
    default:  ✔ Finished after_composer hook in 3s
    default:  * Updating Grunt CLI
    default:  * Updated Grunt CLI
    default:  * Installing/updating WP-CLI
    default:  * Updating wp-cli...
    default:  * WP CLI Nightly updated
    default:  * Disabling debug mods if present before running wp package installs
    default:  * Restarting PHP FPM services so that the change takes effect
    default:  ✔ All PHP Debug mods are now turned off.
    default:  * Updating WP packages
    default:  * WP package updates completed
    default:  ✔ Finished tools_setup hook in 16s
    default:  * Finalizing Tools
    default:  ▷ Running tools_finalize hook
    default:  ✔ The 'tools' provisioner completed in 16 seconds.
```
Seems working with no troubles

Mte90 (approved):

PR #2520, opened 2021-08-29
add vvv_parallel_hook

tomjn:
Adds a version of `vvv_hook` that runs all the functions at the same time

## Checks

* [ ] I've tested this PR

### DIFF
--- a/provision/provision-helpers.sh
+++ b/provision/provision-helpers.sh
@@ -402,6 +402,36 @@ vvv_hook() {
 }
 export -f vvv_hook
 
+# @description Executes a hook. Functions added to this hook will be executed in parallel
+#
+# @example
+#   vvv_parallel_hook before_packages
+#
+# @arg $1 string the hook to execute
+vvv_parallel_hook() {
+  if [[ "${1}" =~ [^a-zA-Z_] ]]; then
+    vvv_error " x Disallowed hookname '${1}'"
+    return 1
+  fi
+
+  local hook_var_prios="VVV_HOOKS_${1}"
+  local start=`date +%s`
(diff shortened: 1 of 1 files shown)


tomjn:
Currently untested, would be good for the tools provisioner once that's merged

tomjn:
This article walks through things:

http://prll.sourceforge.net/shell_parallel.html

It may  be useful in limiting the concurrent functions running. It will never be as good as using GNU Parallel though

tomjn:
Also https://gist.github.com/reasonset/782725e1648aca2d60fa96590433f2f4

tomjn:
and https://coderwall.com/p/k4if5a/throttling-parallel-processes-in-your-shell

Mte90:
It isn't better to use https://www.gnu.org/software/parallel/ to run parallel executions?
As https://opensource.com/article/18/5/gnu-parallel

We have also a ticket about it https://github.com/Varying-Vagrant-Vagrants/VVV/issues/2269

tomjn:
@Mte90 we could but that would require an additional dependency, something to consider as an enhancement

Mte90:
I think that is better to use `parallel` instead of this code

tomjn:
@Mte90 if you know how to do that then that's great, but it's much more difficult to do. The functions to run are an array of function names as strings. So you can't just pass `parallel` the array and tell it to run them all

If it were that easy to do then this PR would have used `xarg` to do it

tomjn:
If you insist on it though, then we can hold this PR and spend several weeks/months figuring out how to do it ( if it can be done ). Or do it as an enhancement in a future PR

Issue #2515, opened 2021-08-28
Avahi/Zeroconf support


tomjn:
Sites that use a `.local` TLD could be advertised using `.local`.

This would give a way to provide a hostname that can be used on devices that do not support host file modification. For example, when provisioning on a dedicated device such as a raspberry Pi connected to an iOS device.

tomjn:
This is the puppet config Chassis uses to do it:

https://github.com/humanmade/Chassis/blob/master/puppet/modules/chassis/manifests/hosts.pp

tomjn:
So installing `avahi-daemon` is enough to get it to expose itself, but it will only expose the machines hostname ( or a different one if we configure it ), and can only do `.local` TLDs.

If we want multiple TLDs though then we need to take extra steps of wwhcih there are a few options

- we can use `/etc/avahi/hosts` to list additional domains, but we need to specify the IP too and restart the service when it changes ( it may not be necessary too restart the avahi service ). The file has the same format as the system hosts file
- we can install `avahi-utils` and use `avahi-publish-address` too add a single hostname, which immediatley advertises the new host, but the command will keep running, it's more like a daemon/background process
  - e.g. `avahi-publish -a -R taxgod.local 192.168.1.2` ( tested locally on a non-VVV system )
  - or `/usr/bin/avahi-publish -a -R taxgod.local 192.168.1.2 &`
 - This stack exchange question creates a service that can be enabled for each host/domain you want using multi-instance support that can be stopped and started: https://serverfault.com/a/986437/29449

Useful links:

 - https://www.shixuen.com/linux/services_avahi.html
 - https://linuxcommandlibrary.com/man/avahi-publish
 - https://stackoverflow.com/questions/775233/how-to-route-all-subdomains-to-a-single-host-using-mdns
 - https://pi3g.com/2019/04/10/avahi-how-to-assign-several-local-names-to-same-ip/

tomjn:
Also, it seems that this won't work on Windows out of the box:

 - Newer builds of Windows 10 have support if you switch a registry flag
 - Anybody who's installed iTunes or some other software may find it came with a client to make this work
 - Installing bonjour print services for Windows also installs a client

tomjn:
#2525 installs the daemon and exposes the `vvv.local` default host, it also installls `avahi-utils`. It doesn't support the other hosts or site hosts though

tomjn:
[...]

PR #2523, opened 2021-09-09
Basic Avahi support

tomjn:
Related to #2515 

## Checks

* [ ] I've tested this PR


tomjn:
With this PR `vvv.local` works without host file modifications.

Currently this PR does not add support for custom sites though, _only `vvv.local`_

Screenshot 2021-09-09 at 22 51 11

msaggiorato:
I confirm this works in windows (given that you have some Zeroconf client installed). I cannot check specifically if Windows 10 nowadays supports this by default, or if something else installed Bonjour or something similar.

PR #2526, opened 2021-09-17
Rename utilities to extensions

tomjn:
## Checks

* [x] I've tested this PR


Mte90:
We have a fallback in case it is not define or an alert to the users for this change?

tomjn:
> We have a fallback in case it is not define or an alert to the users for this change?

Check the changes, it still reads the config utilities sections, and it also moves the utilities folder so nothing is lost. Fallbacks for when extensions are not in the config file already exist and existed before for the utilities

Mte90 (approved):
Tested also the provision and everything is working.

PR #2536, opened 2021-10-10
Test provisioning on vagrant, instead of a linux VM with docker.

msaggiorato:
This tests provisioning in different scenarios with vagrant and virtualbox.

One issue i've noticed:

There seems to be some times where we get failures on the provisioners (likely product of a bad setup of the MacOS VM or something of the sort). Rerunning the actions typically get a valid result.

The error is something like follows:

>   Vagrant can't use the requested machine because it is locked! This
>   means that another Vagrant process is currently reading or modifying
>   the machine. Please wait for that Vagrant process to end and try
>   again. Details about the machine are shown below:
> 
>   Name: default
>   Provider: virtualbox

Again, just re running the action (which will run the 3 jobs again, since they're all part of the same workflow).

Didn't change the changelog, since this doesn't change any features.

## Checks

* [x] I've tested this PR

### DIFF
--- a/.github/workflows/vvv-provisioning.yml
+++ b/.github/workflows/vvv-provisioning.yml
@@ -25,6 +25,9 @@ jobs:
         with:
           ref: stable
 
+      - name: install goodhosts
+        run: vagrant plugin install --local
+
       - name: vagrant up (stable)
         run: vagrant up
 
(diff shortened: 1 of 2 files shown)


tomjn:
Also this PR no longer tests pull requests, as it will just run it on `stable` and `develop` rather than the current branch which is only usefufl for releases

tomjn (approved):

Issue #2546, opened 2021-10-26
Bug: Import ignore excluded db from backup


Mte90:
## Expected Behavior
If you have some DB excluded in the config they are excluded from getting imported as https://github.com/Varying-Vagrant-Vagrants/VVV/blob/develop/database/sql/import-sql.sh#L33 check for that parameter. The issue is that if the file of the backup exists should automatically import and ignore that value.

The issue happens on destroy that don't restore the DB at first provision.

PR #2547, opened 2021-10-26
fix(db): importing when excluded

Mte90:
Fix https://github.com/Varying-Vagrant-Vagrants/VVV/issues/2546

## Checks

* [x] I've tested this PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@ permalink: /docs/en-US/changelog/
 * Service restarts now have dedicated functions
 * Several evals removed from the hook functions
 * Disabled nested virtualisation under Hyper-V
-* Import a DB when is excluded (from creating the dump) and the dump exist
+* Import a DB dump when excluded, the db doesn't exists and the dump file exists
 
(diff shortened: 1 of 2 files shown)


Mte90:
Tested with the latest changes and is working to me with the new parameters

tomjn:
@Mte90 we still need to account for backwards compatibility for when `general.db_restore` is a boolean not a struct, e.g.:

```yaml
general:
  db_restore: false
```

tomjn:
A parameter under `db_restore` to act as its replacement makes sense, but I'm not sure what it should be named

tomjn (approved):

Mte90:
> @Mte90 we still need to account for backwards compatibility for when `general.db_restore` is a boolean not a struct, e.g.:
> 
> ```yaml
> general:
>   db_restore: false
> ```

The only use of that parameter is in that script and this check if it is a bool https://github.com/Varying-Vagrant-Vagrants/VVV/blob/4c74fee6e4bc097cf76620671791c775be46b8b7/database/sql/import-sql.sh#L33

PR #2574, opened 2022-01-23
Expand the composer installation steps output and checks

tomjn:
## Checks

* [ ] I've tested this PR

### DIFF
--- a/config/homebin/switch_php_debugmod
+++ b/config/homebin/switch_php_debugmod
@@ -1,6 +1,5 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -eo pipefail
-
 
 source /srv/provision/provision-helpers.sh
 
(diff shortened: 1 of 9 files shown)


tomjn:
So the downloading of WP CLI is failing sometimes for unknown reasons but succeeding other times. Curl exit code of 3 which implies a malformed request.

I think to get more useful data it might be best to run the command without the silencing option in a sub-shell and inspect the exit code via `$?` then print out the results if it failed so we can see what curl said.

tomjn:
so `npm` being missing after grunts installation is frustrating and difficult to explain

tomjn:
Todo list:

 - figure out how to make the parallelised hooks exit on failure and acknowledge bad return codes
 - I think npm issue can be resolved by switching the node provisioner to nvm, it's been a long time coming
 
For now this PR disabled parallelised tools provisioners to get around the first bullet point, I believe the second bullet point already impacts the main branch, so I vote to merge once changelogs are updated

Mte90 (approved):

Mte90:
We need a ticket to track this other things to do but about the rest ok

PR #2575, opened 2022-01-26
Updated mariadb mirror for connectivity check

Mte90:
as title, so we can avoid issues reported for that


tomjn (approved):

tomjn:
@Mte90 needs changelog update

tomjn (approved):

Issue #2367, opened 2021-02-17
Ability to select PHP version used during site provisioning


spenserhale:
## What was The Command Used To Provision

```shell
composer_setup
```

## What Kind of VVV Provision Was This
This was a reprovision of an already working VVV

## Logs/What Broke
Expected I set PHP74 version in site config, but provisioner ran an older PHP version.
Log:
```
default:  ▷ Running the 'site-demo-sites' provisioner...
    default: Composer detected issues in your platform:
    default:
    default: Your Composer dependencies require a PHP version ">= 7.3.0". You are running 7.2.34-13+ubuntu18.04.1+deb.sury.org+1.
    default:
    default: PHP Fatal error:  Composer detected issues in your platform: Your Composer dependencies require a PHP version ">= 7.3.0". You are running 7.2.34-13+ubuntu18.04.1+deb.sury.org+1. in /srv/www/demo-sites/public_html/vendor/composer/platform_check.php on line 24
    default: Error: Could not process the 'wp-config.php' transformation.
    default: Reason: Unable to locate placement anchor.
```

Config:
```yaml
 demo-sites:
    nginx_upstream: php74
    hosts:
      - demosites.test
    db_name: demo-sites
    tideways: true
    custom:
      wp_type: subdirectory

utilities:
  core:
    - memcached-admin # Object cache management
    - opcache-status # opcache management
    - phpmyadmin # Web based database client
    - webgrind # PHP Debugging
    - tls-ca # SSL/TLS certificates
    - mongodb # needed for Tideways/XHGui
    - tideways # PHP profiling tool, also installs xhgui
    - wpcli-dev # Install a dev environment to use and contribute to WP-CLI
    - php # Install all the php utilities available automatically
```

## Possible Solution
Have the set PHP version by `nginx_upstream`, to be used during its provision

## Steps to Reproduce (for bugs)
1. `vagrant reload --provision`

## Your Environment
```
__ __ __ __
\ V\ V\ V / Varying Vagrant Vagrants
 \_/\_/\_/  v3.5.1-git::stable

Platform: platform-darwin19.2.0 shell:/bin/zsh vagrant-goodhosts vagrant-disksize shared_db_folder_disabled, VVV Path: "/Users/spenserhale/vagrant-local"
Vagrant: v2.2.14, VirtualBox: v6.1.18r142142
```

Mte90:
[...]

PR #2583, opened 2022-03-04
feat(php): Set PHP version per site

Mte90:
Ref: https://github.com/Varying-Vagrant-Vagrants/VVV/issues/2367
Part of: https://github.com/Varying-Vagrant-Vagrants/varyingvagrantvagrants.org/pull/207

## Checks

* [x] I've tested this PR

### DIFF
--- a/config/default-config.yml
+++ b/config/default-config.yml
@@ -36,6 +36,7 @@ sites:
     skip_provisioning: false
     description: "A standard WP install, useful for building plugins, testing things, etc"
     repo: https://github.com/Varying-Vagrant-Vagrants/custom-site-template.git
+    php: 7.4 # change the PHP version to use for the provision and nginx
     custom:
       # locale: it_IT
       delete_default_plugins: true
--- a/provision/provision-site.sh
+++ b/provision/provision-site.sh
@@ -124,7 +124,7 @@ function vvv_provision_site_nginx_config() {
   sed -i "s#{vvv_site_name}#${SITE_NAME}#"  "${TMPFILE}"
   sed -i "s#{vvv_hosts}#${VVV_HOSTS}#"  "${TMPFILE}"
 
-  # if nginx upstream is not 'php' then it has been changed in config.yml
+  # check if the nginx upstream value has been set and is valid
   if [ 'php' != "${NGINX_UPSTREAM}" ] && [ ! -f "/etc/nginx/upstreams/${NGINX_UPSTREAM}.conf" ]; then
     vvv_error " * Upstream value '${NGINX_UPSTREAM}' doesn't match a valid upstream. Defaulting to 'php'.${CRESET}"
     NGINX_UPSTREAM='php'
--- a/provision/provision-site.sh
+++ b/provision/provision-site.sh
@@ -130,12 +130,14 @@ function vvv_provision_site_nginx_config() {
   sed -i "s#{vvv_site_name}#${SITE_NAME}#"  "${TMPFILE}"
   sed -i "s#{vvv_hosts}#${VVV_HOSTS}#"  "${TMPFILE}"
 
+  # if nginx upstream is not 'php' then it has been changed in config.yml
   if [ 'php' != "${NGINX_UPSTREAM}" ] && [ ! -f "/etc/nginx/upstreams/${NGINX_UPSTREAM}.conf" ]; then
     vvv_error " * Upstream value '${NGINX_UPSTREAM}' doesn't match a valid upstream. Defaulting to 'php'.${CRESET}"
     NGINX_UPSTREAM='php'
-    DEFAULTPHP=$(vvv_get_site_config_value 'php' "")
+    DEFAULTPHP=$(vvv_get_site_config_value 'php' "" | tr --delete .)
+    # if php: is configured, set the upstream to match
     if [ '' != "${DEFAULTPHP}" ]; then
-      NGINX_UPSTREAM=$DEFAULTPHP
+      NGINX_UPSTREAM="php${DEFAULTPHP}"
     fi
   fi
   sed -i "s#{upstream}#${NGINX_UPSTREAM}#"  "${TMPFILE}"
--- a/provision/provision-site.sh
+++ b/provision/provision-site.sh
@@ -46,7 +46,7 @@ function vvv_set_php_cli_version() {
   php_version=$(readlink -f /usr/bin/php)
   DEFAULTPHP=$(vvv_get_site_config_value 'php' "${DEFAULTPHP}")
   if [[ $php_version != *"${DEFAULTPHP}"* ]]; then
-    length=$(echo "$DEFAULTPHP" | wc -c)
+    length=$(echo "$DEFAULTPHP" | wc -c  | tr '\n' ''  | tr '\r' '')
     if [[ $length > '3' ]]; then
       vvv_warn " ! Warning: PHP version defined is using a wrong format: '${DEFAULTPHP}' with length '${length}'"
     else
--- a/provision/provision-site.sh
+++ b/provision/provision-site.sh
@@ -23,6 +23,7 @@ NGINX_UPSTREAM=$6
 VVV_PATH_TO_SITE=${VM_DIR} # used in site templates
 VVV_SITE_NAME=${SITE}
 VVV_HOSTS=""
+SUCCESS=1
 
 VVV_CONFIG=/vagrant/config.yml
 
@@ -37,27 +38,20 @@ function vvv_get_site_config_value() {
   echo "${value}"
 }
 
-local DEFAULTPHP=$(vvv_get_site_config_value 'php' "${VVV_BASE_PHPVERSION}")
-echo " * Setting the default PHP CLI version ( ${DEFAULTPHP} ) for this site"
-update-alternatives --set php "/usr/bin/php${DEFAULTPHP}"
-update-alternatives --set phar "/usr/bin/phar${DEFAULTPHP}"
-update-alternatives --set phar.phar "/usr/bin/phar.phar${DEFAULTPHP}"
-update-alternatives --set phpize "/usr/bin/phpize${DEFAULTPHP}"
-update-alternatives --set php-config "/usr/bin/php-config${DEFAULTPHP}"
-
-SUCCESS=1
-
 # @description Reset the PHP global version to the default
+#
+# @internal
+# @noargs
 function vvv_set_php_cli_version() {
   php_version=$(readlink -f /usr/bin/php)
-  if [[ $php_version != *"${VVV_BASE_PHPVERSION}"* ]]; then
-    DEFAULTPHP=$(vvv_get_site_config_value 'php' "${VVV_BASE_PHPVERSION}")
+  DEFAULTPHP=$(vvv_get_site_config_value 'php' "${DEFAULTPHP}")
+  if [[ $php_version != *"${DEFAULTPHP}"* ]]; then
     echo " * Setting the default PHP CLI version ( ${DEFAULTPHP} ) for this site"
-    update-alternatives --set php "/usr/bin/php${DEFAULTPHP}"
-    update-alternatives --set phar "/usr/bin/phar${DEFAULTPHP}"
-    update-alternatives --set phar.phar "/usr/bin/phar.phar${DEFAULTPHP}"
-    update-alternatives --set phpize "/usr/bin/phpize${DEFAULTPHP}"
-    update-alternatives --set php-config "/usr/bin/php-config${DEFAULTPHP}"
+    update-alternatives --set php "/usr/bin/php${DEFAULTPHP}" &> /dev/null
+    update-alternatives --set phar "/usr/bin/phar${DEFAULTPHP}" &> /dev/null
+    update-alternatives --set phar.phar "/usr/bin/phar.phar${DEFAULTPHP}" &> /dev/null
+    update-alternatives --set phpize "/usr/bin/phpize${DEFAULTPHP}" &> /dev/null
+    update-alternatives --set php-config "/usr/bin/php-config${DEFAULTPHP}" &> /dev/null
   fi
 }
 
@@ -134,11 +128,12 @@ function vvv_provision_site_nginx_config() {
   if [ 'php' != "${NGINX_UPSTREAM}" ] && [ ! -f "/etc/nginx/upstreams/${NGINX_UPSTREAM}.conf" ]; then
     vvv_error " * Upstream value '${NGINX_UPSTREAM}' doesn't match a valid upstream. Defaulting to 'php'.${CRESET}"
     NGINX_UPSTREAM='php'
-    DEFAULTPHP=$(vvv_get_site_config_value 'php' "" | tr --delete .)
-    # if php: is configured, set the upstream to match
-    if [ '' != "${DEFAULTPHP}" ]; then
-      NGINX_UPSTREAM="php${DEFAULTPHP}"
-    fi
+  fi
+
+  # if php: is configured, set the upstream to match
+  if [ "${DEFAULTPHP}" != "${VVV_DEFAULTPHP}" ]; then
+    NGINX_UPSTREAM="php${DEFAULTPHP}"
+    NGINX_UPSTREAM=$(echo $NGINX_UPSTREAM | tr --delete .)
   fi
   sed -i "s#{upstream}#${NGINX_UPSTREAM}#"  "${TMPFILE}"
 
@@ -513,7 +508,9 @@ function vvv_custom_folders() {
 }
 
 # -------------------------------
-vvv_set_php_cli_version()
+source /srv/config/homebin/vvv_restore_php_default
+VVV_DEFAULTPHP=$DEFAULT_PHP
+vvv_set_php_cli_version
 
 if [[ true == "${SKIP_PROVISIONING}" ]]; then
   vvv_warn " * Skipping provisioning of <b>${SITE}</b>"
--- a/provision/provision-site.sh
+++ b/provision/provision-site.sh
@@ -129,17 +129,18 @@ function vvv_provision_site_nginx_config() {
   sed -i "s#{vvv_site_name}#${SITE_NAME}#"  "${TMPFILE}"
   sed -i "s#{vvv_hosts}#${VVV_HOSTS}#"  "${TMPFILE}"
 
+  # if php: is configured, set the upstream to match
+  if [ "${DEFAULTPHP}" != "${VVV_DEFAULTPHP}" ]; then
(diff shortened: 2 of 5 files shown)


Mte90:
It isn't yet ready, after the first on set the php version I get:
```
update-alternatives: error: alternative /usr/bin/php for php not registered; not setting
```

On it

Mte90:
Now seems working, now check also if the same version is already configured and avoid to set it again at the same version

config/default-config.yml
```diff
@@ -36,6 +36,8 @@ sites:
     skip_provisioning: false
     description: "A standard WP install, useful for building plugins, testing things, etc"
     repo: https://github.com/Varying-Vagrant-Vagrants/custom-site-template.git
+    php: 7.4 # change the PHP version to use for the provision
+    nginx_upstream: 7.4 # change the php version to use for nginx
```
tomjn, line 40: the hope is we can remove the nginx upstream because PHP was set to 7.4, as a result the provisioner can infer what the upstream value should be

provision/provision-site.sh
```diff
@@ -30,6 +30,26 @@ VVV_CONFIG=/vagrant/config.yml
...
 
+# @description Retrieves a config value for the given site as specified in `config.yml`
+#
+# @arg $1 string the config value to fetch
+# @arg $2 string the default value
+function vvv_get_site_config_value() {
+  local value=$(shyaml -q get-value "sites.${SITE_ESCAPED}.${1}" "${2}" < ${VVV_CONFIG})
+  echo "${value}"
+}
+php_version=$(readlink -f /usr/bin/php)
```
tomjn, line 41: It's possible that a provisioner somewhere else either failed or interfered with this, so we should always change the PHP version used if it's not the default version, e.g. if a previous site switched the PHP version to 8.0 then failed it would test against 8.0 rather than the default of 7.4

provision/provision-site.sh
```diff
@@ -30,6 +30,26 @@ VVV_CONFIG=/vagrant/config.yml
...
+if [[ $php_version != *"${VVV_BASE_PHPVERSION}"* ]]; then
+  DEFAULTPHP=$(vvv_get_site_config_value 'php' "${VVV_BASE_PHPVERSION}")
+  echo " * Setting the default PHP CLI version ( ${DEFAULTPHP} ) for this site"
+  noroot update-alternatives --set php "/usr/bin/php${DEFAULTPHP}"
+  noroot update-alternatives --set phar "/usr/bin/phar${DEFAULTPHP}"
+  noroot update-alternatives --set phar.phar "/usr/bin/phar.phar${DEFAULTPHP}"
+  noroot update-alternatives --set phpize "/usr/bin/phpize${DEFAULTPHP}"
+  noroot update-alternatives --set php-config "/usr/bin/php-config${DEFAULTPHP}"
```
tomjn, line 49: does `noroot` make sense here? We run these commands a lot, maybe it's time to turn them into functions, perhaps:

`vvv_set_php_cli_version` and `vvv_reset_php_cli_version`, where the following behaviour happens:

```sh
vvv_set_php_cli_version "7.2" # php is now at 7.2
vvv_reset_php_cli_version # php is now at the default 7.4
vvv_set_php_cli_version  # no parameter so it calls vvv_reset_php_cli_version
```

config/default-config.yml
Mte90: Right now they are 2 different values but I am not sure what this feedback is just a note or something to do

provision/provision-site.sh
```diff
@@ -504,4 +516,8 @@ if [ "${SUCCESS}" -ne "0" ]; then
   exit 1
 fi
 
+if [[ $php_version != *"${VVV_BASE_PHPVERSION}"* ]]; then
+  noroot /srv/config/homebin/vvv_restore_php_default
```
tomjn, line 520: unsure about the `noroot`, but it looks like one of the commands from my previous feedback! Great minds think alike :)

tomjn: Maybe this check should be inside `vvv_restore_php_default` though?

tomjn (changes requested):
some notes

provision/provision-site.sh
Mte90: the next line does this, check if the php version is different from the default otherwise reset it

config/default-config.yml
tomjn: I don't think we should mention the nginx upstream here, and it should work without it and just `php: 7.4`

tomjn: So in the future the docs would say just put `php: 8.0` and it would run the provisioners in 8.0 and the nginx config would also use PHP 8 instead of needing an nginx upstream value.

Then in a future release we can auto-run the PHP 8 extension before the sites provision based on what the config says

Mte90: So with a single value change both the parameters.
Maybe it's the case to open a ticket just for this.

provision/provision-site.sh
Mte90: I preferred to keep the function as you suggest but read the parameter by the config instead as parameter

tomjn: the concern is more that broken provisioning might break the expectations here causing more broken provisioning which just sounds painful to debug :(

config/default-config.yml
tomjn: but we did open a ticket for this, https://github.com/Varying-Vagrant-Vagrants/VVV/issues/2367 which is what this PR is meant to implement

tomjn: This is still a hard blocker and a major issue, it implies a misunderstanding of what #2367 was asking for. The whole point of the issue is that you no longer need to specify an Nginx upstream to change your PHP version, you only specify an upstream if you specifically and only want to change the upstream, not because you want to change the PHP version used.

provision/provision-site.sh
```diff
@@ -30,6 +30,29 @@ VVV_CONFIG=/vagrant/config.yml
...
+#
+# @arg $1 string the config value to fetch
+# @arg $2 string the default value
+function vvv_get_site_config_value() {
+  local value=$(shyaml -q get-value "sites.${SITE_ESCAPED}.${1}" "${2}" < ${VVV_CONFIG})
+  echo "${value}"
+}
+
+# @description Reset the PHP global version to the default
+function vvv_set_php_cli_version() {
```
tomjn, line 43: This function is not indented correctly

tomjn (changes requested):
I'm seeing no adjustment of the nginx upstream here, this is all I should need to specify that a site uses PHP 8.0:

```yaml
  phpeight:
    php: 8.0
    hosts:
      - phpeight.test
```
But, this PR only changes the PHP version the provisioner uses, and does not touch the Nginx upstream.

 - If I create a site, the nginx upstream will use the default PHP of VVV
 - if I create a site with `php: 8.0` the nginx upstream will use PHP 8.0
 - if i create a site with an `nginx_upstream: abc` it will use `abc` even if `php: 8.0` is set

```
nginx upstream = default PHP
if php is set then
    nginx upstream = php{php config value}
if nginx upstream is set then
   nginx upstream = {nginx upstream config value}
```

config/default-config.yml
Mte90: The problem with this is: what if I want that the provision is executed with a php version and nginx with another one?

Mte90:
Fixed the linting and removed the example for nginx_upstream parameter in default-config.yml.
Also if nginx_upstream is not set and php parameter is set use that value for nginx_upstream (I have to test it).

config/default-config.yml
tomjn: then it would use the nginx upstream in the config because the user has explicitly chosen the value.

 - `php` tells VVV which version of PHP to use for the site
 - `nginx_upstream` allows you to override a value in the nginx config

Right now we use `nginx_upstream` to set the version of PHP used on frontend requests because no other options exist, it's a workaround. This issue is meant to fix that by providing a `php:` parameter

provision/provision-site.sh
```diff
@@ -102,6 +133,10 @@ function vvv_provision_site_nginx_config() {
   if [ 'php' != "${NGINX_UPSTREAM}" ] && [ ! -f "/etc/nginx/upstreams/${NGINX_UPSTREAM}.conf" ]; then
     vvv_error " * Upstream value '${NGINX_UPSTREAM}' doesn't match a valid upstream. Defaulting to 'php'.${CRESET}"
     NGINX_UPSTREAM='php'
+    DEFAULTPHP=$(vvv_get_site_config_value 'php' "")
+    if [ '' != "${DEFAULTPHP}" ]; then
+      NGINX_UPSTREAM=$DEFAULTPHP
+    fi
   fi
   sed -i "s#{upstream}#${NGINX_UPSTREAM}#"  "${TMPFILE}"
```
tomjn, line 141: I don't think this will work, if no `nginx_upstream` is provided then its value will default to `php` and none of this code will run. This code is checking if the nginx upstream the user has specified in the config actually exists, and resets it to `php` if it does not

provision/provision-site.sh
```diff
@@ -102,6 +133,10 @@ function vvv_provision_site_nginx_config() {
   if [ 'php' != "${NGINX_UPSTREAM}" ] && [ ! -f "/etc/nginx/upstreams/${NGINX_UPSTREAM}.conf" ]; then
```
tomjn, line 133:
```suggestion
  # if nginx upstream is not 'php' then it has been changed in config.yml
  if [ 'php' != "${NGINX_UPSTREAM}" ] && [ ! -f "/etc/nginx/upstreams/${NGINX_UPSTREAM}.conf" ]; then
```

provision/provision-site.sh
```diff
@@ -102,6 +133,10 @@ function vvv_provision_site_nginx_config() {
...
     vvv_error " * Upstream value '${NGINX_UPSTREAM}' doesn't match a valid upstream. Defaulting to 'php'.${CRESET}"
     NGINX_UPSTREAM='php'
+    DEFAULTPHP=$(vvv_get_site_config_value 'php' "")
+    if [ '' != "${DEFAULTPHP}" ]; then
+      NGINX_UPSTREAM=$DEFAULTPHP
+    fi
   fi
```
tomjn, line 140:
```suggestion
  else
    # if php: is configured, set the upstream to match
    if [ '' != "${DEFAULTPHP}" ]; then
      # TODO: Check this version of PHP is actually installed
      NGINX_UPSTREAM="${DEFAULTPHP}"
    fi
  fi
```

tomjn: also, aren't PHP versions of the form `8.0` but nginx upstreams of the form `php80`? We have to do some conversion

provision/provision-site.sh
```diff
@@ -484,6 +511,7 @@ function vvv_custom_folders() {
 }
 
 # -------------------------------
+vvv_set_php_cli_version()
```
tomjn, line 514: needs a doc block

provision/provision-site.sh
Mte90: you are right, fixed in a commit

provision/provision-site.sh
tomjn: but no docblock, so it won't have auto-generated documentation when the docs site is updated

tomjn: See https://github.com/Varying-Vagrant-Vagrants/varyingvagrantvagrants.org/blob/main/inject-repo-documentation.sh and https://github.com/reconquest/shdoc

tomjn: and https://varyingvagrantvagrants.org/docs/en-US/function-reference/provision-provision-site/

Mte90:
Fixed the issues you reported, I am not sure if I fixed the doc stuff.

provision/provision-site.sh
```diff
@@ -23,13 +23,38 @@ NGINX_UPSTREAM=$6
...
+# @description Retrieves a config value for the given site as specified in `config.yml`
+#
+# @arg $1 string the config value to fetch
+# @arg $2 string the default value
+function vvv_get_site_config_value() {
+  local value=$(shyaml -q get-value "sites.${SITE_ESCAPED}.${1}" "${2}" < ${VVV_CONFIG})
+  echo "${value}"
+}
+
+# @description Reset the PHP global version to the default
```
tomjn, line 41: not quite true as it checks the site config value, so if the default is 7.4 but my site says `php: 8.0` then this function sets PHP to v8, not v7.4

tomjn:
```suggestion
# @description sets a sites PHP version as the global version, or to the VVV default if none is specified
```

provision/provision-site.sh
```diff
@@ -99,10 +124,17 @@ function vvv_provision_site_nginx_config() {
   sed -i "s#{vvv_site_name}#${SITE_NAME}#"  "${TMPFILE}"
   sed -i "s#{vvv_hosts}#${VVV_HOSTS}#"  "${TMPFILE}"
 
+  # if nginx upstream is not 'php' then it has been changed in config.yml
```
tomjn, line 127:
```suggestion
  # check if the nginx upstream value has been set and is valid
```

provision/provision-site.sh
```diff
@@ -99,10 +124,17 @@ function vvv_provision_site_nginx_config() {
...
   if [ 'php' != "${NGINX_UPSTREAM}" ] && [ ! -f "/etc/nginx/upstreams/${NGINX_UPSTREAM}.conf" ]; then
     vvv_error " * Upstream value '${NGINX_UPSTREAM}' doesn't match a valid upstream. Defaulting to 'php'.${CRESET}"
     NGINX_UPSTREAM='php'
   fi
+
+  # if php: is configured, set the upstream to match
+  if [ "${DEFAULTPHP}" != "${VVV_DEFAULTPHP}" ]; then
+    NGINX_UPSTREAM="php${DEFAULTPHP}"
+    NGINX_UPSTREAM=$(echo $NGINX_UPSTREAM | tr --delete .)
+  fi
```
tomjn, line 137: this should happen beforehand otherwise I could do the following:

```yaml
    php: 74.0
```
And Nginx would fail to load as there is no PHP v74.0

tomjn (changes requested):
Some small changes to make it handle failure gracefully if a user puts a silly value in like `v800.0` or `php: banana`

provision/provision-site.sh
Mte90: ops I switched the lines

Mte90:
Added a version validation that check if the string has a length different from 3. I am not sure if this way is the best one anyway (require some testing).

tomjn:
Things look good, I need to test before I can approve but I think this PR is the priority for releasing 3.10.1

tomjn:
Looks like there are some CI issues finding functions in site templates:

```
    default:  * wordpress-one provisioner clone successful
    default: PHP Notice:  Undefined index: HTTP_HOST in /srv/www/wordpress-one/public_html/wp-includes/functions.php on line 6052
    default: Notice: Undefined index: HTTP_HOST in /srv/www/wordpress-one/public_html/wp-includes/functions.php on line 6052
    default:  * sourcing of vvv-init.sh reported success
    default:  * VVV is adding an Nginx config from /srv/www/wordpress-one/provision/vvv-nginx.conf
    default: nginx: the configuration file /etc/nginx/nginx.conf syntax is ok
    default: nginx: configuration file /etc/nginx/nginx.conf test is successful
    default: /tmp/vagrant-shell: line 556: vvv_restore_php_default: command not found
    default:  ! The 'site-wordpress-one' provisioner ran into problems, the full log is available at '/var/log/provisioners/2022.08.01_12-27-27/provisioner-site-wordpress-one.log'. It completed in 17 seconds.
```

Mte90:
I am not able to replicate your issue maybe in the PR the permissions on that file are wrong?

In the meantime I found another bug and fixed it.

tomjn:
Local tests failed with:

```
    default:  ✔ The 'site-wordpress-two' provisioner completed in 2 seconds.
==> default: Running provisioner: site-phpeight (shell)...
    default: Running: /var/folders/dw/wdtnsdn154db04ymxv90t9tm0000gn/T/vagrant-shell20220907-23209-1feiyc8.sh
    default:  ▷ Running the 'site-phpeight' provisioner...
    default: /tmp/vagrant-shell: line 51: vvv_warning: command not found
    default:  ! The 'site-phpeight' provisioner ran into problems, the full log is available at '/var/log/provisioners/2022.09.07_13-44-29/provisioner-site-phpeight.log'. It completed in 0 seconds.
went wrong.
```

tomjn:
Fixed a `vvv_warning` -> `vvv_warn`, though `8.0` is probably being interpreted as a number as it considers it the wrong format. We can't expect users to always wrap it in quotes so some string manipulation will be needed to format the value in the 0.0 format

tomjn (changes requested):
Needs to be able to handle numbers in the value such as `8.1` or `8.0` without wrapping in string quotes e.g. `"8.0"`

tomjn:
I changed it to:

```
    length=$(echo "$DEFAULTPHP" | wc -c)
    if [[ $length > '3' ]]; then
      vvv_warn " ! Warning: PHP version defined is using a wrong format: '${DEFAULTPHP}' with length '${length}'"
    else
```

and got:

```
    default:  ▷ Running the 'site-phpeight' provisioner...
    default:  ! Warning: PHP version defined is using a wrong format: '8.0' with length '4'
    default:  * Pulling down the master branch of https://github.com/Varying-Vagrant-Vagrants/custom-site-template.git
```

Looks like there's a newline or terminating character that isn't being handled. Incrementing `3` to `4` leads to an unhandled error:

```
    default: Running: /var/folders/dw/wdtnsdn154db04ymxv90t9tm0000gn/T/vagrant-shell20220907-30221-7dsbtn.sh
    default:  ▷ Running the 'site-phpeight' provisioner...
    default:  ! The 'site-phpeight' provisioner ran into problems, the full log is available at '/var/log/provisioners/2022.09.07_13-50-27/provisioner-site-phpeight.log'. It completed in 0 seconds.
went wrong.

```

Mte90:
I added the removal of hidden breaklines but I wasn't able to find a way to validate a "version string" expect for the length.

tomjn:
@Mte90 would it make more sense to validate the presence of that version of PHP rather than the validity of the version string?

e.g. it doesn't matter if `php : Octopus` is a valid because there'll never be a `phpOctupus` nginx upstream or an `octopus` folder and CLI

Mte90:
Well this simplify a lot but this means that they need to have already provisioned to be able to have that PHP version in VVV.
Anyway is better of my fake solution.

tomjn:
@Mte90 extensions always provision before sites so it'll be installed beforehand, but if it is missing it'll just revert to default PHP

tomjn (approved):

Issue #2525, opened 2021-09-13
Bug: MariaDB 10.4 -> 10.5 upgrade


tomjn:
We upgraded to v10.5 but it seems this is not a clean upgrade. If you have v10.4 installed then you need to stop the service and remove it before installing 10.5, which we do not do

tomjn:
The main roadblock to this is an automated check to identify when 10.4 is installed. Most research shows you can get a version string via an SQL query but there are no clean methods of getting _just_ the version

e.g.

```
vagrant@vvv:~$ mysql -V
mysql  Ver 15.1 Distrib 10.5.12-MariaDB, for debian-linux-gnu (x86_64) using readline 5.2
```

Ideally we would get `10.5.12` as the output

tomjn:
re-adding the 10.4 mirrors then pinning the 10.5 mirrors at a higher priority might help

ssnepenthe:
> The main roadblock to this is an automated check to identify when 10.4 is installed. Most research shows you can get a version string via an SQL query but there are no clean methods of getting just the version

After a quick look through the mariadb string functions page I wonder if running the version string through the SUBSTRING_INDEX function would be sufficient?

```shell
vagrant@vvv:~$ mysql -BN -e "SELECT VERSION();"
10.5.12-MariaDB-1:10.5.12+maria~bionic

vagrant@vvv:~$ mysql -BN -e "SELECT SUBSTRING_INDEX((SELECT VERSION()), '-', 1);"
10.5.12
```

tomjn:
This looks promising :)

The pseudocode/steps I envisage:

 - in the mariadb provisioner file in `provision/core/mariadb`
   - if the version is old
     - add it to the list of packages to remove

Pinning the 10.5 mirrors may still be something we want to do, but having a working check also means we can log that it's happening which makes life much easier for everybody

adionditsak:
When running `apt` in the provisioning or manually, I get the following error:
```    
default: Reading package lists...
    default: E
    default: : The repository 'http://sgp1.mirrors.digitalocean.com/mariadb/repo/10.5/ubuntu focal Release' does not have a Release file.
    default: E: The repository 'http://sfo1.mirrors.digitalocean.com/mariadb/repo/10.5/ubuntu focal Release' does not have a Release file.
    default: E: The repository 'http://ams2.mirrors.digitalocean.com/mariadb/repo/10.5/ubuntu focal Release' does not have a Release file. 
```

[...]

PR #2592, opened 2022-04-13
MariaDB APT fix

Mte90:
Ref: https://github.com/Varying-Vagrant-Vagrants/VVV/issues/2525

So the pinning of mariadb packages wasn't fixing the problem and investigating a bit is that we remove mysql-common but is required by mariadb to put aliases for mysql (and it is in the mariadb repository).
Removing that line let everything works with no problem.

## Checks

* [x] I've tested this PR

### DIFF
--- a/provision/core/vvv/provision.sh
+++ b/provision/core/vvv/provision.sh
@@ -8,9 +8,6 @@ function vvv_register_packages() {
     python-pip
     python-setuptools
 
-    # remove mysql-common to ensure mariadb installation works
-    mysql-common
-
     # remove nodesource js etc we have nvm for that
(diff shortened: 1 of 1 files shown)


tomjn:
@Mte90 nobody has implemented the MariaDB pinning changes yet, that's still on the todo list

Mte90:
> @Mte90 nobody has implemented the MariaDB pinning changes yet, that's still on the todo list

I tried implementing but in this case the pinning wasn't avoiding the package removal as mariadb are various packages, so we need to put all of them. This fixes avoided the removal at all of Mariadb.

Issue #2593, opened 2022-04-13
Unsafe repository ('/srv/provision/extensions/core' is owned by someone else)


Sriom:
## What was The Command Used To Provision

```shell
vagrant up --provision
```

## What Kind of VVV Provision Was This

This was a reprovision of an already working VVV

## Logs/What Broke

    default: fatal: unsafe repository ('/srv/provision/extensions/core' is owned by someone else)
    default: To add an exception for this directory, call:
    default: 
    default: 	git config --global --add safe.directory /srv/provision/extensions/core
    default:  ! The 'extension-source-core' provisioner ran into problems, the full log is available at '/var/log/provisioners/2022.04.13_11-26-23/provisioner-extension-source-core.log'. It completed in 0 seconds.
The SSH command responded with a non-zero exit status. Vagrant
assumes that this means the command failed. The output for this command
should be in the log above. Please read the output to determine what
went wrong.

```yaml
---

# This file is a YAML formatted file. YAML indenting is done in spaces not
# tabs, and whitespace is significant. If you don't stick to this, it will
# fail on provision

#
# IMPORTANT, if you change this file, you have to reprovision,  no exceptions
# Do this by running either this command:
# vagrant reload --provision

# Or, if your machine is already turned on:
# vagrant provision
#

# These are your websites, and their names map on to the folders they're
# located in. See the docs for how to define these, and what all the keys
# and options are
sites:

  # latest version of WordPress, can be used for client work and testing
  # Check the readme at https://github.com/Varying-Vagrant-Vagrants/custom-site-template
  wordpress-one:
    skip_provisioning: true
    description: "A standard WP install, useful for building plugins, testing things, etc"
    repo: https://github.com/Varying-Vagrant-Vagrants/custom-site-template.git
    hosts:
      - one.wordpress.test
    custom:
      wpconfig_constants:
        WP_DEBUG: true
        WP_DEBUG_LOG: true
        WP_DISABLE_FATAL_ERROR_HANDLER: true # To disable in WP 5.2 the FER mode

  wordpress-two:
    skip_provisioning: true
    description: "A standard WP install, useful for building plugins, testing things, etc"
    repo: https://github.com/Varying-Vagrant-Vagrants/custom-site-template.git
    custom:
      # locale: it_IT
      delete_default_plugins: true
      install_plugins:
        - query-monitor
    hosts:
      - two.wordpress.test

  WPCORE:
    skip_provisioning: true
[...]

PR #2595, opened 2022-04-13
Fix git issue

Mte90:
Ref: https://github.com/Varying-Vagrant-Vagrants/VVV/issues/2593

So this is a test that is adding the repository in the safe list without fixing the permission itself

## Checks

* [ ] I've tested this PR

### DIFF
--- a/provision/provision-extension-source.sh
+++ b/provision/provision-extension-source.sh
@@ -12,6 +12,7 @@ if [[ false != "${NAME}" && false != "${REPO}" ]]; then
   # Clone or pull the extension repository
   if [[ ! -d "${DIR}/.git" ]]; then
     vvv_info " * Cloning the \"${NAME}\" extension, see \"${REPO}\""
+    git config --global --add safe.directory "${DIR}"
     git clone "${REPO}" --branch "${BRANCH}" "${DIR}" -q
     cd "${DIR}"
     git checkout "${BRANCH}" -q


tomjn:
we would need this in the site provisioner too, it'll probably impact some utilities provisioners as well, is it not possible to run the command with the necessary user and ownership to avoid needing this?

Mte90:
> we would need this in the site provisioner too, it'll probably impact some utilities provisioners as well, is it not possible to run the command with the necessary user and ownership to avoid needing this?

I want to see what happen in the CI before to do other changes.

provision/provision-extension-source.sh
```diff
@@ -12,6 +12,7 @@ if [[ false != "${NAME}" && false != "${REPO}" ]]; then
   # Clone or pull the extension repository
   if [[ ! -d "${DIR}/.git" ]]; then
     vvv_info " * Cloning the \"${NAME}\" extension, see \"${REPO}\""
+    git config --global --add safe.directory "${DIR}"
     git clone "${REPO}" --branch "${BRANCH}" "${DIR}" -q
```
tomjn, line 16: wouldn't `noroot git clone` do the job? Or `mkdir` then `chown` then `noroot git`? It's checking if DIR is owned by the current user, seeing it is not, and exiting

Mte90: My guess is that will work on new instances but on old will works as always? They can have different permissions.

tomjn: on virtualbox at least, the folder will have the ownership of the shared mount

tomjn, line 16:
```suggestion
    noroot git clone "${REPO}" --branch "${BRANCH}" "${DIR}" -q
```

tomjn: likewise the two checkout and the pull command should have `noroot`

Mte90:
So with safe works let's see with `noroot`.

Mte90:
So seems that this PR fix that issue.

tomjn (dismissed):

tomjn:
so other provisioners already do this and aren't affected, at least in the main VVV git repo

tomjn (dismissed):

Mte90:
I see that error, working on a fix:
```
    default: fatal: unsafe repository ('/home/vagrant/.nvm' is owned by someone else)
    default: To add an exception for this directory, call:
    default: 
    default:    git config --global --add safe.directory /home/vagrant/.nvm
    default: fatal: unsafe repository ('/home/vagrant/.nvm' is owned by someone else)
    default: To add an exception for this directory, call:
    default: 
    default:    git config --global --add safe.directory /home/vagrant/.nvm
```

tomjn (approved):

Issue #2600, opened 2022-04-20
Error with tools provisioner when installing WP CLI


unfulvio:
I have updated `develop` branch to use 3.9.1 recently (can't remember earlier version but it wasn't much older).

I wanted to reprovision my sites after that, but I stumbled into an error and now I'm stuck with all sites not working due not being correctly provisioned.

## What was The Command Used To Provision

```shell
vagrant up --provision
```

or 

```shell
vagrant reload --provision
``` 

## What Kind of VVV Provision Was This

- This was an attempt to upgrade to a new version of VVV
- This was a reprovision of an already working VVV

## Logs/What Broke

While running the commands above the provisioner always halts with this error:

```
    default:  * [WP-CLI]: Updating packages
    default: Error: Failed to update packages (Composer return code 2).
! The 'tools' provisioner ran into problems, the full log is available at '/var/log/provisioners/2022.04.20_06-30-03/provisioner-tools.log'. It completed in 8 seconds.
The SSH command responded with a non-zero exit status. Vagrant
assumes that this means the command failed. The output for this command
should be in the log above. Please read the output to determine what
went wrong.
```

The logs would then contain the following:

```
Problem 1
    - Root composer.json requires wp-cli/doctor-command dev-master, it is found wp-cli/doctor-command[dev-master] in the lock file and wp-cli/doctor-command[dev-fix/adapt-branch-alias, dev-main, v0.1.0, v2.0.0, v2.1.0] from vcs repo (github git@github.com:wp-cli/doctor-command.git) but these do not match your constraint and are therefore not installable. Make sure you either fix the constraint or avoid updating this package to keep the one from the lock file.

Use the option --with-all-dependencies (-W) to allow upgrades, downgrades and removals for packages currently locked to specific versions.
Running update with --no-dev does not mean require-dev is ignored, it just means the packages will not be installed. If dev requirements are blocking the update you have to resolve those problems.
---
Error: Failed to update packages (Composer return code 2).
 ! The 'tools' provisioner ran into problems, the full log is available at '/var/log/provisioners/2022.04.20_06-30-03/provisioner-tools.log'. It completed in 8 seconds.
```

The following is my config yml (note that I have tried disabling customfile will not change outcome - I was using customfile before to allow NFS shares and nothing else).

```yaml
---
sites:
  
  skyverge:
    skip_provisioning: false
    description: "SkyVerge Plugin Development"
[...]

PR #2601, opened 2022-04-20
Adjust wp cli package updating so it doesn't fail the provisioner

tomjn:
Related to #2600, this doesn't fix WP CLI Doctor updating, but it means the provisioner will continue to run so it's a soft failure rather than a hard failure

## Checks

* [ ] I've tested this PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ permalink: /docs/en-US/changelog/
 
 # Changelog
 
+## 3.10 ( 2022 TBD )
+
+### Enhancements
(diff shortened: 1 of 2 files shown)


tomjn:
@unfulvio this PR may be interesting to you, I'd be very happy if you could help test it!

unfulvio:
hey @tomjn this looks right to me - it may help pinning down the error or future WP CLI issues probably
I will still open that doc PR just in case

atticus7:
I've tested this PR and it works

PR #2605, opened 2022-04-29
Add/nginx config install testing

### LINKED ISSUE #2604: Gracefully handle nginx config install failures
When a user puts a typo in their custom nginx configs and reprovisions, all reprovisioning fails.

Instead, it should attempt to use the config, see that it fails, run `nginx -t` and remove the config so that the dashboard and other sites are still reachable. This also opens the possibility that we can continue provisioning and log the error somewhere and show it in the dashboard

tomjn:
Closes #2604

## Checks

* [ ] I've tested this PR

### DIFF
--- a/provision/provision-helpers.sh
+++ b/provision/provision-helpers.sh
@@ -519,7 +519,7 @@ vvv_package_install() {
 
   return 0
 }
-export -f vvv_package_install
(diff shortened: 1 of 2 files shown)


tomjn:
one place I haven't finished testing yet is broken nginx configs

PR #2606, opened 2022-05-02
Add ruby version in logs

Mte90:
## Checks

* [x] I've tested this PR


tomjn:
Could this be underneath? Abs does it need the word version? Eg Ruby:2.6

tomjn:
I was thinking alongside all the other debug items under the logo rather than next to it

PR #2608, opened 2022-05-16
Improve db scripts

msaggiorato:
This is a _backport from the port_ I've made of this scripts from VVV into a docker solution we're currently using (and I intend to slowly port back to VVV).

Basically allows to:

### Allow to restore specific databases with `db_restore`

By running `db_restore db_name_1 db_name_2` you can only restore specific databases only.

### Force restoration when running `db_restore`

This allows to run `db_restore --force` or `db_restore -f` to delete all databases being restored before restoring them. Allows to easily force restore to a dump. Most useful in conjunction with specific databases usage

### Minor other improvements

* Skip `sys` database when backing up or restoring.
* Root user now also has .my.cnf file in their home dir which grants mysql default user password (root / root).
* Due to the above, all mysql commands run in provisioners do not require user and password. Removed.
* When backing up, count should start at 1 (not 0)
* Normalized wording of exclude/skip on the restore script.

Didn't test this myself on VVV (because i cannot run it right now). Will see how the provisioner goes, but none of these changes should be breaking.

Once we agree if this is to be merged, I'll update the CHANGELOG.md file.

## Checks

* [ ] I've tested this PR

### DIFF
--- a/config/homebin/db_backup
+++ b/config/homebin/db_backup
@@ -53,7 +53,7 @@ ext=".sql"
 if [[ "${gzip}" == "True" ]]; then
     ext=".sql.gz"
 fi
-count=0
+count=1
 for db in "${databases[@]}"
 do
     OUTPUT=$(printf "<info>   - %2s/%s Backing up </info><b>%-23s</b><info> to </info><b>'database/backups/%s${ext}'</b>" "${count}" "${#databases[@]}" "'${db}'" "${db}")
--- a/config/homebin/db_backup
+++ b/config/homebin/db_backup
@@ -4,6 +4,7 @@
 # are imported automatically during an initial provision if
 # the databases exist per the import-sql.sh process.
 set -eo pipefail
+set -u
 
 trap 'rm -rf $TMPFIFODIR' EXIT; TMPFIFODIR=$(mktemp -d); mkfifo $TMPFIFODIR/dbnames
 
--- a/config/homebin/db_backup
+++ b/config/homebin/db_backup
@@ -35,7 +35,7 @@ while read db_name; do
     done
 
     if [ ${skip} == "true" ]; then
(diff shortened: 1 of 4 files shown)


provision/core/mariadb/provision.sh
```diff
@@ -106,6 +106,10 @@ function mysql_setup() {
   cp -f "/srv/provision/core/mariadb/config/my.cnf" "/etc/mysql/my.cnf"
   vvv_info " * Copied /srv/provision/core/mariadb/config/my.cnf               to /etc/mysql/my.cnf"
 
+  cp -f  "/srv/provision/core/mariadb/config/root-my.cnf" "/root/.my.cnf"
```
tomjn, line 109: @msaggiorato did you forgot to commit `root-my.cnf`?

tomjn:
Not sure why the provisioners are failing with `/srv/provision/provision-helpers.sh: line 314: $2: unbound variable
` but probably related to `set -u`

provision/core/mariadb/provision.sh
Mte90: yes the file is missing :-)

msaggiorato: It's not a new file, it's the same file copied into `/home/vagrant/.my.cnf`

msaggiorato:
Fixed undefined variable issues

config/homebin/db_backup
```diff
@@ -53,15 +53,15 @@ ext=".sql"
 if [[ "${gzip}" == "True" ]]; then
     ext=".sql.gz"
 fi
-count=0
+count=1
```
tomjn, line 56: Would this mean the count is always 1 or higher even if nothing got backed up?

msaggiorato: count starts at 1 if there's something, variable is not used at all if nothing goes through the loop

PR #2609, opened 2022-06-07
Disable gcrypt hardware optimization as we are a VM

Mte90:
This fix an issue with hash mismatch as disable the hardware optimization for hash calculation and use the software version that works always.

## Checks

* [x] I've tested this PR


provision/core/vvv/provision.sh
```diff
@@ -133,6 +133,9 @@ function apt_hash_missmatch_fix() {
     vvv_info " * Copying /srv/provision/core/vvv/apt-conf-d/99hashmismatch to /etc/apt/apt.conf.d/99hashmismatch"
     cp -f "/srv/provision/core/vvv/apt-conf-d/99hashmismatch" "/etc/apt/apt.conf.d/99hashmismatch"
   fi
+  # reference https://askubuntu.com/a/1242739
+  mkdir /etc/gcrypt
+  echo all >> /etc/gcrypt/hwf.deny
```
tomjn, line 138:
```suggestion
  fi

  # Avoid bad hardware implementations that interfere with gcrypt by disabling hardware support
  # reference https://askubuntu.com/a/1242739
  mkdir /etc/gcrypt
  echo all >> /etc/gcrypt/hwf.deny
```

PR #2635, opened 2022-09-27
Fix #2634 - adding full path to vvv_restore_php_default in provision-site.sh

### LINKED ISSUE #2634: Bug: vvv_restore_php_defaults needs full path in provision-site.sh or provision is unsuccessful.
### Are you using the latest stable or develop branch version of VVV?

Yes (develop)

### Is it a new VVV or an existing VVV that used to work?

Existing, worked but now broken

### Whats the problem?

This PR: https://github.com/Varying-Vagrant-Vagrants/VVV/pull/2583 introduced an error into the development branch that can clearly be seen in the CI provision test failure on that PR. I can reproduce the error locally:
 
`default: /tmp/vagrant-shell: line 578: vvv_restore_php_default: command not found
    default:  ! The 'site-*********' provisioner ran into problems, the full log is available at '/var/log/provisioners/2022.09.27_16-01-23/provisioner-site-*********.log'. It completed in 206 seconds.`

I do believe the problem is that /srv/config/homebin is not in the path when executing the provisioner scripts so line 578 fails when provision-site.sh attempts to execute vvv_restore_php_default: https://github.com/Varying-Vagrant-Vagrants/VVV/blob/468b77f9eb8d63b55d989119a66bd6a6c9547fd3/provision/provision-site.sh#L578

I am unclear, though, why this script must be run multiple times? The provision.sh file runs that same script regardless of whether the site was provisioned successfully or not: https://github.com/Varying-Vagrant-Vagrants/VVV/blob/468b77f9eb8d63b55d989119a66bd6a6c9547fd3/provision/provisioners.sh#L40

I ask whether the fix is to add the full path to that command on line 578 of provision-site.sh or remove it all together given that it runs in provisioner.sh? I have verified that setting the full path to vvv_restore_php_default on line 578 of provision-site.sh fixes the error and allows the provision to complete without exiting.

I want to add that this does not seem to break VVV, at least for the single site we have configured. But the provision does not succeed, exits right before the end, and erodes confidence that everything is set up correctly.

### How do we reproduce it?

Checkout the develop branch, copy default-config.yml to config.yml and run vagrant up --provision.

### VVV Status screen

```shell
__ __ __ __
\ V\ V\ V / v3.11 Ruby:2.6.6, Path:"/Users/dwashko/mygitstuff/storyeditor"
 \_/\_/\_/  git::develop(3a1d5e80)

Platform: darwin19.2.0 shell:/bin/bash vagrant-goodhosts shared_db_folder_disabled
Vagrant: v2.2.14, virtualbox: v6.1.32

Dashboard:  http://vvv.test

Current machine states:

default                   running (virtualbox)

The VM is running. To stop this VM, you can run `vagrant halt` to
shut it down forcefully, or you can run `vagrant suspend` to simply
suspend the virtual machine. In either case, to restart it again,
simply run `vagrant up`.
```

### Which Operating System are you using?

Apple MacOS (Intel)

### Which provider are you using?

VirtualBox

dwashko:
This PR adds the full path to the vvv_restore_php_default script because it is not in the path when the script is executed.

## Checks

* [X] I've tested this PR
* [X] This PR is for the `develop` branch not the `stable` branch.
* [X] This PR is complete and ready for review.

### DIFF
--- a/provision/provision-site.sh
+++ b/provision/provision-site.sh
@@ -575,6 +575,6 @@ if [ "${SUCCESS}" -ne "0" ]; then
   exit 1
 fi
 
-vvv_restore_php_default
+/srv/config/homebin/vvv_restore_php_default
 
 provisioner_success


tomjn:
Thanks :)

PR #2638, opened 2022-11-05
Fixing faulty Vagrant plugin check

aldavigdis:
This replaces some `Vagrant.has_plugin?` checks  in the Vagrantfile with `config.vagrant.plugins.include?` as the latter is not available when the `Vagrantfile` is run in version 2.1.4 and newer of Vagrant.

There are other `Vagrant.has_plugin?` calls in the Vagrant file, but they don't seem to affect things in the same way. I can mend that in a separate PR, but I am not sure about the effectiveness, even if `config.vagrant.plugins.include?` seems more reliable.

See also: https://github.com/hashicorp/vagrant/issues/10161

## Checks

* [x] I've tested this PR

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -851,29 +851,26 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   # located in the www/ directory and in config/config.yml.
   #
 
-  if Vagrant.has_plugin?('vagrant-goodhosts')
+  if config.vagrant.plugins.include? 'vagrant-goodhosts'
     config.goodhosts.aliases = vvv_config['hosts']
     config.goodhosts.remove_on_suspend = true
-  elsif Vagrant.has_plugin?('vagrant-hostsmanager')
+  elsif config.vagrant.plugins.include? 'vagrant-hostsmanager'
     config.hostmanager.aliases = vvv_config['hosts']
     config.hostmanager.enabled = true
     config.hostmanager.manage_host = true
     config.hostmanager.manage_guest = true
     config.hostmanager.ignore_private_ip = false
     config.hostmanager.include_offline = true
-  elsif Vagrant.has_plugin?('vagrant-hostsupdater')
+  elsif config.vagrant.plugins.include? 'vagrant-hostsupdater'
     # Pass the found host names to the hostsupdater plugin so it can perform magic.
     config.hostsupdater.aliases = vvv_config['hosts']
     config.hostsupdater.remove_on_suspend = true
-  else
-    show_check = true if %w[up halt resume suspend status provision reload].include? ARGV[0]
-    if show_check
-      puts ""
-      puts " X ! There is no hosts file vagrant plugin installed!"
-      puts " X You need the vagrant-goodhosts plugin (or HostManager/ HostsUpdater ) for domains to work in the browser"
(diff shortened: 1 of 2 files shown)


aldavigdis:
By the same logic, the `Vagrant.has_plugin?` calls in the `unless` statement in the block staring in line 376 will always return `false`, so VVV is always going to insist on `vagrant-goodhosts` being installed no matter if the other host management plugins are present.

Are `vagrant-hostsupdater` and `vagrant-hostsmanager` still supported and working?

Vagrantfile
```diff
@@ -851,29 +851,20 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
...
     config.hostsupdater.remove_on_suspend = true
-  else
-    show_check = true if %w[up halt resume suspend status provision reload].include? ARGV[0]
-    if show_check
-      puts ""
-      puts " X ! There is no hosts file vagrant plugin installed!"
-      puts " X You need the vagrant-goodhosts plugin (or HostManager/ HostsUpdater ) for domains to work in the browser"
-      puts " X Run 'vagrant plugin install --local' to fix this."
-      puts ""
-    end
```
tomjn, line 876: we do want this check else users may forget to install at least one of the plugins and then get confused why their sites don't load

tomjn (changes requested):

tomjn:
@aldavigdis I read through the vagrant issue and it seems Chris Roberts made a fix or to use a change similar to:

https://github.com/hashicorp/vagrant/issues/10161#issuecomment-417701063

In terms of policy, goodhosts is best maintained and we have some ties with the owners as @Mte90 built the vagrant plugin, hostsupdater is no longer maintained and hostsmanager has issues

Vagrantfile
aldavigdis: Yeah I can revert this, but that check is done by Vagrant as `config.vagrant.plugins` is set to `['vagrant-goodhosts']` is always executed as the checks around line 376 don't work as expected.

Perhaps a more comprehensive set of changes is needed?

tomjn: hmm we can restore it at a later date then

aldavigdis: I had restored this locally already, so here it is. :)

aldavigdis:
@tomjn We could consider this a quick fix to a specific issue, but it is tempting for me to streamline the Vagrantfile in a separate PR to reflect the state of hostsupdater and hostsmanager by removing references to them.

Is it okay if I create an issue for it and refer to it in a larger PR next week or so?

tomjn:
@aldavigdis I know there are people who still use hostsupdater and hostmanager who would kick up a fuss about it, some because of issues with Windows and large numbers of hosts. A lot of the idiosyncracies in the vagrant file are related to rubycop

I do think it would be nice to move a lot of it out into separate files in a `.vvv/vagrant/` subfolder, and make it a little more functional and OO, e.g. the mapped folders for example don't need re-declaring every single time for each folder, and for each provider, too much repetition

tomjn (approved):

aldavigdis:
I don't even think most rubocop setups would be too happy about the current Vagrantfile TBF and there is no `.rubocop.yml` file present as far as I can see, but I'll be happy to have a more comprehensive look at this in the next couple of days or weeks.

PR #2647, opened 2022-12-27
Disable on Master branch

tomjn:
Tells users to switch to the stable branch and gives them the commands to do so.

This should avoid users being stuck on VVV v3.9.1 without updates and bug fixes, and us maintaining a duplicate of the `stable` branch

## Checks

* [x] I've tested this PR


tomjn:
Screenshot 2022-12-27 at 18 02 55

tomjn:
Simpler message:

Screenshot 2022-12-27 at 18 33 08

tomjn:
Separated the commands as && doesn't work well on Windows:

Screenshot 2022-12-27 at 18 34 57

Vagrantfile
```diff
@@ -6,922 +6,19 @@
...
-        end
-      end
-    end
-  end
+puts "                                              #{boldred}__ #{green}__ #{blue}__ __#{creset}                                     #{creset}"
+puts " The master branch is no longer supported!    #{boldred}\\ V#{green}\\ V#{blue}\\ V /#{creset}"
+puts " Run these commands to switch to #{purple}stable#{creset}:      #{boldred} \\_/#{green}\\_/#{blue}\\_/ #{creset}"
+puts " "
+puts " #{green}git checkout stable#{creset}"
+puts " #{green}git pull#{creset}"
```
Mte90, line 22: maybe suggest to retry again with the vagrant command too?

tomjn: we don't know what the vagrant command was, could have been as simple as `vagrant status`

Mte90: So maybe a message like "try again your command after those".
As I think that the users will just try those commands and not again the vagrant one.

tomjn: I don't see why they wouldn't retry whatever command they were previously attempting, I'm keen to make the message as short and to the point as possible to avoid confusion.

If the message isn't glaringly obvious and short enough, people will ignore it and come to slack instead. They already do this with the `vagrant plugin expunge` message when updating vagrant, they don't bother to read it, too much text

Mte90 (approved):

PR #2658, opened 2023-02-22
change the default PHP to v8.0

tomjn:
## Checks

* [x] I've tested this PR


Mte90:
I have some doubts for this as wordpress still doesn't support officially php 8.0

tomjn:
Most Enterprise WordPress is on 8.0, and 7.4 is unsupported. It's likely 7.4 will be unavailable by the end of the year if I remember the support docs right, some may even suggest we use 8.1.

As for official support, I don't think that will come until the function argument naming project has completed

Mte90:
But enterprise wordpress can change the value in the settings, force all the users to this bump when WP itself doesn't support it to me is like to rush a lot.

tomjn:
I'm already getting warnings from shared hosts about sites running 7.4, if it's necessary couldn't users specify `php: 7.4`?

Mte90:
> I'm already getting warnings from shared hosts about sites running 7.4, if it's necessary couldn't users specify `php: 7.4`?

I understand that, I am wondering on what we should focus. I mean to wordpress contributors or wordpress developers, because this will mean different things for the project as we should not follow anymore so much strict wordpress.org stuff

PR #2659, opened 2023-02-23
update list of supported PHP versions

paulschreiber:
Update list of supported PHP versions in default-config.yml

## Checks

* [X] I've tested this PR
* [X] This PR is for the `develop` branch not the `stable` branch.
* [X] This PR is complete and ready for review.

### DIFF
--- a/config/default-config.yml
+++ b/config/default-config.yml
@@ -94,6 +94,8 @@ extensions:
     #- php73
     #- php74
     #- php80
+    #- php81
+    #- php82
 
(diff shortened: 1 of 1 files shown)


tomjn:
This PR is obsoleted by https://github.com/Varying-Vagrant-Vagrants/VVV/pull/2658

The list is also not a list of what we support, it's purely demonstrative to show users it's possible, PHP versions can and will exist that are not in that list

PR #2677, opened 2023-07-04
Use ubuntu 20 for workflows

tomjn:
Ubuntu 18 is deprecated and gives warnings, also cache vagrant boxes

## Checks

* [ ] I've tested this PR

### DIFF
--- a/.github/workflows/vvv-provisioning.yml
+++ b/.github/workflows/vvv-provisioning.yml
@@ -15,7 +15,7 @@ jobs:
   on-docker:
     name: Docker Provisioner
     # The type of runner that the job will run on
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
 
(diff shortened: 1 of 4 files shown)


tomjn:
Looks like the MacOS 11 image doesn't have vagrant built in, do we need to use MacOS for all of them? Ubuntu might result in easier/quicker workflows with more availability

tomjn:
```
 ▷ Running after_composer hook
 * [PHPCS]: Provisioning PHP_CodeSniffer (phpcs), see https://github.com/squizlabs/PHP_CodeSniffer

In JsonFile.php line 198:
                                                                         
  The file "/home/runner/.config/composer/config.json" is not readable. 
```

tomjn:
I'm leaving the MacOS issue for another time as this change lets us do basic tests with Ubuntu 20

PR #2687, opened 2023-10-01
VVV specific dockerfile, gets built on provision

tomjn:
Moves the dockerfile from pentatonic funk over to VVV, this does centralise things here, though ideally we build this and put it up as an image with a VVV name. I also used `vvv@tomjn.com` but that's only slightly less problematic ( is it optional? ).

The other Q is, what can we put in here to save time during the main VVV provisioner that only docker can do? Do we need a compose file to setup Nginx/MariaDB containers and would that mean changing other parts of VVV to make that work? Would all `wp-config.php` files need to change on docker based setups?

## Checks

* [ ] I've tested this PR

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,60 @@
+FROM ubuntu:20.04
+
+LABEL maintainer="vvv@tomjn.com"
+
+ENV DEBIAN_FRONTEND noninteractive
+
+# install common dependencies
+# ca-certificates usually needed by vagrant to download stuff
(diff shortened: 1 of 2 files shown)


tomjn:
I'm thinking this could be expanded to execute the main provisioner, and perhaps some of the tools provisioner ( parts that don't use shared/mapped folders at runtime ). Then this could be set on a toggle/switch as a way to build a VVV container for a stable version

PR #2690, opened 2023-11-28
Switch to php 8.2 as the default

tomjn:
Switch to 8.2 as the default PHP, 8.0 is no longer supported, 8.1 is security fixes only, and 8.3 is brand new, so 8.2 it is!

If people need 8.0 or other versions they can use the core extensions

## Checks

* [x] I've tested this PR


tomjn:
Also noting that a number of hosts are skipping 8.1 as an option, e.g. WP Engine, and VIP is on 8.2

Mte90 (dismissed):

tomjn:
Checks failed, missed an 8.0 in the site provisioner

Mte90:
There is still something using php 8.0

tomjn:
Ended up needing some fixes in the site template to avoid a `sed` issue in docker

PR #2698, opened 2024-01-10
Remove macos testers on CI

tomjn:
the old versions no longer have runners and we need to upgrade but newer versions are missing necessary software, we can re-add them later once they're fixed and useful again

## Checks

* [x] I've tested this PR


msaggiorato:
Wouldn't macos12 still work?

PR #2699, opened 2024-01-30
Readd runners to run on macos12

msaggiorato:


msaggiorato:
This is not working due to changed hardware in GH runners (info [here](https://github.com/actions/runner-images/issues/8730) and [here](https://github.com/actions/runner-images/issues/8642))

I guess we cannot test anymore.

PR #2714, opened 2024-06-18
Missing dashboard message

tomjn:
If a user deletes the entire `www` folder and the dashboard with it there'll be no `www/default/index.php` loader file and Nginx will show a forbidden HTTP error when visiting `vvv.test`.

Instead it now shows this:

Screenshot 2024-06-18 at 15 18 03

Making it clear what happened and suggesting a solution.

## Checks

* [x] I've tested this PR


tomjn:
I was able to test this by renaming `index.php` and visiting the dashboard

PR #2724, opened 2024-08-04
Set HyperV correctly

mbwhite:
The vm_config provider setting doesn't get set to HyperV. Even though HyperV is detected, it is still attempting to use virtualbox.

## Checks

* [ x] I've updated the changelog.
* [ x] I've tested this PR
* [ x] This PR is for the `develop` branch not the `stable` branch.
* [ x] This PR is complete and ready for review.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -280,6 +280,7 @@ if show_logo
     platform << 'cygwin ' if Vagrant::Util::Platform.cygwin?
     if Vagrant::Util::Platform.windows_hyperv_enabled?
       platform << 'HyperV-Enabled '
+      vvv_config['vm_config']['provider'] = 'hyperv'
     end
     platform << 'HyperV-Admin ' if Vagrant::Util::Platform.windows_hyperv_admin?
     if Vagrant::Util::Platform.windows_admin?


tomjn:
@mbwhite this would force all Windows users to HyperV regardless of their config file, but it's unlikely to solve your problem.

If you create a virtualbox VM by accident, vagrant will continue to use Virtualbox, even if you tell it in the vagrantfile to use something else. You need to destroy the virtualbox machine before vagrant will attempt to create a new Hyper-V machine.

If this change did help you somehow then that heavily implies you've made a formatting mistake of some kind in the YAML config file

mbwhite:
@tomjn thank you for your comments - I'll review the configuration; I don't recall making any changes to the YAML config; which might be the issue - If I've left it unchanged will it default to VirtualBox? Would a docs update here be useful.. can contribute that happily.

For my own understanding, you say this change would force all users to HyperV regardless of config... if HyperV is enabled, that precludes the user of virtual box; does it preclude vmware_desktop?  I'm not familiar with vmware.  Though yes docker as a provider wouldn't be affected..

tomjn:
@mbwhite what it does is if hyper-v is available it would overwrite the config option in code to force hyper-v. This would impact most Windows users if they ever destroyed their machine and recreated it the new VM would be a Hyper-V machine, even if they'd previously used another provider such as VMWare docker or virtualbox

What I suspect has happened is that you've destroyed and recreated the VM and it used the default provider Virtualbox, and now vagrant is trying to prevent data loss/destroying that machine when you specify Hyper-V. Which might be how you first set it up with `--provider=hyperv`

`vagrant status` should confirm if you have a virtualbox VM, and you can tell VVV what provider you prefer via `config.yml`, there's a commented out section with lots of example values at https://github.com/Varying-Vagrant-Vagrants/VVV/blob/develop/config/default-config.yml#L99-L125, uncomment the line for the provider you prefer. Note that vagrant will ignore it if there's already a VM, destroy that VM and reprovision and it should use it

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/Varying-Vagrant-Vagrants/VVV/blob/develop/config/default-config.yml">VVV/config/default-config.yml at develop · Varying-Vagrant-Vagrants/VVV</a></strong></div><div>An open source Vagrant configuration for developing with WordPress - Varying-Vagrant-Vagrants/VVV</div></blockquote>

mbwhite:
Yes - it was an old vm that was there - I hadn't realized that the config.yaml needed updating to spec the provider as well.

PR #2735, opened 2024-09-20
Milliseconds provisioner/hook timing

tomjn:
Refactor how provisioner and hook timing are measured so that milliseconds are included

Screenshot 2024-09-20 at 10 15 38

## Checks

* [x] I've updated the changelog.
* [x] I've tested this PR
* [x] This PR is for the `develop` branch not the `stable` branch.
* [x] This PR is complete and ready for review.

PR #2736, opened 2024-09-20
Improved Network check failure and mariadb service failure

tomjn:
If a network check fails, try it again without the quiet flag so thatwe can see the issue, also fix the listing of failed checks to show just the failed ones

## Checks

* [ ] I've tested this PR


tomjn:
I was hoping this would yield clues to why the CI was failing on github for #2735 , just my luck that it ends up passing anyway

PR #2739, opened 2024-10-20
Update Docker Image to Ubuntu 24.04 ( Noble )

pentatonicfunk:
- Update docker image to use Ubuntu 24.04 ( noble )
-- Add missing `noble` repository list
-- Hacky fix of pip shyaml install on Ubuntu 24.04: see https://pimylifeup.com/python-externally-managed-environment/
- Update mariadb repo to OVH ( only on focal and noble ): see https://github.com/Varying-Vagrant-Vagrants/VVV/issues/2727#issuecomment-2365125433
- Introduce `vvv_get_mysql_service_name`, in attempt to improve `mariadb` vs `mysql` service inconsistency
-- Unable to replicate the service name inconsistency, but it happened. ref: Slack #docker
- Minor output copy fixes / improvements
- Introduce customizable `general.php_cli_version`
-- It does have use case for me personally at least
-- Generally this provides user with better control of PHP CLI version rather than meddling / replicating `vvv_restore_php_default` to change PHP CLI version: https://github.com/Varying-Vagrant-Vagrants/varyingvagrantvagrants.org/blob/main/docs/en-US/adding-a-new-site/changing-php-version.md?plain=1#L83

## Checks

* [x] I've tested this PR: **Partially**

Note on testing:
- Currently i only have linux host testing machine. I've tested with Docker provider, and it went okay, logs attached
- Tried using Virtualbox, but it keeps timing out / error
```shell
Timed out while waiting for the machine to boot. This means that
Vagrant was unable to communicate with the guest machine within
the configured ("config.vm.boot_timeout" value) time period.

If you look above, you should be able to see the error(s) that
Vagrant had when attempting to connect to the machine. These errors
are usually good hints as to what may be wrong.

If you're using a custom box, make sure that networking is properly
working and you're able to connect to the machine. It is a common
problem that networking isn't setup properly in these boxes.
Verify that authentication configurations are also setup properly,
as well.

If the box appears to be booting properly, you may want to increase
the timeout ("config.vm.boot_timeout") value.
```
- If anyone can help testing it, please

Changelogs:
- Please advise


tomjn:
@pentatonicfunk how much of this can we split into separate PRs? Upgrading Ubuntu isn't a small task but adding the sources could be done and released pretty quickly to simplfy merging. Likewise some of the other changes are useful and quick to test on their own.

I am mindful that upgrading Ubuntu the last time resulted in some weirdness that needed investigation.

Also, the service changes, I was thinking of changing how services are restarted to use a VVV specific function to insulate against `service` and `systemctl` since there's a lot of duplication going on

pentatonicfunk:
i will try to cherry picks into separate branches next week.

config/default-config.yml
```diff
@@ -146,6 +146,9 @@ general:
   # GitHub token to use from composer
   #github_token: xxxxxx
 
+  # Set the default version of PHP CLI
+  # php_cli_version: 8.2
```
tomjn: note that by itself this would lead to 8.2 being installed then VVV failing once it restored PHP default as the newly set version wouldn't be installed. It also wouldn't change the dashboard and other sites which would still be configured to 8.2 as the default upstream. There's a lot more work needed to support this, and previously we've added the `php: ...` parameter to sites to make it easier to wrangle since most requests to change the default PHP were to fix the provisioner and CLI using the wrong PHP to run stuff which `php:..` fixes

Vagrantfile
```diff
@@ -480,7 +480,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
 
   # Docker use image.
   config.vm.provider :docker do |d, override|
-    d.image = 'pentatonicfunk/vagrant-ubuntu-base-images:20.04'
+    d.image = 'pentatonicfunk/vagrant-ubuntu-base-images:24.04'
```
tomjn, line 485: does this trigger a VM rebuild for exsting Ubuntu 22 instances that upgrade?

config/homebin/vagrant_provision
```diff
@@ -29,11 +31,11 @@ else
...
-if sudo service mariadb status > /dev/null; then
-  sudo service mariadb restart
-else
-  sudo service mariadb start
-fi
+if sudo service "${mysql_service_name[@]}" status > /dev/null; then
+    sudo service "${mysql_service_name[@]}" restart
+  else
+    sudo service "${mysql_service_name[@]}" start
+  fi
```
tomjn: I'd like to wrap all this in a helper function as we have similar stuff in the vagrant up and provision hooks as well as lots of provisioners and it's super repetitive

provision/core/mariadb/provision.sh
```diff
@@ -125,8 +125,10 @@ function mysql_setup() {
   vvv_info " * Copied /srv/provision/core/mariadb/config/debian.cnf          to /etc/mysql/debian.cnf"
 
   # Due to systemd dependencies, in docker, mysql service is not auto started
+  # also docker isn't systemd based, so thee service name is different: see: https://mariadb.com/kb/en/systemd/ vs https://mariadb.com/kb/en/sysvinit/
```
tomjn:
```suggestion
  # also docker isn't systemd based, so the service name is different: see: https://mariadb.com/kb/en/systemd/ vs https://mariadb.com/kb/en/sysvinit/
```

provision/core/vvv/provision.sh
```diff
@@ -96,8 +96,21 @@ function shyaml_setup() {
   # Used for passing custom parameters to the bash provisioning scripts
   if [ ! -f /usr/local/bin/shyaml ]; then
     vvv_info " * Installing Shyaml for bash provisioning.."
-    sudo pip3 install wheel
-    sudo pip3 install shyaml
+
+    local OSVERSION_NUMBER=$(lsb_release lsb_release -sr)
```
tomjn: shellcheck will probably flag this and a few other things, it should be something like this:
```suggestion
    local OSVERSION_NUMBER
    OSVERSION_NUMBER=$(lsb_release lsb_release -sr)
```

provision/core/vvv/provision.sh
```diff
@@ -96,8 +96,21 @@ function shyaml_setup() {
...
+
+    # Ubuntu 24 making it hard to install pip packages, throwing externally-managed-environment error
+    # https://stackoverflow.com/a/75722775
+    if dpkg --compare-versions "${OSVERSION_NUMBER[@]}" ge "24.04"
+    then
+      # to make it available globally this is the last workaround, hopefully it doesn't break the system
+      # TODO: try to find a better alternative way to install
```
tomjn: is there somewhere I can read up on why this happens or where the fix came from? Would be good to have it in the comments, I've created a separate PR for shyaml fixes

provision/provision-helpers.sh
```diff
@@ -731,3 +732,13 @@ function vvv_search_replace_in_file() {
   fi
 }
 export -f vvv_search_replace_in_file
+
+# @description Identify the service name for the MySQL service via /etc/init.d/ contents
```
tomjn: in the past both have worked and been alias'd somehow

tomjn: that wasn't something we did though, but an observation of how it came out the box

Vagrantfile
pentatonicfunk: good question, i dont know. it would be vagrant territory, no ? the decision whether docker container need rebuild or not would be based on what vagrant flow will be issued. in any case better to test it, i'll try to test

provision/core/vvv/provision.sh
pentatonicfunk: - https://peps.python.org/pep-0668/
- https://pimylifeup.com/python-externally-managed-environment/
- https://stackoverflow.com/questions/75608323/how-do-i-solve-error-externally-managed-environment-every-time-i-use-pip-3/75722775#75722775

provision/provision-helpers.sh
pentatonicfunk: exactly yea, it seems how mariadb installer works ( not vvv ). i am thinking maybe the mariadb vs mysql thing is a glitch that i experienced one time, as i can't replicate it again. my last provision, it gets correctly `mariadb`.
i will do 1 more tests, and maybe revert the changes mariadb vs mysql service thing from this PR

Vagrantfile
tomjn: main concerns is if vagrant would kick up a fuss about mismatching images or if it would discard the container ( and the database + data ) to create a new 24 container

config/default-config.yml
pentatonicfunk: that is valid unintended effect that i didn't foresee. so i think its better to revert this change as it could break things rather than do good.

tomjn: I get why people want it, but I want to make sure it's not an X Y solution as it'll get abused and has lots of gotchas, I'm curious about your use case though

tomjn: I'm also mindful we will want a generic PHP provisioner at some point with templated PHP ini files etc rather than a dedicated hardcoded provisioner for each

pentatonicfunk: my use case maybe not that generic and/or critical enough, but here it is
- im trying to match php version in WP CLI to the one that i have on my production server, that way when i do test my script locally ( in vvv ), it ensure script run successfully in that specific php version. Changing PHP version for WP CLI in runtime is not that straightforward, so having the PHP CLI to be statically same with my production server, would avoid me make a wrong judgement.

the solution that i offered here, maybe just too far and/or not actually the solution that i actually need. in my other fork, i have this script in homebin: `vvv_change_php_default` and that was enough.
```php
#!/bin/bash

DEFAULTPHP=${1:-"7.4"}
echo " * Changing the default PHP CLI version ( ${DEFAULTPHP} )"
update-alternatives --set php "/usr/bin/php${DEFAULTPHP}"
update-alternatives --set phar "/usr/bin/phar${DEFAULTPHP}"
update-alternatives --set phar.phar "/usr/bin/phar.phar${DEFAULTPHP}"
update-alternatives --set phpize "/usr/bin/phpize${DEFAULTPHP}"
update-alternatives --set php-config "/usr/bin/php-config${DEFAULTPHP}"
echo " * Change completed"
```
i just assumed since its similar code with `vvv_restore_php_default`, then maybe having it in that script would be better, which is wrong assumption obviously.

tomjn: hmmm, something I wanted to do was make it auto-switch PHP based on the folder you were in, but it's not as straightforward especially if you've multiple sessions. There's a ticket to add the PHP version into the prompt so it's more obvious for people who aren't aware of all of this and mistakenly run in the wrong version they expect

provision/provision-helpers.sh
pentatonicfunk: so i've reverted these, it seems the issue is something in my lab machine.

i've been testing the mariadb with multiple vagrant destroy and vagrant up --provision, its consistently working fine with mariadb
```shell
vagrant@vvv:~$ sudo service mariadb status
 * /usr/bin/mariadb-admin  Ver 10.0 Distrib 10.11.9-MariaDB, for debian-linux-gnu on x86_64
Copyright (c) 2000, 2018, Oracle, MariaDB Corporation Ab and others.

Server version          10.11.9-MariaDB-ubu2404
Protocol version        10
Connection              Localhost via UNIX socket
UNIX socket             /run/mysqld/mysqld.sock
Uptime:                 12 sec

Threads: 1  Questions: 63  Slow queries: 0  Opens: 33  Open tables: 26  Queries per second avg: 5.250
```

config/default-config.yml
pentatonicfunk: this also reverted, as you've mentioned that it can cause issues down the road. and my use case is too specific to be having this in default config

provision/core/vvv/provision.sh
pentatonicfunk: adjusted

pentatonicfunk: huh, this adjustments causing error for some reason
```
lsb_release: error: No arguments are permitted
```
https://github.com/Varying-Vagrant-Vagrants/VVV/actions/runs/11580749924/job/32239965581?pr=2739

pentatonicfunk: reverted it to the original changes. the checks / linting seems to be fine: https://github.com/Varying-Vagrant-Vagrants/VVV/actions/runs/11581054132/job/32240928199?pr=2739
and we already doing these kind of local syntax previously: https://github.com/Varying-Vagrant-Vagrants/VVV/blob/develop/provision/core/vvv/provision.sh#L62-L64

tomjn: ah I think the problem is a mistake on my part, `lsb_release` vs `lsb_release lsb_release`

config/homebin/vagrant_provision
pentatonicfunk: reverted this shenanigans as well https://github.com/Varying-Vagrant-Vagrants/VVV/pull/2739#discussion_r1821351163

provision/core/vvv/provision.sh
pentatonicfunk: adjusted, and seems to be ran okay: https://github.com/Varying-Vagrant-Vagrants/VVV/actions/runs/11858839047/job/33050382482?pr=2739
let me know if it needs further adjustments

Vagrantfile
pentatonicfunk: so i tested this, this is how i tested it

### use `upstream/develop` branch
- Vagrantfile still using 20.04 image
- do `vagrant up --provision`
- verify ubuntu version via `vagarant ssh`
```shell
vagrant@vvv:~$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.6 LTS
Release:        20.04
Codename:       focal
```
- create new post in one.wordpress.test to verify DB okay
image

### switch to `origin/docker-provider`
- Vagrantfile now 24.4
-  do `vagrant halt` to stop docker
- do `vagrant reload --provision` to trigger reprovision
- verify ubuntu version via `vagarant ssh`, expecting the Ubuntu should stay in 20.04
```shell
Last login: Fri Nov 15 15:33:51 2024 from 172.17.0.1
vagrant@vvv:~$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.6 LTS
Release:        20.04
Codename:       focal
```
- verify whether DB not discarded ( post still exists )
image
 
 i think this is what we expected, right ?
- no complains on provisioner
- container not discarded
- DB not discarded

i think what we can do is, add somewhere in docs ( if we haven't), if user wants to upgrade their OS is by doing 
- `db_backup` 
- `vagrant destroy`
- `vagrant up --provision` --> this step will trigger 24.04
- `db_restore`

tomjn:
I've made updates to switch the other providers to Ubuntu 24 LTS and tweak the workflows, it looks like the docker provisioner action is failing though

tomjn:
Running this locally with:

```
act -j 'on-docker' --container-architecture linux/arm64 -P ubuntu-24.04=ghcr.io/catthehacker/ubuntu:act-24.04
```

tomjn:
TLDR the symlinks were the problem

tomjn:
@pentatonicfunk it's been a long time coming but I think it's time to merge

